PR #18, opened 2025-03-18
add create_issue tool

juruen:
## Context

This PR adds a **tool** for creating an issue. We initially postponed this feature to wait for **mcp-go** to support arrays for describing arguments.

The library finally added support for arrays a few days ago. However, my first implementation, which used them, caused issues in VS Code.

Since VS Code is my main test platform, I decided to implement the array using a comma-separated list instead. We can revisit this decision once the issue is resolved.

Screenshot 2025-03-18 at 15 41 26


juruen:
> OK @juruen can you make an issue to re-visit the CSV style, and use proper JSON RPC arrays once it works?

Done! https://github.com/github/github-mcp-server/issues/31

PR #28, opened 2025-03-20
feat: enterprise host support

SamMorrowDrums:
Adds support for a `gh-host` param so that enterprises work. Tested it out with failing attempt to call localhost:

`go run cmd/github-mcp-server/main.go stdio --gh-host github.localhost`

```
GitHub MCP Server running on stdio
{
  "jsonrpc": "2.0",
  "id": 3,
  "error": {
    "code": -32603,
    "message": "failed to get user: Get \"https://api.github.localhost/user\": dial tcp [::1]:443: connect: connection refused"
  }
}
```

So the override is working well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -376,7 +376,7 @@ Run **Preferences: Open User Settings (JSON)**, and create or append to the `mcp
         "args": ["stdio"],
         "env": {
           "GITHUB_PERSONAL_ACCESS_TOKEN": "${input:githubpat}"
-        }
+        },
       }
     }
   }


toby:
Awesome! Do think it would make sense to support an env var (in addition to or instead of the flag)? Just because we're using them elsewhere.

cmd/github-mcp-server/main.go
```diff
@@ -92,6 +95,20 @@ func runStdioServer(logger *log.Logger, logCommands bool) error {
...
+		}
+
+		uploadURL, err := url.Parse(fmt.Sprintf("https://uploads.%s/", host))
+		if err != nil {
+			return fmt.Errorf("failed to parse provided GitHub host URL: %w", err)
+		}
+
+		ghClient.BaseURL = parsedURL
+		ghClient.UploadURL = uploadURL
+	}
```
juruen: Maybe we can use the `WithEnterpriseURLs()` function in the `gogithub.Client` instead?

SamMorrowDrums:
> Awesome! Do think it would make sense to support an env var (in addition to or instead of the flag)? Just because we're using them elsewhere.

Yeah, that's easy. I was considering doing so too. Just didn't :sweat_smile:

README.md
```diff
@@ -375,7 +376,7 @@ Run **Preferences: Open User Settings (JSON)**, and create or append to the `mcp
         "args": ["stdio"],
         "env": {
           "GITHUB_PERSONAL_ACCESS_TOKEN": "${input:githubpat}"
-        },
+        }
```
```suggestion
        },
```

PR #29, opened 2025-03-20
feat: read-only flag

SamMorrowDrums:
Add a read-only flag and gate endpoints that would otherwise write behind it, so that people can avoid the risk that an LLM can write to their GitHub account entirely.

### DIFF
--- a/cmd/github-mcp-server/main.go
+++ b/cmd/github-mcp-server/main.go
@@ -32,12 +32,14 @@ var (
 		Long:  `Start a server that communicates via standard input/output streams using JSON-RPC messages.`,
 		Run: func(cmd *cobra.Command, args []string) {
 			logFile := viper.GetString("log-file")
+			readOnly := viper.GetBool("read-only")
+			exportTranslations := viper.GetBool("export-translations")
 			logger, err := initLogger(logFile)
 			if err != nil {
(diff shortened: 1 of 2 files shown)


juruen (approved):
🚀

formulahendry (commented):
One issue still worth fixing before merge:

If a config file overrides only `storage.dir`, the default `storage.inboxDir` does not get re-derived from the new storage root. `defaultConfig()` initializes `storage.inboxDir` to `~/.wechat-acp/inbox`, then `Object.assign(config.storage, fileConfig.storage ?? {})` may update `storage.dir` without updating `inboxDir`.

Example:

```json
{ "storage": { "dir": "/secure/wechat-acp" } }
```

Received binary files would still be saved under `~/.wechat-acp/inbox`, not `/secure/wechat-acp/inbox` as documented by the `<storage.dir>/inbox` default. This can put sensitive received files in an unexpected location.

Suggested fix: distinguish an explicitly configured `storage.inboxDir` from the derived default. After config file, `--instance`, and CLI overrides are resolved, set `config.storage.inboxDir = path.join(config.storage.dir, "inbox")` only when the user did not explicitly set `storage.inboxDir` and did not pass `--no-inbox`.

PR #35, opened 2025-03-24
validate tools params

juruen:
## Context

This PR changes the way request parameters are used to make sure that they are always validated before they are used. We were previously relying on the parameters honoring the expected types, but now we are explicitly checking and returning an error if the parameter types are not as expected.

This way, even if there's a schema mismatch between the the schema itself and the implementation, the server won't crash trying to cast invalid types with something like `request.Params.Arguments["owner"].(string)`.

For the implementation itself, I've added a few convenience functions that are used across the different tools.

Some tests have been modified to reflect the new behavior for error returning.

Apologies for the largish PR, but as we are doing the same: validating and fetching the tools parameters, it felt like we could do it in a single PR. As always, I'm happy to split it out.

Thanks! ❤️

### DIFF
--- a/pkg/github/code_scanning.go
+++ b/pkg/github/code_scanning.go
@@ -30,15 +30,15 @@ func getCodeScanningAlert(client *github.Client, t translations.TranslationHelpe
 			),
 		),
 		func(ctx context.Context, request mcp.CallToolRequest) (*mcp.CallToolResult, error) {
-			owner, err := requiredStringParam(request, "owner")
+			owner, err := requiredParam[string](request, "owner")
 			if err != nil {
 				return mcp.NewToolResultError(err.Error()), nil
 			}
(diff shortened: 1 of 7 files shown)


juruen:
> Could we use generics to make this shorter?

That was my initial attempt :) Unfortunately, we get `floats` for numbers, and we require `ints` in the GH API, so using generics didn't work well there.  Or it only worked well for strings and booleans. You could still make it work, but it became a bit more cumbersome, and at that point I decided not to go with generics.

juruen:
I gave @SamMorrowDrums  suggestion a go in https://github.com/github/github-mcp-server/pull/35/commits/3c6467ffef6b2493d69533d34776d5cdd97b65ce

we use generics for types that are comparable, in our case, this is mostly used by `strings` and `booleans`, and separate functions for `ints` to take care of the float to int cast.

PR #42, opened 2025-03-26
[docker] build arm64

juruen:
## Context

Small PR to also build docker images for arm64.


zuobubedingyideren:
申请接入GitHub官方MCP服务

PR #43, opened 2025-03-26
update install instructions

juruen:
## Context

Update doc on how to install and test this out in VS insiders. Show Docker option as the first choice to do so.

As variable inputs don't seem to work to introduce the token, we remove that from the README until they work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,120 @@ GitHub MCP Server implemented in Go.
 Create a GitHub Personal Access Token with the appropriate permissions
 and set it as the GITHUB_PERSONAL_ACCESS_TOKEN environment variable.
 
+## Testing on VS Code Insiders
+
+### Requirements
+
+You can either use a Docker image or build the binary from the repo.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -351,40 +351,103 @@ export GITHUB_MCP_TOOL_ADD_ISSUE_COMMENT_DESCRIPTION="an alternative description
...
+```bash
+GOPRIVATE=github.com/github go install github.com/github/github-mcp-server/cmd/github-mcp-server@latest
+```
+
+This will install the `github-mcp-server` binary in your `$GOPATH/bin` directory.
+
+Find where the binary is installed by running:
+
+```bash
+which github-mcp-server
```
SamMorrowDrums: This implies the `$GOPATH/bin` is in your `$PATH` which it might not be, but I think it's still OK to keep.
```suggestion
(note this assumes $GOPATH/bin is in your $PATH)
which github-mcp-server 
```

PR #45, opened 2025-03-27
Update VS Code config JSON

williammartin:
## Description

It is now possible in VS Code to use `${input:github-pat}` to indicate that the user should be prompted for an input. This helps keep credentials out of settings files.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,32 +63,36 @@ code-insiders
 
 Another way is to set the environment variable in your shell configuration file (e.g., `.bashrc`, `.zshrc`, etc.).
 
-Run **Preferences: Open User Settings (JSON)**, and create or append to the `mcp` setting:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -63,33 +63,37 @@ code-insiders
 
 Another way is to set the environment variable in your shell configuration file (e.g., `.bashrc`, `.zshrc`, etc.).
 
-Run **Preferences: Open User Settings (JSON)**, and create or append to the `mcp` setting:
+Create a new file `.vscode/mcp.json` and provide this configuration:
```
juruen, line 66: Maybe we should remove the lines above that mention setting the token as env variable?

williammartin: I have no strong opinion on that. I didn't want to take away another option that might be in use. It does seem like this `input` approach should be elevated to a recommended approach at least?

PR #52, opened 2025-03-30
feat: enable secret protection

SamMorrowDrums:
Enable checking of repository settings, and add a specific tool for toggling secret scanning settings only.

https://github.com/user-attachments/assets/46a67ec8-a0e5-4e61-b8db-067c93238ddd

Advanced Security is a pre-requisite for using this on private repos (and or Secret Protection SKU when it launches next week), so we will need to handle those quirks.

I had to make some manual http calls, because the go wrapper does not wrap this feature.

TODO before merge:

- [ ] decide on cleanest way to handle the public/private repo stuff where there is a pre-requisite to enabling secret protection
- [ ] we could just have a catch-all configuration setting endpoint, with the massive number of possible settings?
- [ ] I can make all fields optional, but the annoying thing is that you can't have push protection on, without secret protection. The API is not conducive to wrapping neatly in a tool, but I must find a way.
- [ ] can we also enable Code Scanning default setup?
- [ ] add tests

### DIFF
--- a/pkg/github/repositories.go
+++ b/pkg/github/repositories.go
@@ -1,6 +1,7 @@
 package github
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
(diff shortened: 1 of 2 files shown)


williammartin:
@SamMorrowDrums what do you want to do with this PR?

Should we kick this back to an issue to discuss?

SamMorrowDrums:
I think this is that issue:

https://github.com/github/github-mcp-server/issues/176

And now that we have the secret protection toolset, I think we can probably proceed.

PR #58, opened 2025-04-01
docs: update README

toby:


README.md
```diff
@@ -1,130 +1,106 @@
...
-If the above command works, you are good to go.
-
-#### Build from repo
-First, install `github-mcp-server` by cloning the repo and running the following command:
-
-```bash
-go install ./cmd/github-mcp-server
 ```
+When you start the server, VS Code will prompt for your token. On top of
+`servers`, you should see a `Start` link to start the server.
```
SamMorrowDrums: This didn't quite track for me. Perhaps:

```suggestion
VS Code is now configured to prompt for your token when you start the server. On top of
`servers`, you should see a `Start` link to start the server.
```

> On top of
`servers`, you should see a `Start` link to start the server.

Where? It is not clear to me what part of the UO this is referring to.

README.md
```diff
@@ -1,130 +1,106 @@
...
-```
+The flag `--gh-host` and the environment variable `GH_HOST` can be used to set
+the GitHub Enterprise Server hostname.
 
-Another way is to set the environment variable in your shell configuration file (e.g., `.bashrc`, `.zshrc`, etc.).
+## i18n / Overriding Descriptions
 
-Create a new file `.vscode/mcp.json` and provide this configuration:
+The descriptions of the tools can be overridden by creating a
+github-mcp-server.json file in the same directory as the binary.
```
```suggestion
github-mcp-server-config.json file in the same directory as the binary.
```

Double check the name of the file, @williammartin  changed it for us because it was otherwise trying to read the binary and reporting a warning on startup.

README.md
```diff
@@ -1,130 +1,106 @@
...
-        "command": "path-to-your/github-mcp-server",
-        "args": ["stdio"],
-        "env": {   }
-      }
-    }
-  }
-}
+```sh
+./github-mcp-server --export-translations
+cat github-mcp-server.json
```
```suggestion
cat github-mcp-server-config.json
```

I could be wrong on the name but it's specified in the code, I am on mobile and can't look, but this does need updating also.

README.md
```diff
@@ -1,130 +1,106 @@
...
+- Building AI powered tools and applications that interact with GitHub's ecosystem.
 
-## Testing in VS Code Insiders
+## Prerequisites
 
-### Requirements
+[Create a GitHub Personal Access Token](https://github.com/settings/personal-access-tokens/new)
+with the permissions you want to give the MCP server. You can also install
+[Docker](https://www.docker.com/) to run the server in a container or build the
+binary from the repo.
```
```suggestion
with the permissions you want to give the MCP server. To run the server in a container, you will need to have
[Docker](https://www.docker.com/) installed.
Alternatively you can build the binary from the source code in your repo.
```

README.md
```diff
@@ -1,130 +1,106 @@
 # GitHub MCP Server
 
-GitHub MCP Server implemented in Go.
+The GitHub MCP Server is a Model Context Protocol (MCP) server that provides
```
mntlty:
```suggestion
The GitHub MCP Server is a [Model Context Protocol (MCP)](https://modelcontextprotocol.io/introduction) server that provides
```

README.md
```diff
@@ -1,130 +1,106 @@
...
+seamless integration with GitHub's APIs, enabling advanced automation and
```
mntlty:
```suggestion
seamless integration with GitHub APIs, enabling advanced automation and
```

README.md
toby: good catch, it is indeed `github-mcp-server-config.json` https://github.com/github/github-mcp-server/blob/450fa651a68a6b3040151c81e26ba35b5b62e290/pkg/translations/translations.go#L27

README.md
```diff
@@ -1,130 +1,106 @@
...
-and set it as the GITHUB_PERSONAL_ACCESS_TOKEN environment variable.
+- Automating GitHub workflows and processes.
+- Extracting and analyzing data from GitHub repositories.
+- Building AI powered tools and applications that interact with GitHub's ecosystem.
 
-## Testing in VS Code Insiders
+## Prerequisites
 
-### Requirements
+[Create a GitHub Personal Access Token](https://github.com/settings/personal-access-tokens/new)
```
mntlty: the current link doesn't give context to people who aren't familiar with PATs and won't work if a user isn't logged in. suggest pointing to our public docs instead
```suggestion
[Create a GitHub Personal Access Token](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/managing-your-personal-access-tokens)
```

README.md
```diff
@@ -1,130 +1,106 @@
...
+with the permissions you want to give the MCP server. You can also install
```
mntlty: which permissions do people need for their token? to avoid confusion it would help to list those out, possibly broken into readonly and and write sections.

README.md
mntlty: agree with breaking this out, the current sentence make it sound like you build the binary in docker.

do we have a preference? if so we should point users to it.

would also help to link to some docs for how to build with go

README.md
```diff
@@ -1,130 +1,106 @@
...
+## Installation
 
-#### Docker image
-
-As of now, this repo is private, and hence the docker image is not available publicly. To pull it,
-you need to make sure you can access the GitHub docker registry. See [this](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry#authenticating-with-a-personal-access-token-classic)
-for more details.
-
-To make sure you can access the GitHub docker registry, run the following command:
+### Usage with VS Code
```
mntlty, line 24: does this require that they have completed the prerequisite step above to have it installed? what about commands to get it running?

README.md
```diff
@@ -1,130 +1,106 @@
...
 
-If the above command works, you are good to go.
-
-#### Build from repo
-First, install `github-mcp-server` by cloning the repo and running the following command:
-
-```bash
-go install ./cmd/github-mcp-server
 ```
+When you start the server, VS Code will prompt for your token. On top of
```
mntlty: we should tell them how to start the server

README.md
```diff
@@ -1,130 +1,106 @@
...
+    }
+  }
+}
 ```
 
-### Start VS Code Insiders
+### Build from source
 
-Start VS Code Insiders and make sure you pass the `GITHUB_PERSONAL_ACCESS_TOKEN` environment variable to the process.
+If you don't have Docker, you can use `go` to build the binary in the
```
mntlty, line 63: this makes it sounds like docker is preferred over building from source. is that the case?

README.md
```diff
@@ -1,130 +1,106 @@
...
+`cmd/github-mcp-server` directory, and use the `github-mcp-server stdio`
```
mntlty, line 64: this is hard to parse and will be harder for people who aren't used to building go projects

README.md
```diff
@@ -1,130 +1,106 @@
...
-you need to make sure you can access the GitHub docker registry. See [this](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry#authenticating-with-a-personal-access-token-classic)
-for more details.
-
-To make sure you can access the GitHub docker registry, run the following command:
+### Usage with VS Code
 
 ```bash
-docker pull ghcr.io/github/github-mcp-server:main
-```
+code --add-mcp '{"name":"github","command":"docker","args":["run", "-i", "--rm", "-e", "GITHUB_PERSONAL_ACCESS_TOKEN", "ghcr.io/github/github-mcp-server:main"], "env": {"GITHUB_PERSONAL_ACCESS_TOKEN": "${input:github-pat}"}, "inputs": [{ "id": "github-pat", "type": "promptString", "description": "Github Personal Access Token", "password": true}]}'
```
mntlty, line 33: is `GITHUB_PERSONAL_ACCESS_TOKEN` required both as an env var to docker and as an env var for `code`? I'm not quite sure how to parse this one so maybe it's pulling it in from code and setting it as an env var in the docker 🤔 

would be good to explain what this is doing

README.md
mntlty:
```suggestion
`github-mcp-server.json` file in the same directory as the binary.
```

README.md
```diff
@@ -1,130 +1,106 @@
...
+```sh
+export GITHUB_MCP_TOOL_ADD_ISSUE_COMMENT_DESCRIPTION="an alternative description"
 ```
 
-## GitHub Enterprise Server
-
-The flag `--gh-host` and the environment variable `GH_HOST` can be used to set the GitHub Enterprise Server hostname.
-
-
 ## Tools
```
mntlty, line 111: suggest breaking this out into a different file which we link to here. ideally one that is dynamically generated so this doesn't have to be kept up to date manually

README.md
toby: It's part of the VS Code config. For this one, I modeled it on the Playwright readme:  https://github.com/microsoft/playwright-mcp?tab=readme-ov-file#installation-in-vs-code

PR #69, opened 2025-04-02
repository resource tests

mntlty:
Refactors repository_resource to make it easier to test, then adds some tests mostly copied from https://github.com/github/github-mcp-server/blob/d8b0056998c621ab9be8678259a5da79d11b5c15/pkg/github/repositories_test.go#L18 which seems quite similar.

### DIFF
--- a/pkg/github/repository_resource.go
+++ b/pkg/github/repository_resource.go
@@ -4,7 +4,10 @@ import (
 	"context"
 	"encoding/base64"
 	"errors"
+	"fmt"
+	"io"
 	"mime"
+	"net/http"
 	"path/filepath"
 	"strings"
 
@@ -113,7 +116,12 @@ func repositoryResourceContentsHandler(client *github.Client) func(ctx context.C
 			for _, entry := range directoryContent {
 				mimeType := "text/directory"
 				if entry.GetType() == "file" {
-					mimeType = mime.TypeByExtension(filepath.Ext(entry.GetName()))
+					// this is system dependent, and a best guess
+					ext := filepath.Ext(entry.GetName())
+					mimeType = mime.TypeByExtension(ext)
+					if ext == ".md" {
+						mimeType = "text/markdown"
+					}
 				}
 				resources = append(resources, mcp.TextResourceContents{
 					URI:      entry.GetHTMLURL(),
@@ -127,28 +135,62 @@ func repositoryResourceContentsHandler(client *github.Client) func(ctx context.C
 		}
 		if fileContent != nil {
 			if fileContent.Content != nil {
-				decodedContent, err := fileContent.GetContent()
+				// download the file content from fileContent.GetDownloadURL() and use the content-type header to determine the MIME type
+				// and return the content as a blob unless it is a text file, where you can return the content as text
+				req, err := http.NewRequest("GET", fileContent.GetDownloadURL(), nil)
 				if err != nil {
-					return nil, err
+					return nil, fmt.Errorf("failed to create request: %w", err)
 				}
 
-				mimeType := mime.TypeByExtension(filepath.Ext(fileContent.GetName()))
+				resp, err := client.Client().Do(req)
+				if err != nil {
+					return nil, fmt.Errorf("failed to send request: %w", err)
+				}
+				defer func() { _ = resp.Body.Close() }()
 
+				if resp.StatusCode != http.StatusOK {
+					body, err := io.ReadAll(resp.Body)
+					if err != nil {
+						return nil, fmt.Errorf("failed to read response body: %w", err)
+					}
+					return nil, fmt.Errorf("failed to fetch file content: %s", string(body))
+				}
+
+				ext := filepath.Ext(fileContent.GetName())
+				mimeType := resp.Header.Get("Content-Type")
+				if ext == ".md" {
+					mimeType = "text/markdown"
+				} else if mimeType == "" {
+					// backstop to the file extension if the content type is not set
+					mimeType = mime.TypeByExtension(filepath.Ext(fileContent.GetName()))
+				}
+
+				// if the content is a string, return it as text
 				if strings.HasPrefix(mimeType, "text") {
+					content, err := io.ReadAll(resp.Body)
+					if err != nil {
+						return nil, fmt.Errorf("failed to parse the response body: %w", err)
+					}
+
 					return []mcp.ResourceContents{
 						mcp.TextResourceContents{
 							URI:      request.Params.URI,
 							MIMEType: mimeType,
-							Text:     decodedContent,
+							Text:     string(content),
 						},
 					}, nil
+				}
+				// otherwise, read the content and encode it as base64
+				decodedContent, err := io.ReadAll(resp.Body)
+				if err != nil {
+					return nil, fmt.Errorf("failed to parse the response body: %w", err)
 				}
 
 				return []mcp.ResourceContents{
 					mcp.BlobResourceContents{
 						URI:      request.Params.URI,
 						MIMEType: mimeType,
-						Blob:     base64.StdEncoding.EncodeToString([]byte(decodedContent)), // Encode content as Base64
+						Blob:     base64.StdEncoding.EncodeToString(decodedContent), // Encode content as Base64
 					},
 				}, nil
 			}
--- a/pkg/github/repository_resource_test.go
+++ b/pkg/github/repository_resource_test.go
@@ -2,7 +2,6 @@ package github
 
 import (
 	"context"
-	"encoding/base64"
 	"net/http"
 	"testing"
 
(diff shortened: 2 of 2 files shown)


pkg/github/repository_resource.go
```diff
@@ -14,109 +14,137 @@ import (
 )
 
 // getRepositoryContent defines the resource template and handler for the Repository Content API.
-func getRepositoryContent(client *github.Client, t translations.TranslationHelperFunc) (mainTemplate mcp.ResourceTemplate, reftemplate mcp.ResourceTemplate, shaTemplate mcp.ResourceTemplate, tagTemplate mcp.ResourceTemplate, prTemplate mcp.ResourceTemplate, handler server.ResourceTemplateHandlerFunc) {
```
mntlty, line 17: this format of returning multiple values made it quite hard to evaluate and test, instead I broke them up into individual functions

pkg/github/repository_resource.go
```diff
@@ -14,109 +14,137 @@ import (
...
 
-			owner := request.Params.Arguments["owner"].([]string)[0]
-			repo := request.Params.Arguments["repo"].([]string)[0]
-			// path should be a joined list of the path parts
-			path := strings.Join(request.Params.Arguments["path"].([]string), "/")
+func repoContentsResourceHandler(client *github.Client) func(ctx context.Context, request mcp.ReadResourceRequest) ([]mcp.ResourceContents, error) {
+	return func(ctx context.Context, request mcp.ReadResourceRequest) ([]mcp.ResourceContents, error) { // Extract parameters from request.Params.URI
 
-			opts := &github.RepositoryContentGetOptions{}
+		owner := request.Params.Arguments["owner"].([]string)[0]
```
mntlty: this is the previous implementation, which I found a bit problematic.
1. I'm not sure why this would be a slice of strings, rather than a string since these are keyed arguments. if we can change the type here I would very much like to do so
2. as it is currently implemented, this can panic from either the type assertion or the index out of bounds if we get an empty slice with length 0.

If we cannot change 1, I would like to make the code safer by addressing 2

pkg/github/repository_resource.go
```diff
@@ -14,109 +14,137 @@ import (
...
-			if ok {
-				opts.Ref = sha[0]
-			}
+		opts := &github.RepositoryContentGetOptions{}
 
-			branch, ok := request.Params.Arguments["branch"].([]string)
-			if ok {
-				opts.Ref = "refs/heads/" + branch[0]
-			}
+		sha, ok := request.Params.Arguments["sha"].([]string)
```
mntlty, line 90: left this as it was previously, where one function was shared by all ResourceTemplates. I think it would be better to continue the refactor by creating individual functions that only parse arguments that are expected, for example the `sha` argument would only be parsed when getting a commit

pkg/github/repository_resource.go
```diff
@@ -14,109 +14,137 @@ import (
...
+			opts.Ref = "refs/pull/" + prNumber[0] + "/head"
+		}
+
+		// Use the GitHub client to fetch repository content
+		fileContent, directoryContent, _, err := client.Repositories.GetContents(ctx, owner, repo, path, opts)
+		if err != nil {
+			return nil, err
+		}
+
+		if directoryContent != nil {
```
mntlty, line 114: not sure if directoryContent and fileContent are mutually exclusive, that is the previous behavior which I haven't changed (I only removed the unnecessary `else` clause)

pkg/github/repository_resource_test.go
```diff
@@ -0,0 +1,172 @@
...
+	"testing"
+
+	"github.com/google/go-github/v69/github"
+	"github.com/mark3labs/mcp-go/mcp"
+	"github.com/migueleliasweb/go-github-mock/src/mock"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func Test_RepoContentsResourceHandler(t *testing.T) {
```
mntlty: largely copied from https://github.com/github/github-mcp-server/blob/d8b0056998c621ab9be8678259a5da79d11b5c15/pkg/github/repositories_test.go#L18

pkg/github/repository_resource_test.go
```diff
@@ -0,0 +1,172 @@
...
+			Name:    github.Ptr("src"),
+			Path:    github.Ptr("src"),
+			SHA:     github.Ptr("def456"),
+			HTMLURL: github.Ptr("https://github.com/owner/repo/tree/main/src"),
+		},
+	}
+	expectedDirContent := []mcp.TextResourceContents{
+		{
+			URI:      "https://github.com/owner/repo/blob/main/README.md",
+			MIMEType: "",
```
mntlty: feels like this shouldn't be empty, but that's what the current code returns 🤷

pkg/github/repository_resource.go
```diff
@@ -13,110 +13,131 @@ import (
...
+					mcp.BlobResourceContents{
+						URI:      request.Params.URI,
+						MIMEType: mimeType,
+						Blob:     base64.StdEncoding.EncodeToString([]byte(decodedContent)), // Encode content as Base64
+					},
+				}, nil
+			}
 		}
+
+		return nil, nil
```
mntlty: I added a test for this behavior, but it seems wrong - I feel like we should be returning an error here

pkg/github/repository_resource_test.go
```diff
@@ -0,0 +1,194 @@
...
+			}
+
+			require.NoError(t, err)
+			require.ElementsMatch(t, resp, tc.expectedResult)
+		})
+	}
+}
+
+func Test_getRepositoryResourceContent(t *testing.T) {
+	tmpl, _ := getRepositoryResourceContent(nil, translations.NullTranslationHelper)
```
mntlty, line 262: I could not find a sensible assertion for the handler func, it's execution is tested above

pkg/github/repository_resource_test.go
```diff
@@ -0,0 +1,194 @@
...
+			Name:    github.Ptr("src"),
+			Path:    github.Ptr("src"),
+			SHA:     github.Ptr("def456"),
+			HTMLURL: github.Ptr("https://github.com/owner/repo/tree/main/src"),
+		},
+	}
+	expectedDirContent := []mcp.TextResourceContents{
+		{
+			URI:      "https://github.com/owner/repo/blob/main/README.md",
+			MIMEType: "text/markdown; charset=utf-8",
```
mntlty: frustratingly, this value exists on the ubuntu runner https://github.com/github/github-mcp-server/actions/runs/14212790977/job/39822938856?pr=69 and does not exist on the mac runner https://github.com/github/github-mcp-server/actions/runs/14212837651/job/39823067424?pr=69 , so one of the other fails. not sure how best to handle this

mntlty: maybe another file type would be simpler

mntlty: or maybe not, looks like ` "text/directory"` is also an issue between the two runner types

pkg/github/server.go
```diff
@@ -25,13 +25,11 @@ func NewServer(client *github.Client, readOnly bool, t translations.TranslationH
...
 
 	// Add GitHub Resources
-	defaultTemplate, branchTemplate, tagTemplate, shaTemplate, prTemplate, handler := getRepositoryContent(client, t)
-
-	s.AddResourceTemplate(defaultTemplate, handler)
-	s.AddResourceTemplate(branchTemplate, handler)
-	s.AddResourceTemplate(tagTemplate, handler)
-	s.AddResourceTemplate(shaTemplate, handler)
-	s.AddResourceTemplate(prTemplate, handler)
+	s.AddResourceTemplate(getRepositoryResourceContent(client, t))
```
mntlty, line 28: an alternative approach here would be to return a list and apply each one, that way changes to the resource type don't need to be remembered to be applied here. on the flip side, we would also lose some control and flexibility 🤔

pkg/github/repository_resource.go
SamMorrowDrums: We are only returning the names of things with `directoryContent`, and not the actual file data. When we are returning the file data, we actually serialise the payload.

That's the best I could come up with, the alternative would be to not have a way to inspect directory content, and just return the fact it's a directory.

It's not a perfect fit, but no way are we returning all the blobs in a folder :sweat_smile:

pkg/github/repository_resource.go
SamMorrowDrums: Yeah, I just wanted to quickly scale out what I had written to multiple endpoints - but in truth, the shared logic could be shared, with separate `opts` generation and that is likely a better direction.

tl;dr I think you're right. I just spent enough time trying to decide how to wrap it as a resource, and make it work, I didn't spend much time on how it should be written.

pkg/github/repository_resource.go
SamMorrowDrums: You are correct, and we shouldn't actually be able to hit it in practice, because the real server would falter here:

https://github.com/github/github-mcp-server/pull/69/files#diff-8c2319afb75694e93b084f504d1c1fc8de262f4301c1888c05ab067d5bf0a21bR91

So this should be one of those `if you are here, there be dragons` kind of `something went wrong` errors.

pkg/github/repository_resource_test.go
SamMorrowDrums: Hmmm, at very least I suppose we could special case markdown. I used the stdlib mime type lib, but it's perhaps not for for purpose.

pkg/github/server.go
SamMorrowDrums: Definitely something to consider

pkg/github/repository_resource.go
mntlty: went with number 2

Issue #93, opened 2025-04-03
Cross platform license generation


SamMorrowDrums:
The license library we use omits licences that are not in the build for the OS we are making it on. Attempt to use build flags for a cross-platform binary to ensure we gather distribution licence attribution for all operating system binaries.

PR #98, opened 2025-04-04
chore: support x-platform licence generation

SamMorrowDrums:
Closes: https://github.com/github/github-mcp-server/issues/93

Generate licences/attribution for all platforms we build for.

### DIFF
--- a/README.md
+++ b/README.md
@@ -284,7 +284,6 @@ export GITHUB_MCP_TOOL_ADD_ISSUE_COMMENT_DESCRIPTION="an alternative description
   - `description`: Repository description (string, optional)
   - `private`: Whether the repository is private (boolean, optional)
   - `auto_init`: Auto-initialize with README (boolean, optional)
-  - `gitignore_template`: Gitignore template name (string, optional)
 
 - **get_file_contents** - Get contents of a file or directory
 
--- a/pkg/github/issues.go
+++ b/pkg/github/issues.go
@@ -261,6 +261,9 @@ func createIssue(client *github.Client, t translations.TranslationHelperFunc) (t
 					},
 				),
 			),
+			mcp.WithNumber("milestone",
+				mcp.Description("Milestone number"),
+			),
 		),
 		func(ctx context.Context, request mcp.CallToolRequest) (*mcp.CallToolResult, error) {
 			owner, err := requiredParam[string](request, "owner")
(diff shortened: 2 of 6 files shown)


script/licenses
```diff
@@ -1,5 +1,16 @@
...
-go-licenses save ./... --save_path=third-party --force || echo "Ignore warnings"
+
+rm -rf third-party
+mkdir -p third-party
+export TEMPDIR="$(mktemp -d)"
+
+trap "rm -fr ${TEMPDIR}" EXIT
+
+for goos in linux darwin windows ; do
+    GOOS="${goos}" go-licenses save ./... --save_path="${TEMPDIR}/${goos}" --force || echo "Ignore warnings"
```
williammartin, line 17: That's nice you can just use `GOOS`

script/licenses
```diff
@@ -1,5 +1,16 @@
...
+    GOOS="${goos}" go-licenses report ./... --template .github/licenses.tmpl > third-party-licenses.${goos}.md || echo "Ignore warnings"
```
williammartin, line 18: Would appreciate a comment describing what the ignored warnings are about, and why they are ok to ignore?

script/licenses-check
```diff
@@ -1,12 +1,16 @@
...
-# if not diff -s then exit 1
 
-if ! diff -s third-party-licenses.copy.md third-party-licenses.md; then
-  echo "License check failed. Please update the license file."
-  exit 1
-fi
+for goos in linux darwin windows ; do
+    GOOS="${goos}" go-licenses report ./... --template .github/licenses.tmpl > third-party-licenses.${goos}.copy.md || echo "Ignore warnings"
+    if ! diff -s third-party-licenses.${goos}.copy.md third-party-licenses.${goos}.md; then
+        echo "License check failed. Please update the license file."
```
williammartin: So just to be clear, is the expectation that we update licenses in the same commit / PR we change dependencies in? Maybe an instruction here on how to update the license file would be useful?

SamMorrowDrums: Yeah, that's good feedback I'll add as this does get printed in CI so you will see it!

script/licenses
SamMorrowDrums: So they do need to be reviewed, but the process itself shouldn't bail out.

An example

```
W0404 12:46:16.008112  721012 library.go:101] "golang.org/x/sys/unix" contains non-Go code that can't be inspected for further dependencies:
/home/sam/go/pkg/mod/golang.org/x/sys@v0.31.0/unix/asm_linux_amd64.s
```

PR #103, opened 2025-04-04
Fix VS Code inputs link

digitarald:
Without it VS Code will show an error


williammartin (approved):
Insiders link works for me now.

PR #104, opened 2025-04-04
add vscode launch configuration

mntlty:
Adds a launch configuration for vscode to enable debugging the stdio server in standard and readonly modes


SamMorrowDrums (commented):
Does this enable you to actually pass requests to stdio or is it just to verify it builds?

Also does it need an env flag?

mntlty:
> Does this enable you to actually pass requests to stdio or is it just to verify it builds?
> 
> Also does it need an env flag?

@SamMorrowDrums here's a video of launching it, you can see me pasting to stdin on the left pane and it running the command with the `GITHUB_PERSONAL_ACCESS_TOKEN` set in the environment, while preserving the normal debugger terminal on the right side

https://github.com/user-attachments/assets/f24354c5-0947-4bfe-a8e9-3e708ffe440c

SamMorrowDrums (approved):
🔥🔥🔥

Issue #113, opened 2025-04-05
Add support for pretty printing json output


mntlty:
When running the server on stdio, the JSON output can be hard to read for humans. It would be nice to have a flag that allows for pretty printing the output.

Current behavior:
```json
GitHub MCP Server running on stdio
{"jsonrpc":"2.0","id":3,"params":{},"method":"tools/list"}
[...]

PR #114, opened 2025-04-05
pretty print json

mntlty:
- **create a config struct for running the server**
- **pretty print json**

Closes: #113

Introduces a new flag `pretty-print-json` which will format json output from the server for easier visualization.
As the list of configuration options was growing, I opted to add a config struct instead of adding one more to the pile.

(partial output in the screenshot below)
Screenshot 2025-04-04 at 5 30 41 PM

With all of this code in main, there isn't a good way to add tests here and it felt a bit out of place to add to pkg/github/server.go
A refactor would be very helpful to make this easier.

I did not add a new option to the launch.json for it, as I feel like it is still best to be disabled by default.

A future enhancement might be to incorporate https://github.com/cli/go-gh/blob/61bf393cf4aeea6d00a6251390f5f67f5b67e727/pkg/jsonpretty/format.go

### DIFF
--- a/cmd/github-mcp-server/main.go
+++ b/cmd/github-mcp-server/main.go
@@ -44,7 +44,13 @@ var (
 				stdlog.Fatal("Failed to initialize logger:", err)
 			}
 			logCommands := viper.GetBool("enable-command-logging")
-			if err := runStdioServer(readOnly, logger, logCommands, exportTranslations); err != nil {
+			cfg := runConfig{
+				readOnly:           readOnly,
+				logger:             logger,
+				logCommands:        logCommands,
+				exportTranslations: exportTranslations,
+			}
+			if err := runStdioServer(cfg); err != nil {
 				stdlog.Fatal("failed to run stdio server:", err)
 			}
 		},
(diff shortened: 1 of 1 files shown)


cmd/github-mcp-server/main.go
```diff
@@ -95,15 +107,36 @@ func initLogger(outPath string) (*log.Logger, error) {
 	return logger, nil
 }
 
-func runStdioServer(readOnly bool, logger *log.Logger, logCommands bool, exportTranslations bool) error {
+type runConfig struct {
```
toby, line 110: I'd probably just call this `config` but other than that lgtm!

toby (approved):

cmd/github-mcp-server/main.go
mntlty: going to leave it so you don't have to do another review

kwanawyly911-svg:
Hu

cmd/github-mcp-server/main.go
kwanawyly911-svg: > I'd probably just call this `config` but other than that lgtm!

Issue #112, opened 2025-04-05
create_pull_request_review tool does not support inline comments with `line` parameter


ashwin-ant:
### Describe the bug

When an inline comment is specified for the `create_pull_request_review` tool, it currently only allows comments with the `position` field specified. The [GitHub REST API endpoint for creating pull requests](https://docs.github.com/en/rest/pulls/reviews?apiVersion=2022-11-28#create-a-review-for-a-pull-request) supports `line` as an alternative to `position`. Would it be possible to support this option in the MCP tool?

### Affected version

`server version v0.1.0 (b89336793c5bc9b9abdd5100d876babbc1031f5d) 2025-04-04T15:38:21Z`

### Steps to reproduce the behavior

Try posting a pull request with a comments field like:

```
comments: [
  {
    path: 'file.md',
    line: 6,
    body: 'body text here'
  }
]
```

### Expected vs actual behavior

**Expected:** Review posts with an inline comment on line 6 of `file.md`

**Actual:** tool call fails with error `each comment must have a position`

PR #118, opened 2025-04-05
Add line parameter support to create_pull_request_review tool

ashwin-ant:
Inline comments in pull request reviews only supported `position` previously. This adds `line` support as well ([the API requires one or the other](https://docs.github.com/en/rest/pulls/reviews?apiVersion=2022-11-28#create-a-review-for-a-pull-request)).

- Updated schema to make path and body the only required fields
- Added line parameter as alternative to position for inline comments
- Updated handler to accept either position or line based on GitHub API spec
- Added new test case that verifies line parameter works properly
- Updated error messages for better validation

Closes: #112 


juruen:
@ashwin-ant thank you so much for your contribution!

We are using `position`–which is deprecated– to mimic what the [Anthropic server was doing](https://github.com/modelcontextprotocol/servers/blob/main/src/github/operations/pulls.ts#L65). However, now that the initial release is out, it's probably time to update it.

We should probably start using not only `line` but also [`side`, `start_line`, and `star_side`](https://docs.github.com/en/rest/pulls/comments?apiVersion=2022-11-28#create-a-review-comment-for-a-pull-request). This will allow for multi-line comments.

We can probably leave `position` for a while until we confirm that using the new parameters is working fine.

If you want to update this PR to fully support that, it'd be awesome. Otherwise, I can try to get it done at some point next week.

ashwin-ant:
@juruen updated! How does this look?

juruen:
> @juruen updated! How does this look?

@ashwin-ant looks great! 😺

PR #90, opened 2025-04-03
Conform list tools

williammartin:
## Description

In this PR you will find a conformance test to check this server against Anthropic's `list/tools` method. This is gated by a `conformance` build tag because it's really a point in time test that we would expect to remove later. It intentionally ignores any fields named `description` because there were too many minor differences in text.

Here is the test output:

```
--- FAIL: TestListTools (0.60s)
    conformance_test.go:123: waiting for anthropic server to start...
    conformance_test.go:123: waiting for github server to start...
    conformance_test.go:36: sending anthropic: {"jsonrpc":"2.0","id":1,"method":"tools/list","params":{}}

[...]

PR #137, opened 2025-04-06
pkg/github: fix use of per page parameter

### LINKED ISSUE #136: `list_commits` returns 30 commits despite `perPage` set to 1
### Describe the bug

`list_commits` does not interpret the `perPage` parameter correctly and returns 30 commits (default).

```py
list_commits({
  "owner": "X",
  "repo": "Y",
  "perPage": 1
})
```

Maybe it's a mismatch between https://github.com/github/github-mcp-server/blob/6f7458ae3c97f15b3e6676e7565082a1dfd98e75/pkg/github/repositories.go#L35 and https://github.com/github/github-mcp-server/blob/6f7458ae3c97f15b3e6676e7565082a1dfd98e75/pkg/github/repositories.go#L56 as a result of changes from https://github.com/github/github-mcp-server/pull/90.

### Affected version

Please run ` docker run -i --rm ghcr.io/github/github-mcp-server ./github-mcp-server --version` and paste the output below

`server version v0.1.0 (b89336793c5bc9b9abdd5100d876babbc1031f5d) 2025-04-04T15:38:21Z`

1. Agent config:

```py
    agent = Agent(
        name="Assistant",
        instructions=f"Answer questions about Git repositories.",
        mcp_servers=[mcp_server],
    )
```

2. User input:  `Summarize the last change in the repository X/Y`
3. LLM translated the input into the function call:

```py
list_commits({
  "owner": "X",
  "repo": "Y",
  "perPage": 1
})
```
4. The output from the tool call contains 30 commits, which is the default set in https://github.com/github/github-mcp-server/blob/6f7458ae3c97f15b3e6676e7565082a1dfd98e75/pkg/github/repositories.go#L56

The `perPage` parameter should be correctly respected.

### Logs

```sh
$ wc -c output.json
  179574 output.json

$ cat output.json | jq -r .text | jq '.[].sha' | wc -l
      30
```

AlexanderYastrebov:
Page size tool parameter names were changed to `perPage` within #90 while GitHub API uses `per_page` parameter name.

This change fixes overlooked inconsistencies.

Follow up on #90
Follow up on #129
Fixes #136

Closes:


AlexanderYastrebov:
Tested with code-insiders that list_commits now returns one item.

pkg/github/search.go
```diff
@@ -86,7 +86,7 @@ func searchCode(client *github.Client, t translations.TranslationHelperFunc) (to
 				mcp.Description("Sort order ('asc' or 'desc')"),
 				mcp.Enum("asc", "desc"),
 			),
-			mcp.WithNumber("per_page",
+			mcp.WithNumber("perPage",
```
williammartin: I'm not totally opposed to this, but the reason some of these existed in snake case form was because that's [what the anthropic server exposed](https://github.com/modelcontextprotocol/servers/blob/d9f2cb0b1d32acec2d9f1d8d45cf8abaf24791b6/src/github/operations/search.ts#L8), and initially we just wanted to provide parity before moving forwards.

Obviously I introduced a few silly bugs in doing that last minute 😅

williammartin:
Thanks for this, let me just run https://github.com/github/github-mcp-server/pull/137#discussion_r2030534339 by the others maintainers. I'd be in favour of it, because as seen, very easy to make mistakes.

SamMorrowDrums (dismissed):

pkg/github/search.go
AlexanderYastrebov: Hm, they actually use a mix:
* `per_page`: https://github.com/modelcontextprotocol/servers/blob/d9f2cb0b1d32acec2d9f1d8d45cf8abaf24791b6/src/github/operations/search.ts#L8
* `perPage` https://github.com/modelcontextprotocol/servers/blob/d9f2cb0b1d32acec2d9f1d8d45cf8abaf24791b6/src/github/operations/commits.ts#L9

IMO, if that's not crucial it would be better to revert all to `per_page` because that's what GitHub API uses and to reduce confusion.

williammartin: > Hm, they actually use a mix

Right, so in #90 I changed the tool schemas to match the mix (but messed up the implementation).

I think for the moment let's keep this as `perPage`, and then we can talk about broader design principles at another time, looking at the entire API surface area. There's definitely a bit of tension between consistency within github-mcp-server, with the JSON-RPC envelope (all camel cased) with the GitHub API (mostly/all snake cased?).

I talked to the team and we're happy to get this in, I'm just going to take a closer look and maybe add some tests to avoid this in future. Thanks!

williammartin:
Hey @AlexanderYastrebov, I've pushed a couple of commits on top of your branch to hopefully avoid these kind of shenanigans in the future. Please have a glance and let me know if anything looks surprising to you. Thanks for your contribution!

SamMorrowDrums (approved):
Amazing! Thanks both, this is a nice improvement.

AlexanderYastrebov:
> Please have a glance and let me know if anything looks surprising to you.

All good, thank you for the prompt review :rocket:

PR #143, opened 2025-04-06
Add tools for one-off PR comments and replying to PR review comments

ashwin-ant:
- Added two new tools for working with pull request review comments:
  - add_pull_request_review_comment: Create line, multi-line, or file-level review comments (not attached to a PR review). Corresponds to [this endpoint](https://docs.github.com/en/rest/pulls/comments?apiVersion=2022-11-28#create-a-review-comment-for-a-pull-request).
  - reply_to_pull_request_review_comment: Reply to existing PR review comments. Corresponds to [this endpoint](https://docs.github.com/en/rest/pulls/comments?apiVersion=2022-11-28#create-a-reply-for-a-review-comment).
- Verified functionality manually in MCP Insector


ashwin-ant:
@juruen mind taking a look? Would love to get this in if possible 😄

juruen:
I manually tested this and things worked as expected.

Screenshot 2025-04-11 at 11 02 35

I'd say we are good to merge!

juruen:
@ashwin-ant although what we have now seems to work, there's something else I'd like to explore here: the possibility of using a single tool for both cases, as `add_pull_request_review_comment` also supports getting an `in_reply_to`.

I've tried to reply to a comment with that tool and hasn't worked, so I'd like us to spend a bit of time on it to try to figure that out.


Thanks!

ashwin-ant:
@juruen Updated to only have a single tool. Turns out the `add_pull_request_review_comment` tool was failing when used with `in_reply_to` due to a subtlety in the underlying `go-github` library, but [there was a way to work around it](https://github.com/google/go-github/pull/950).

juruen:
> @juruen Updated to only have a single tool. Turns out the `add_pull_request_review_comment` tool was failing when used with `in_reply_to` due to a subtlety in the underlying `go-github` library, but [there was a way to work around it](https://github.com/google/go-github/pull/950).

Thank you, so much! I've manually tested the latest version and works perfectly for both cases!

PR #188, opened 2025-04-08
feat: partition tools by product/feature

SamMorrowDrums:
# Partitioned Features 

This PR proposes a method for partitioning features by product area, and provides ENV and CLI flag options to configure it on server startup.

## Features Configuration

The GitHub MCP Server supports enabling or disabling specific groups of functionalities via the `--features` flag. This allows you to control which GitHub API capabilities are available to your AI tools.

### Available Features

The following feature groups are available:

| Feature    | Description                                                   | Default Status |
| --------------- | ------------------------------------------------------------- | -------------- |
| `repos`         | Repository-related tools (file operations, branches, commits) | Enabled        |
| `issues`        | Issue-related tools (create, read, update, comment)           | Enabled        |
| `search`        | Search functionality (code, repositories, users)              | Enabled        |
| `pull_requests` | Pull request operations (create, merge, review)               | Enabled        |
| `code_security` | Code scanning alerts and security features                    | Disabled       |
| `experiments`   | Experimental features (not considered stable)                 | Disabled       |
| `everything`    | Special flag to enable all features                           | Disabled       |

### Specifying Features

You can enable specific features in two ways:

1. **Using Command Line Argument**:

   ```bash
   github-mcp-server --features repos,issues,pull_requests,code_security
   ```

2. **Using Environment Variable**:
   ```bash
   GITHUB_FEATURES="repos,issues,pull_requests,code_security" ./github-mcp-server
   ```

The environment variable `GITHUB_FEATURES` takes precedence over the command line argument if both are provided.

### Default Enabled Features

By default, the following features are enabled:

- `repos`
- `issues`
- `pull_requests`
- `search`

### Using With Docker

When using Docker, you can pass the features as environment variables:

```bash
docker run -i --rm \
  -e GITHUB_PERSONAL_ACCESS_TOKEN=<your-token> \
  -e GITHUB_FEATURES="repos,issues,pull_requests,code_security,experiments" \
  ghcr.io/github/github-mcp-server
```

### The "everything" Feature

The special feature `everything` can be provided to enable all available features regardless of their individual settings:

```bash
./github-mcp-server --features everything
```

Or using the environment variable:

```bash
GITHUB_FEATURES="everything" ./github-mcp-server
```


pkg/features/features.go
```diff
@@ -0,0 +1,68 @@
...
+}
+
+func NewFeatureSet() *FeatureSet {
+	return &FeatureSet{
+		Features:     make(map[string]Feature),
+		everythingOn: false,
+	}
+}
+
+func (fs *FeatureSet) AddFeature(name string, description string, enabled bool) {
```
mntlty: Nit: if you ever need to use https://pkg.go.dev/io/fs in here one or the other will need to be renamed

SamMorrowDrums: Great point!

toby:
I might consider calling them "toolsets" instead of features, just to be a bit more explicit.

SamMorrowDrums:
@toby toolsets does make sense, I hesitate a little because resources and prompts are also part of it, so I was sort of avoiding the word tool, but perhaps we just run with it?

Definitely not committed to any naming.

toby (approved):

toby:
This is *so* great!!

tonytrg (approved):

tonytrg:
Aside from the linter, do you plan on merging this soon?

SamMorrowDrums:
Yeah, I just have to finish a couple of bits. Mostly the auto tool updates. Just keep pushing the latest.

toby:
What do you think about these options?

1. By default we list all tools (`all`), _except_: `list_available_toolsets`, `enable_toolset` and `get_toolset_tools` since you won't really need them with all enabled.
2. Use a `--dynamic-toolsets` flag to enable `list_available_toolsets`, `enable_toolset` and `get_toolset_tools`, all other tools (except `get_me`) are hidden until enabled
3. Use the `--toolsets` flag to specify active toolsets. If used in conjunction with `--dynamic-toolsets` they will be hidden until enabled, toolsets not specified, cannot be enabled. If the `--dynamic-toolsets` flag is not set, all of the tools listed in the toolsets you've enabled are listed by default.

SamMorrowDrums:
I think those suggestions get us closer to a usable interface for sure, will riff on them on Monday I hope, see if we end up with something mergeable. Feels close.

cmd/github-mcp-server/main.go
```diff
@@ -43,12 +44,25 @@ var (
 			if err != nil {
 				stdlog.Fatal("Failed to initialize logger:", err)
 			}
+
+			enabledToolsets := viper.GetStringSlice("toolsets")
```
williammartin, line 48: **Suggestion**

If we addressed the `viper.SetEnvPrefix("APP")` debris then this would load the env var negating the need for a separate block.

pkg/github/tools.go
```diff
@@ -0,0 +1,110 @@
...
+	repos := toolsets.NewToolset("repos", "GitHub Repository related tools").
+		AddTemplateResources(
+			toolsets.NewResourceTemplate(GetRepositoryResourceContent(getClient, t)),
+			toolsets.NewResourceTemplate(GetRepositoryResourceBranchContent(getClient, t)),
+			toolsets.NewResourceTemplate(GetRepositoryResourceCommitContent(getClient, t)),
+			toolsets.NewResourceTemplate(GetRepositoryResourceTagContent(getClient, t)),
+			toolsets.NewResourceTemplate(GetRepositoryResourcePrContent(getClient, t)),
+		).
+		AddReadTools(
+			toolsets.NewServerTool(SearchRepositories(getClient, t)),
```
williammartin, line 24: **Nit**

I know this is representative of the current state but it speaks to the challenge of classifying things, this is about searching and in fact the help for searching says "Search functionality (code, repositories, users)", and yet it is under repositories, which also makes sense.

Should multiple toolsets contain the same tool?

README.md
```diff
@@ -96,6 +96,79 @@ If you don't have Docker, you can use `go` to build the binary in the
...
+
+### Default Enabled Toolsets
+
+By default, the following toolsets are enabled:
+
+- `repos`
+- `issues`
+- `pull_requests`
+- `search`
+- `context-_ools`
```
williammartin: Typo

pkg/toolsets/toolsets.go
```diff
@@ -0,0 +1,167 @@
...
+}
+
+type Toolset struct {
+	Name              string
+	Description       string
+	Enabled           bool
+	readOnly          bool
+	writeTools        []server.ServerTool
+	readTools         []server.ServerTool
+	resourceTemplates []ResourceTemplate
```
williammartin: **Feedback**

Was a bit confusing to me that toolsets also manage resource templates. I don't see a relationship between them. Might be indicative that the abstraction is trying to do too much.

pkg/github/tools.go
SamMorrowDrums: I think I should remove Search and bake it into the most relevant toolsets.

pkg/toolsets/toolsets.go
SamMorrowDrums: I'll leave it separate then, initially the idea was that all features pertaining to a product area (not just tools) would be adjusted - and the naming reflected that - but I merged it into toolsets.

For now I'll leave it separate, and it can be revisited.

ismith:
"none" also might be useful?  As in, "I do not currently want any of these tools enabled, but nor do I want to remove my config entirely from settings" (and since .vscode/mcp.json is, well, json, commenting out the entire block seems like a non-starter).

SamMorrowDrums:
Interesting idea @ismith 

> commenting out the entire block seems like a non-starter
Actually VSCode config supports jsonc, so you can use comments, and also you can de-select the tools for an entire server in the UI, if you click on the symbol.

CleanShot 2025-04-15 at 00 33 28

Some other software might not support that, but I'm tempted to say we once this merges we could track the `none` idea in an issue if you like, and see if there are others requesting the same?

pkg/github/tools.go
SamMorrowDrums: I moved the contents. Search is too vague to be useful.

Issue #193, opened 2025-04-09
Nil values for optional arrays returns error


A-Georgiou:
### Describe the bug

When attempting to use the MCP tools in Claude, experimenting with the `create_issue` tool, I found that optional arrays are passed in as 'nil'. This is not handled currently by [`server.go`](https://github.com/github/github-mcp-server/blob/56c1fce434d98cc5d4935979cc4c661a29bd0ed6/pkg/github/server.go#L234)

Example Error:

> ```
> {
>   `body`: `This is a test issue created via Claude to verify GitHub API functionality. Feel free to close it when no longer needed.`,
>   `repo`: `SmartDNN`,
>   `owner`: `A-Georgiou`,
>   `title`: `Test Issue`
> }
> Error executing code: MCP error -32603: failed to create issue: POST https://api.github.com/repos/A-Georgiou/SmartDNN/issues: 422 Invalid request.
> 
> For 'properties/assignees', nil is not an array.
> For 'properties/labels', nil is not an array. []
> ```
> 

### Affected version

Please run ` docker run -i --rm ghcr.io/github/github-mcp-server ./github-mcp-server --version` and paste the output below

`server version v0.1.0-rc.1 (87d4407d1c3dd12eacf9f134534f44d016fd363c) 2025-04-03T21:37:05Z
`
### Steps to reproduce the behavior

1. Query the Claude MCP to create a simple test issue on a local repo

### Expected vs actual behavior

A clear and concise description of what you expected to happen and what actually happened.

Expected: The MCP tool to create an issue on the respective repo
Actual: The tool call fails and the LLM attempts to recover by repeatedly calling with different parameters

A-Georgiou:
This issue might have been caused by this change to switch from strings with comma separated lists to using an actual array: https://github.com/github/github-mcp-server/pull/82

PR #194, opened 2025-04-09
Fix handling nil values for optional string array parameters,

A-Georgiou:
When attempting to use Claude for MCP tooling, any optional arrays it currently submits a nil value instead of an empty array which results in a returned error. This change adjusts that to return an empty array instead of an error for nil optional arrays.

Closes: https://github.com/github/github-mcp-server/issues/193


tonytrg (approved):
LGTM!

PR #195, opened 2025-04-09
chore: groundwork for multi-user to server

SamMorrowDrums:
In order to facilitate multi-user (MCP Host) to server communication, we need to support new `*github.Client` clients per request, rather than globally.

This change is a NOOP that makes it possible.

### DIFF
--- a/pkg/github/code_scanning.go
+++ b/pkg/github/code_scanning.go
@@ -13,7 +13,7 @@ import (
 	"github.com/mark3labs/mcp-go/server"
 )
 
-func GetCodeScanningAlert(client *github.Client, t translations.TranslationHelperFunc) (tool mcp.Tool, handler server.ToolHandlerFunc) {
+func GetCodeScanningAlert(getClient GetClientFn, t translations.TranslationHelperFunc) (tool mcp.Tool, handler server.ToolHandlerFunc) {
 	return mcp.NewTool("get_code_scanning_alert",
 			mcp.WithDescription(t("TOOL_GET_CODE_SCANNING_ALERT_DESCRIPTION", "Get details of a specific code scanning alert in a GitHub repository.")),
 			mcp.WithString("owner",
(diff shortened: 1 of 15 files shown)


pkg/github/code_scanning.go
```diff
@@ -13,7 +13,7 @@ import (
 	"github.com/mark3labs/mcp-go/server"
 )
 
-func GetCodeScanningAlert(client *github.Client, t translations.TranslationHelperFunc) (tool mcp.Tool, handler server.ToolHandlerFunc) {
+func GetCodeScanningAlert(getClient func(ctx context.Context) (*github.Client, error), t translations.TranslationHelperFunc) (tool mcp.Tool, handler server.ToolHandlerFunc) {
```
juruen: should we create a type with this function signature?

SamMorrowDrums: Yeah, I knew I should really - not hard to fix.

pkg/github/server_test.go
```diff
@@ -18,7 +18,7 @@ import (
 func Test_GetMe(t *testing.T) {
 	// Verify tool definition
 	mockClient := github.NewClient(nil)
-	tool, _ := GetMe(mockClient, translations.NullTranslationHelper)
+	tool, _ := GetMe(func(_ context.Context) (*github.Client, error) { return mockClient, nil }, translations.NullTranslationHelper)
```
williammartin: If you wouldn't mind I think a lot of noise could go away with:

```go
func stubGetClientFn(client *github.Client) GetClientFn {
	return func(_ context.Context) (*github.Client, error) {
		return client, nil
	}
}

func Test_GetMe(t *testing.T) {
	// Verify tool definition
	mockClient := github.NewClient(nil)
	tool, _ := GetMe(stubGetClientFn(mockClient), translations.NullTranslationHelper)
...
```

Or perhaps:

```go
func stubbedGetClientFn() GetClientFn {
	return func(_ context.Context) (*github.Client, error) {
		return github.NewClient(nil), nil
	}
}

func Test_GetMe(t *testing.T) {
	// Verify tool definition
	tool, _ := GetMe(stubbedGetClientFn(), translations.NullTranslationHelper)
...
```

Slight preference for the former because it's clearer how the client is being injected.

williammartin (approved):
Approved but minor request. Happy for you to bypass under the assumption you just address this if you make any changes!

pkg/github/server_test.go
williammartin: Made minor edit to names above, just in case you'd already read it.

PR #211, opened 2025-04-10
Added get_issue_events , get_issue_timeline and get_issue_event tools

artemsaveliev:
Implements 3 tools dealing with issue events and timeline:
- get_issue_events - all events for the issue https://docs.github.com/en/rest/issues/events?apiVersion=2022-11-28#list-issue-events-for-a-repository
- get_issue_timeline - issue events timeline https://docs.github.com/en/rest/issues/timeline
- get_issue_event - issue event details https://docs.github.com/en/rest/issues/events?apiVersion=2022-11-28#get-an-issue-event


pkg/github/issues.go
```diff
@@ -711,6 +711,218 @@ func GetIssueComments(getClient GetClientFn, t translations.TranslationHelperFun
...
+				mcp.Required(),
+				mcp.Description("Repository name"),
+			),
+			mcp.WithNumber("issue_number",
+				mcp.Required(),
+				mcp.Description("Issue number"),
+			),
+			mcp.WithNumber("page",
+				mcp.Description("Page number"),
+			),
```
SamMorrowDrums: I believe you can use WithPagination here.

https://github.com/github/github-mcp-server/blob/50043c32de657b4fc17c6860fec30b704ce61a36/pkg/github/server.go#L178

pkg/github/issues.go
```diff
@@ -711,6 +711,218 @@ func GetIssueComments(getClient GetClientFn, t translations.TranslationHelperFun
 		}
 }
 
+// GetIssueTimeline creates a tool to get timeline for a GitHub issue.
+func GetIssueTimeline(getClient GetClientFn, t translations.TranslationHelperFunc) (tool mcp.Tool, handler server.ToolHandlerFunc) {
+	return mcp.NewTool("get_issue_timeline",
+			mcp.WithDescription(t("TOOL_GET_ISSUE_TIMELINE_DESCRIPTION", "Get timeline for a GitHub issue")),
```
SamMorrowDrums: All tools need Annotations now: https://github.com/github/github-mcp-server/blob/eca853b28a0d1d3567ea1958647708dc0184880d/pkg/github/pullrequests.go#L22

It shows the end user a clean name, and it also means the agent can run read-only commands without prompting for consent.

SamMorrowDrums (dismissed):
I think with the pagination helper and the annotations this is good to merge. The code is great, and I'm curious to see what models do with the tools!

Thanks again @artemsaveliev

pkg/github/issues.go
artemsaveliev: Added

pkg/github/issues.go
artemsaveliev: Added

SamMorrowDrums:
Having lots of fun

image

image


Hmm not convinced it was right: 

image

SamMorrowDrums:
@artemsaveliev as an update, I have been experimenting - and I have played with tool descriptions and things, and it works for the basic cases, but for more complex cases the models do tend to get stuck and then just keep repeating the same requests ad-infinitum!

artemsaveliev:
@SamMorrowDrums The getting-stuck-in-a-loop, and more importantly running out of time/context window is more specific to VSCode Agent implementation - it does not work great with large amounts of data. 

Specifically, to me the most interesting use case that the timeline tool covers is discovering cross-reference timeline events - this will allow agent to traverse entire set of issues, as long as they contain links to each other. 
https://docs.github.com/en/rest/using-the-rest-api/issue-event-types?apiVersion=2022-11-28#cross-referenced
In my testing, the agent gave up after about a minute of navigating links - but again, this is specific to VSCode agent implementation. Other agents, especially the ones that have larger context and/or implementation of "memory", should be able to navigate better.

Update: There is a setting in vscode `chat.agent.maxRequests` that controls how long agent can spend querying tools - by default it's 25. Increasing this number lets agent run in loop for potentially very long time, so it can take advantage of all cross-references between issues and navigate them, for example.

galshi:
Hi, any updates on when will this be merged? The timeline tool looks great!

tommaso-moro:
Update - issues tracking this:
- [Add tools to get issue_events and get_issue_event #984
](https://github.com/github/github-mcp-server/issues/984)
- [Add tool to get_issue_timeline #985
](https://github.com/github/github-mcp-server/issues/985)

jakemcallister:
Hi - any news when this will get merged?

eanagel-joby:
This would sure help me out and I'm sure a bunch of other people. It's been here for well over a year now. Is there an issue preventing it from getting merged?

louim:
@SamMorrowDrums would you merge an up to date version of this branch? I can submit a PR

SamMorrowDrums:
@louim yes, if it's behind a feature flag and @artemsaveliev is given attribution also. 🙏

I had been meaning to do this myself, so thanks for suggesting.

PR #218, opened 2025-04-11
Allow passing through server options

omgitsads:
Adds support the the underlying mcp-go `ServerOptions` so that [hook functions](https://github.com/mark3labs/mcp-go/blob/main/server/hooks.go#L84-L107) can be attached to the server.


SamMorrowDrums (approved):
Nice

PR #222, opened 2025-04-11
Minor additions to gitignore and readme

tonytrg:
This pr adds:
- excludes a few paths for go binaries
- adds a bit more context for the `Prerequisites` section in the readme.


README.md
```diff
@@ -18,7 +18,8 @@ automation and interaction capabilities for developers and tools.
 2. Once Docker is installed, you will also need to ensure Docker is running.
 3. Lastly you will need to [Create a GitHub Personal Access Token](https://github.com/settings/personal-access-tokens/new).
 The MCP server can use many of the GitHub APIs, so enable the permissions that you feel comfortable granting your AI tools (to learn more about access tokens, please check out the [documentation](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/managing-your-personal-access-tokens)).
-
+    * Your PAT Token needs at least `read:packages` permission to be able to pull from `ghcr.io`.
```
williammartin: Did you run into some issue here? As far as I'm aware the image is public and:

> You can also access public container images anonymously.

https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry#about-the-container-registry

tonytrg: ah whoops ignore this pr for now :)

I just tested and it works without the pat.

I had some issues with the setup earlier. So I wiped everything and go through every option in the readme to check if I can add useful stuff to the readme.

Issue #215, opened 2025-04-10
Publish server as a GitHub cli extension


gillisandrew:
### Describe the feature or problem you’d like to solve

Provide a simplified installation and configuration experience to users with the GitHub cli already installed and configured.

### Proposed solution

Publish the stdio server as a github extension that reuses the cli's token to authorize api actions. Installation instructions would be something like:

1. Run `gh extension install github-mcp-server`
2. Add the following in VS Code User Settings (JSON):
```json
{
  "mcp": {
    "servers": {
      "github": {
        "command": "gh",
        "args": [
          "github-mcp-server",
          "stdio"
        ]
      }
    }
  }
}
```

### Additional context

https://docs.github.com/en/github-cli/github-cli/creating-github-cli-extensions

williammartin:
We're already looking into doing this, so thanks for the feedback on the direction!

gillisandrew:
@williammartin Great happy to add to the discussion. Piggybacking off of the CLI's OAuth token would be particularly interesting since provisioning PATs with long lifetimes is a non-starter for many enterprise users. Thanks for looking into it!

MayorFaj:
@williammartin  I have built something for this already, how do we make it available for users

williammartin:
Happy for you to share it here and we can talk about it, thanks. However, it's likely that if we did ship an extension, we would want to take ownership over the maintenance and release cycle.

MayorFaj:
@williammartin would you rather I raise a PR for that here, or a paste here a repo link containing the code?

williammartin:
Go ahead and link it here for the moment, thank you.

MayorFaj:
@williammartin Here you go  [gh-github-mcp-server](https://github.com/MayorFaj/gh-github-mcp-server)


![Image](https://github.com/user-attachments/assets/1cc13cdf-5615-44ff-bb9d-2f8e9f19852d)

williammartin:
Thanks @MayorFaj, I think the decision we need to make here is how to ship the extension so that it is installable on various supported platforms. This requires having the executable artifacts attached to a release, which means we either:
 * Pull in built artifacts from this repository
 * Export the application logic in a way that can be called from a `main.go` in `gh-github-mcp-server`, and use `gh-extension-precompile` there
 * Relax some requirements in `gh` (e.g. `gh` prefix) so that extensions can be installed directly from this repository

With each of those we then need to make a decision on how best to do releases of the extension.

[...]

Issue #199, opened 2025-04-09
Update readme to include docker-less instructions


alexlwn123:
### Describe the feature or problem you’d like to solve

From the readme, it's not obvious how this can be run without docker.

### Proposed solution

The docker daemon has large memory requirements. Naturally, this is problematic for lower-end devices or for folks who need other heavy dev infrastructure running locally.

Thinking the readme should suggest an 1st class option to run without docker. 

### Additional context

Are the instructions on the old published npm package still valid? It includes run with `npx` instructions. [here](https://www.npmjs.com/package/@modelcontextprotocol/server-github)

StefMa:
The old instructions seems to be not valid anymore. Those were published and maintained by anthropic. Github decided to use go instead of Javascript/typescript.

Anyways I opened a PR [here](https://github.com/github/github-mcp-server/pull/224) to clarify how to use the project without docker.

I haven't tested it, maybe someone wants to try it out and report back how it works 🙃

tonytrg:
https://github.com/github/github-mcp-server/pull/264 covers this

PR #224, opened 2025-04-11
Update build from source instructions

StefMa:
I just updated the description of the `build from source` section to make them more clear/obvious.

I just thought it make sense because something like this was already requested
https://github.com/github/github-mcp-server/issues/215 which is IMHO more complicate to do than just use plain old `go`.

There is also this issues (https://github.com/github/github-mcp-server/issues/199) which ask for "Dockerless" solutions.

To me it seems that people needs some more help or guidance.

Let me know what do you think and if I should change something.

P.s.:
I haven't tested it
Just added what I think should work 🙃

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,7 +112,7 @@ The MCP servers from VS Code and Claude needs to be adjusted respectively:
 +  "stdio"
   ],
   "env": {
-   "GITHUB_PERSONAL_ACCESS_TOKEN": "<YOUR_TOKEN>"
+   "GITHUB_PERSONAL_ACCESS_TOKEN": "${input:github_token} OR <YOUR_TOKEN>"
   }
 }
 ```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -92,9 +92,30 @@ More about using MCP server tools in VS Code's [agent mode documentation](https:
...
+-  "run",
+-  "-i",
+-  "--rm",
+-  "-e",
+-  "GITHUB_PERSONAL_ACCESS_TOKEN",
+-  "ghcr.io/github/github-mcp-server"
++  "stdio"
+  ],
+  "env": {
+   "GITHUB_PERSONAL_ACCESS_TOKEN": "${input:github_token} OR <YOUR_TOKEN>"
```
gillisandrew, line 115: Users shouldn't be hardcoding PATs in a config file.

StefMa: I just copied it from the Claude Desktop section. What should it be instead? 🤔 And should I also update it in the Claude Desktop section?

gillisandrew: Oh fair enough, vscode supports password inputs which prompts for input when the server is launched. Claude doesn't appear to have that feature.

toby:
We ended up going with #264 to clarify source install. Thank you for this though!

PR #225, opened 2025-04-11
feat: add GitHub notifications tools for managing user notifications

sridharavinash:
# GitHub Notifications Tooling

This PR adds support for managing GitHub notifications 

### New Tools
- `get_notifications`: Retrieve a list of notifications for the authenticated GitHub user with filtering options for read/unread status, participation, and time ranges
- `mark_notification_read`: Mark a specific notification thread as read
- `mark_notification_done`: Mark a specific notification thread as done
- `mark_all_notifications_read`: Mark all notifications as read with an optional timestamp
- `get_notification_thread`: Fetch details of a specific notification thread
### Technical Implementation
- Implements proper error handling and status code validation
- Provides RFC3339/ISO8601 timestamp parsing for time-related parameters
- Integrates with the existing GitHub API client infrastructure
- Respects read-only mode settings for mutating operations

### DIFF
--- a/e2e/e2e_test.go
+++ b/e2e/e2e_test.go
@@ -62,7 +62,8 @@ func getRESTClient(t *testing.T) *gogithub.Client {
 
 	// Create a new GitHub client with the token
 	ghClient := gogithub.NewClient(nil).WithAuthToken(token)
-	if host := getE2EHost(); host != "https://github.com" {
+
+	if host := getE2EHost(); host != "" && host != "https://github.com" {
 		var err error
 		// Currently this works for GHEC because the API is exposed at the api subdomain and the path prefix
 		// but it would be preferable to extract the host parsing from the main server logic, and use it here.
--- a/pkg/github/notifications.go
+++ b/pkg/github/notifications.go
@@ -1,0 +1,300 @@
+package github
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"strconv"
+	"time"
+
+	"github.com/github/github-mcp-server/pkg/translations"
+	"github.com/google/go-github/v69/github"
+	"github.com/mark3labs/mcp-go/mcp"
+	"github.com/mark3labs/mcp-go/server"
+)
+
+// getNotifications creates a tool to list notifications for the current user.
+func GetNotifications(getClient GetClientFn, t translations.TranslationHelperFunc) (tool mcp.Tool, handler server.ToolHandlerFunc) {
+	return mcp.NewTool("get_notifications",
+			mcp.WithDescription(t("TOOL_GET_NOTIFICATIONS_DESCRIPTION", "Get notifications for the authenticated GitHub user")),
+			mcp.WithBoolean("all",
+				mcp.Description("If true, show notifications marked as read. Default: false"),
+			),
+			mcp.WithBoolean("participating",
+				mcp.Description("If true, only shows notifications in which the user is directly participating or mentioned. Default: false"),
+			),
+			mcp.WithString("since",
+				mcp.Description("Only show notifications updated after the given time (ISO 8601 format)"),
+			),
+			mcp.WithString("before",
+				mcp.Description("Only show notifications updated before the given time (ISO 8601 format)"),
+			),
+			mcp.WithNumber("per_page",
+				mcp.Description("Results per page (max 100). Default: 30"),
+			),
+			mcp.WithNumber("page",
+				mcp.Description("Page number of the results to fetch. Default: 1"),
+			),
+		),
+		func(ctx context.Context, request mcp.CallToolRequest) (*mcp.CallToolResult, error) {
+			client, err := getClient(ctx)
+			if err != nil {
+				return nil, fmt.Errorf("failed to get GitHub client: %w", err)
+			}
+
+			// Extract optional parameters with defaults
+			all, err := OptionalParamWithDefault[bool](request, "all", false)
+			if err != nil {
+				return mcp.NewToolResultError(err.Error()), nil
+			}
+
+			participating, err := OptionalParamWithDefault[bool](request, "participating", false)
+			if err != nil {
+				return mcp.NewToolResultError(err.Error()), nil
+			}
+
+			since, err := OptionalParam[string](request, "since")
+			if err != nil {
+				return mcp.NewToolResultError(err.Error()), nil
+			}
+
+			before, err := OptionalParam[string](request, "before")
+			if err != nil {
+				return mcp.NewToolResultError(err.Error()), nil
+			}
+
+			perPage, err := OptionalIntParamWithDefault(request, "per_page", 30)
+			if err != nil {
+				return mcp.NewToolResultError(err.Error()), nil
+			}
+
+			page, err := OptionalIntParamWithDefault(request, "page", 1)
+			if err != nil {
+				return mcp.NewToolResultError(err.Error()), nil
+			}
+
+			// Build options
+			opts := &github.NotificationListOptions{
+				All:           all,
+				Participating: participating,
+				ListOptions: github.ListOptions{
+					Page:    page,
+					PerPage: perPage,
+				},
+			}
+
+			// Parse time parameters if provided
+			if since != "" {
+				sinceTime, err := time.Parse(time.RFC3339, since)
+				if err != nil {
+					return mcp.NewToolResultError(fmt.Sprintf("invalid since time format, should be RFC3339/ISO8601: %v", err)), nil
+				}
+				opts.Since = sinceTime
+			}
+
+			if before != "" {
+				beforeTime, err := time.Parse(time.RFC3339, before)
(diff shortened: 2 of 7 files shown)


e2e/e2e_test.go
```diff
@@ -1525,3 +1536,298 @@ func TestPullRequestReviewDeletion(t *testing.T) {
...
+	require.NoError(t, err)
+	require.False(t, resp.IsError)
+	textContent, ok = resp.Content[0].(mcp.TextContent)
+	require.True(t, ok)
+	require.Contains(t, textContent.Text, "deleted")
+
+	// Validate with REST client
+	sub, resp2, err := restClient.Activity.GetThreadSubscription(ctx, notificationID)
+	// According to GitHub API, a deleted subscription returns 404
+	require.Error(t, err)
```
williammartin: **Concern**

This test is failing for me on this line. Even with a pause in the debugger to take a moment, there is a subscription successfully returned:

```
*github.Subscription {Subscribed: *false, Ignored: *true, Reason: *"", CreatedAt: *github.com/google/go-github/v69/github.Timestamp {Time: (*time.Time)(0x1400022e810)}, URL: *"https://api.github.com/notifications/threads/16530071814/subscription", RepositoryURL: *string nil, ThreadURL: *"https://api.github.com/notifications/threads/16530071814"}
```

This is despite the `DeleteThreadSubscription` call in the tool returning 204 as expected:

```
*github.Response {Response: *net/http.Response {Status: "204 No Content", StatusCode: 204, Proto: "HTTP/2.0", ProtoMajor: 2, ProtoMinor: 0, Header: net/http.Header [...], Body: io.ReadCloser(net/http.http2noBodyReader) *(*io.ReadCloser)(0x1400015fbd0), ContentLength: 0, TransferEncoding: []string len: 0, cap: 0, nil, Close: false, Uncompressed: false, Trailer: net/http.Header nil, Request: *(*"net/http.Request")(0x140001f7e00), TLS: *(*"crypto/tls.ConnectionState")(0x140000dc840)}, NextPage: 0, PrevPage: 0, FirstPage: 0, LastPage: 0, NextPageToken: "", Cursor: "", Before: "", After: "", Rate: github.com/google/go-github/v69/github.Rate {Limit: 15000, Remaining: 14982, Used: 18, Reset: (*"github.com/google/go-github/v69/github.Timestamp")(0x1400014c440), Resource: "core"}, TokenExpiration: github.com/google/go-github/v69/github.Timestamp {Time: (*time.Time)(0x1400014c468)}}
```

Where do you see the deleted subscription should return 404 in the API docs: https://docs.github.com/en/rest/activity/notifications?apiVersion=2022-11-28#get-a-thread-subscription-for-the-authenticated-user, it seems to me that we have a "subscription" but the `Subscribed` field is false?

e2e/e2e_test.go
```diff
@@ -1525,3 +1536,298 @@ func TestPullRequestReviewDeletion(t *testing.T) {
...
+	require.NoError(t, err)
+	require.False(t, resp.IsError)
+	textContent, ok = resp.Content[0].(mcp.TextContent)
+	require.True(t, ok)
+	require.Contains(t, textContent.Text, "deleted")
+
+	// Validate with REST client
+	sub, resp2, err := restClient.Activity.GetRepositorySubscription(ctx, owner, repo)
+	// According to GitHub API, a deleted subscription returns 404
+	require.Error(t, err)
```
williammartin: Have the same issue here as above.

e2e/e2e_test.go
```diff
@@ -1525,3 +1536,298 @@ func TestPullRequestReviewDeletion(t *testing.T) {
...
+	require.NoError(t, err)
+	require.False(t, resp.IsError)
+	textContent, ok := resp.Content[0].(mcp.TextContent)
+	require.True(t, ok)
+	var notifications []struct {
+		ID string `json:"id"`
+	}
+	err = json.Unmarshal([]byte(textContent.Text), &notifications)
+	require.NoError(t, err)
+	require.NotEmpty(t, notifications)
```
williammartin: I think this probably needs:

```
	if len(notifications) == 0 {
		t.Skip("No notifications available to test dismissal")
	}
```

The test failed for me in a surprising way because I had no notifications.

e2e/e2e_test.go
```diff
@@ -1525,3 +1536,298 @@ func TestPullRequestReviewDeletion(t *testing.T) {
 	require.NoError(t, err, "expected to unmarshal text content successfully")
 	require.Len(t, noReviews, 0, "expected to find no reviews")
 }
+
+func TestE2E_ListNotifications(t *testing.T) {
```
williammartin: **Question**

Is there a reason you prefixed all of these `TestE2E`? They are in the `e2e` package so it seems a little redundant and out of sync with the rest of the existing tests.

pkg/github/notifications_test.go
```diff
@@ -0,0 +1,281 @@
...
+	"testing"
+
+	"github.com/github/github-mcp-server/pkg/translations"
+	"github.com/google/go-github/v69/github"
+	"github.com/migueleliasweb/go-github-mock/src/mock"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func Test_ListNotifications(t *testing.T) {
```
williammartin, line 16: **Request**

Please use the same testing format as the rest of the other tool handler tests, or if you feel strongly, document why this is a different pattern that you are proceeding with here.

pkg/github/notifications.go
```diff
@@ -0,0 +1,498 @@
...
+					return mcp.NewToolResultError(fmt.Sprintf("invalid before time format, should be RFC3339/ISO8601: %v", err)), nil
+				}
+				opts.Before = beforeTime
+			}
+
+			var notifications []*github.Notification
+			var resp *github.Response
+
+			if owner != "" && repo != "" {
+				notifications, resp, err = client.Activity.ListRepositoryNotifications(ctx, owner, repo, opts)
```
williammartin, line 116: **Request**

I think we should probably have at least a unit test for this branch which is currently uncovered.

pkg/github/notifications.go
```diff
@@ -0,0 +1,498 @@
...
+			var resp *github.Response
+
+			if state == "done" {
+				// for some inexplicable reason, the API seems to have threadID as int64 and string depending on the endpoint
+				var threadIDInt int64
+				threadIDInt, err = strconv.ParseInt(threadID, 10, 64)
+				if err != nil {
+					return mcp.NewToolResultError(fmt.Sprintf("invalid threadID format: %v", err)), nil
+				}
+				resp, err = client.Activity.MarkThreadDone(ctx, threadIDInt)
```
williammartin, line 182: **Request**

I think we should probably have at least a unit test for this branch which is currently uncovered.

pkg/github/notifications.go
```diff
@@ -0,0 +1,498 @@
...
+				lastReadTime = time.Now()
+			}
+
+			markReadOptions := github.Timestamp{
+				Time: lastReadTime,
+			}
+
+			var resp *github.Response
+			if owner != "" && repo != "" {
+				resp, err = client.Activity.MarkRepositoryNotificationsRead(ctx, owner, repo, markReadOptions)
```
williammartin, line 260: **Request**

I think we should probably have at least a unit test for this branch which is currently uncovered.

pkg/github/notifications.go
```diff
@@ -0,0 +1,498 @@
...
+				apiErr error
+			)
+
+			switch action {
+			case NotificationActionIgnore:
+				sub := &github.Subscription{Ignored: toBoolPtr(true)}
+				result, resp, apiErr = client.Activity.SetThreadSubscription(ctx, notificationID, sub)
+			case NotificationActionWatch:
+				sub := &github.Subscription{Ignored: toBoolPtr(false), Subscribed: toBoolPtr(true)}
+				result, resp, apiErr = client.Activity.SetThreadSubscription(ctx, notificationID, sub)
```
williammartin, line 380: **Request**

I think we should probably have at least a unit test for this branch which is currently uncovered.

pkg/github/notifications.go
```diff
@@ -0,0 +1,498 @@
...
+				apiErr error
+			)
+
+			switch action {
+			case RepositorySubscriptionActionIgnore:
+				sub := &github.Subscription{Ignored: toBoolPtr(true)}
+				result, resp, apiErr = client.Activity.SetRepositorySubscription(ctx, owner, repo, sub)
+			case RepositorySubscriptionActionWatch:
+				sub := &github.Subscription{Ignored: toBoolPtr(false), Subscribed: toBoolPtr(true)}
+				result, resp, apiErr = client.Activity.SetRepositorySubscription(ctx, owner, repo, sub)
```
williammartin, line 469: **Request**

I think we should probably have at least a unit test for this branch which is currently uncovered.

williammartin (changes requested):

williammartin (approved):
Thanks!

PR #231, opened 2025-04-11
feat: use existing GitHub CLI token as a fallback

gillisandrew:
This pull request adds support for using the existing GitHub CLI OAuth token for authorizing API requests. This shouldn't be a breaking change, it simply modifies the behavior when the `GITHUB_PERSONAL_ACCESS_TOKEN` environment variable is unset. Instead of erroring it will check if the specified host has already been authenticated in the CLI.

This relates to #215


cmd/github-mcp-server/main.go
```diff
@@ -113,34 +114,29 @@ func runStdioServer(cfg runConfig) error {
...
-	if host != "" {
-		var err error
-		ghClient, err = ghClient.WithEnterpriseURLs(host, host)
-		if err != nil {
-			return fmt.Errorf("failed to create GitHub client with host: %w", err)
-		}
+	// Create GitHub API client
+	httpClient, err := api.NewHTTPClient(api.ClientOptions{
+		Host:      host,
+		AuthToken: os.Getenv("GITHUB_PERSONAL_ACCESS_TOKEN"),
```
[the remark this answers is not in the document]
gillisandrew: In this case, the fallback logic comes from `github.com/cli/go-gh/v2/pkg/api` which will retrieve any existing token from the cli's cache.

cmd/github-mcp-server/main.go
```diff
@@ -113,34 +114,30 @@ func runStdioServer(cfg runConfig) error {
...
 	}
 
-	if host != "" {
-		var err error
-		ghClient, err = ghClient.WithEnterpriseURLs(host, host)
-		if err != nil {
-			return fmt.Errorf("failed to create GitHub client with host: %w", err)
-		}
+	// Create GitHub API client
+	httpClient, err := api.NewHTTPClient(api.ClientOptions{
```
williammartin: I'm not totally sure about bringing in the entire api client here, since it comes with some behaviour that may or may not be desired, and it might be easier to defer that to another feature request.

I think a more direct solution to the problem here would be using `auth.TokenForHost(host)` and providing that to the `gogithub.NewClient`.

That said, need to speak to the team about this feature, I have some minor concerns about the clarity of the chosen token, since it's no longer explicitly provided. I don't want people to be surprised that the LLM is suddenly working resources that they didn't expect and didn't want. Thanks!

SamMorrowDrums: My hot take on the above is to agree with @williammartin and its the one change I would want to make if we shipped this as a cli extension, where the implicit token and paved path for overriding the token with ENV at the gh level makes more sense.

Also, in Docker it will never have access, so the documentation would be confusing to users unless there was a clear instruction like 

> you can authenticate automatically via the gh cli by installing the MCP server as as an extension.

> __instructions on how to do so__

gillisandrew: @williammartin Oh of course, this got things working for my liking but I can see why we wouldn't necessarily want all the features of the client.

The behavior could be placed behind a flag to be more transparent to the user, though for a gh extension release it should be the default behavior IMO.

pauldthomson:
Wondering if this should be leveraging `cli/go-gh` like it is or potentially instances of the MCP server being their own OAuth applications i.e. having their own client ID which could be implemented using [this](https://github.com/cli/oauth)

gillisandrew:
@pauldthomson Definitely agree, there should be some tooling provided by `github-mcp-server` to acquire the token so users aren't stuck generating PATs.

There is a broader discussion to be had about how fine-grained authorization could be handled by the server, given many users want to grant varying levels of access to different resources (e.g. read-write access to current repo, but read-only access to other repos). It could go a long way to addressing the perennial [claims of github-mcp-server being "exploited"](https://invariantlabs.ai/blog/mcp-github-vulnerability) by poisoned prompts to extract private data.

raj4036:
> This PR adds support for using the GitHub CLI's token

jojkoloski:
X

PR #237, opened 2025-04-12
chore: extend user agent

SamMorrowDrums:
This gets you things like: `github-mcp-server/<version> (Visual Studio Code - Insiders/<version>)` if provided by the host.

### DIFF
--- a/cmd/github-mcp-server/main.go
+++ b/cmd/github-mcp-server/main.go
@@ -13,6 +13,7 @@ import (
 	iolog "github.com/github/github-mcp-server/pkg/log"
 	"github.com/github/github-mcp-server/pkg/translations"
 	gogithub "github.com/google/go-github/v69/github"
+	"github.com/mark3labs/mcp-go/mcp"
 	"github.com/mark3labs/mcp-go/server"
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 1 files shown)


cmd/github-mcp-server/main.go
```diff
@@ -137,11 +138,19 @@ func runStdioServer(cfg runConfig) error {
 
 	t, dumpTranslations := translations.TranslationHelper()
 
+	beforeInit := func(_ context.Context, _ any, message *mcp.InitializeRequest) {
+		ghClient.UserAgent = fmt.Sprintf("github-mcp-server/%s (%s/%s)", version, message.Params.ClientInfo.Name, message.Params.ClientInfo.Version)
```
juruen, line 142: Haven't looked into how hooks work in depth, so bear with me, one question that I have is whether we need to protect this with a mutex.

juruen: Disregard my comment. Even we wanted to protect it with a mutex, we couldn't because this is in the github.Client.

juruen: So, if I understand this correctly. This is going to be OK from a race condition standpoint if host/client always initializes the server before doing anything else.

The only way to cause a race here is by calling a tool and initializing at the same time.

juruen: If you wanted to completely remove the possibility of a race, you could make things slightly more complex and set the client info into a different variable protected by a mutex. And then, you could set the client agent accordingly.

SamMorrowDrums: Yeah, that's a good shout. I was thinking about this too, but it is only a one-shot event, and if it happens a second time it can also update the info, that's fine.

PR #261, opened 2025-04-13
feat: implement get_repository_discussions tool with GraphQL support

sridharavinash:
Closes:

### DIFF
--- a/pkg/github/discussions.go
+++ b/pkg/github/discussions.go
@@ -1,0 +1,199 @@
+package github
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+
+	"github.com/github/github-mcp-server/pkg/translations"
+	"github.com/mark3labs/mcp-go/mcp"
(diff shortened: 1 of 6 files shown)


pkg/github/discussions.go
```diff
@@ -0,0 +1,260 @@
...
+	Body         string    `json:"body"`
+	CreatedAt    string    `json:"createdAt"`
+	UpdatedAt    string    `json:"updatedAt"`
+	URL          string    `json:"url"`
+	Category     string    `json:"category"`
+	Author       string    `json:"author"`
+	Locked       bool      `json:"locked"`
+	UpvoteCount  int       `json:"upvoteCount"`
+	CommentCount int       `json:"commentCount"`
+	Comments     []Comment `json:"comments,omitempty"`
```
rociomontes, line 36: Really glad we would be able to also have access to the comments

dsyme:
My understanding is that this is outdated

Issue #268, opened 2025-04-14
Add support for Actions APIs


joshmgross:
### Describe the feature or problem you’d like to solve

GitHub Actions has a suite of APIs that users may want to use with their AI-powered tools.

### Proposed solution

We add tools for Actions, starting with some of the core APIs like:

1. Listing workflows
2. Listing workflow runs
3. Triggering a workflow (via `workflow_dispatch`)
4. Re-running workflows
5. Downloading logs for a run

### Additional context

https://docs.github.com/rest/actions

jantmanAtCox:
This is extremely important for projects that have CI/CD systems that aren't feasible to run locally, where the only way for the agent to get feedback on changes is to be able to access workflow run status and logs.

DexterPOSH:
really like this proposal, at work we were in the midst of adding these tools for GH Actions API, it'd be great if the GH MCP server supports this out of the box.

I am happy to contribute as well to add this support. I already see a PR from @joshmgross for running a workflow 👍🏼

PR #269, opened 2025-04-14
Add support for running an Actions workflow

joshmgross:
Supports #268

This adds a tool to allow running an Actions workflow via the [`workflow_dispatch` API](https://docs.github.com/rest/actions/workflows?apiVersion=2022-11-28#create-a-workflow-dispatch-event).

**Demo**

https://github.com/user-attachments/assets/55d2de6b-a916-41a3-bcb6-26cec6ed44d1

(Tested using these changes built from source per the [instructions](https://github.com/github/github-mcp-server?tab=readme-ov-file#build-from-source))

MCP Triggered Run

https://github.com/joshmgross/actions-testing/actions/runs/14454560690/job/40534749404


pkg/github/actions.go
```diff
@@ -0,0 +1,99 @@
+package github
```
joshmgross, line 1: There are a *lot* of Actions APIs, I suspect the single-file per API area will break down at some point.

I tried defining a separate `actions` package, but it created some circular dependencies with this `github` package.
I'd be happy to explore that more, but I wanted to keep these changes focused on this tool.

pkg/github/actions.go
```diff
@@ -0,0 +1,99 @@
...
+				inputs, _ = inputsObj.(map[string]any)
+			}
+
+			// Convert inputs to the format expected by the GitHub API
+			inputsMap := make(map[string]any)
+			if inputs != nil {
+				for k, v := range inputs {
+					inputsMap[k] = v
+				}
+			}
```
joshmgross: This was mostly generated by Copilot, not sure if there's a better way to process an object parameter.

pkg/github/actions.go
```diff
@@ -0,0 +1,97 @@
...
+func RunWorkflow(getClient GetClientFn, t translations.TranslationHelperFunc) (tool mcp.Tool, handler server.ToolHandlerFunc) {
+	return mcp.NewTool("run_workflow",
+			mcp.WithDescription(t("TOOL_RUN_WORKFLOW_DESCRIPTION", "Trigger a workflow run")),
+			mcp.WithString("owner",
+				mcp.Required(),
+				mcp.Description("The account owner of the repository. The name is not case sensitive."),
+			),
+			mcp.WithString("repo",
+				mcp.Required(),
+				mcp.Description("Repository name"),
```
jww3, line 24: No mention of case-sensitivity here.  Does that imply that it is case-sensitive?  Should we be explicit?

pkg/github/actions.go
```diff
@@ -0,0 +1,97 @@
...
+			),
+			mcp.WithString("workflowId",
+				mcp.Required(),
+				mcp.Description("The ID of the workflow. You can also pass the workflow file name as a string."),
```
jww3: Just the workflow file name itself is acceptable?  Or the relative path to the workflow file?

pkg/github/actions.go
joshmgross: This matches the description used by most/all of the other `repo` MCP inputs in this repo, so I stuck to being consistent with them.

I was thinking it might make sense to define some constant/reusable input definitions for these common inputs like repo & owner for consistency.

pkg/github/actions.go
joshmgross: The API supports both

Relative path
```shell
gh api -X POST /repos/joshmgross/actions-testing/actions/workflows/hello-world.yaml/dispatches -f "ref=main"
```

Full path (note the `/` must be escaped)
```shell
gh api -X POST /repos/joshmgross/actions-testing/actions/workflows/.github%2Fworkflows%2Fhello-world.yaml/dispatches -f "ref=main"
```


I don't think `github/go-github` will escape the `/` by default though, so we might need to handle that 🤔 

```go
func (s *ActionsService) CreateWorkflowDispatchEventByFileName(ctx context.Context, owner, repo, workflowFileName string, event CreateWorkflowDispatchEventRequest) (*Response, error) {
	u := fmt.Sprintf("repos/%v/%v/actions/workflows/%v/dispatches", owner, repo, workflowFileName)

	return s.createWorkflowDispatchEvent(ctx, u, &event)
}
```
https://github.com/google/go-github/blob/6f8bceff0b000e6c8f5a3e60922ca57188de264e/github/actions_workflows.go#L170-L179

SamMorrowDrums: I'd be tempted to flip this around too, and call it `workflow_file` or something, as it has file access already it can probably work out what workflows it can provide. The ID is only useful where it has received it from a list, and the list endpoint can probably return the file location as part of the list, which is also more useful than the ID, in terms of editing the file.

Does that make sense? The goal here is to make an API that LLMs comprehend and use natively, so definitely have a think on this.

pkg/github/actions.go
SamMorrowDrums: Appreciate you mentioning though, it's same reason I didn't consider new packages for tools yet.

Pretty sure the helpers and server can be extracted to their own packages to avoid this.

CC @williammartin @juruen

juruen: > There are a _lot_ of Actions APIs, I suspect the single-file per API area will break down at some point.
> 
> I tried defining a separate `actions` package, but it created some circular dependencies with this `github` package.

This might open an interesting debate that really resonates with me!

When I started writing Go, I came from other languages where quickly deciding to use packages was the norm. So I followed suit, and my initial reaction was to do the same. However, over time, I've come to terms with the fact that this might not be entirely idiomatic. 

Take `go-github`, for example — they implement almost the entire GitHub API within [a single package](https://github.com/google/go-github/tree/master/github).

So my point here is that maybe we don't need to introduce more packages right away. We might consider another alternative, such as splitting the functionality into files like `actions_foo.go`, `actions_bar.go`, etc., and that might be more idiomatic.

Having said that,  I don't have a strong opinion on it, but I'd like us to consider the above before we make the decision.

Thanks! ❤️

pkg/github/actions.go
joshmgross: Renamed to `workflow_file` - https://github.com/github/github-mcp-server/pull/269/commits/034621bddc7051538015b052113a38e183bbeb58

pkg/github/tools.go
```diff
@@ -82,6 +86,7 @@ func InitToolsets(passedToolsets []string, readOnly bool, getClient GetClientFn,
 	tsg.AddToolset(users)
 	tsg.AddToolset(pullRequests)
 	tsg.AddToolset(codeSecurity)
+	tsg.AddToolset(actions)
```
joshmgross, line 95: @SamMorrowDrums let me know if I added this correctly

Looking through #188, it seems like we don't want some toolsets enabled by default but it's not clear to me what enables that.

> Default Enabled Features
> By default, the following features are enabled:
>...

As far as I can tell, the default is "all"

https://github.com/github/github-mcp-server/blob/614f2267249ee9e4aeb8696d29caa93f3752196a/pkg/github/tools.go#L14

tonytrg:
Thanks for the contrib @joshmgross ❤️ 

But it seems https://github.com/github/github-mcp-server/pull/491 covers this and some more.

PR #270, opened 2025-04-14
Add "Mark notification as done" functionality

rfearing:
### Add `MarkNotificationAsDone` to notifications tooling

**Note:** Target branch is `notifications-tooling` https://github.com/github/github-mcp-server/pull/225

#### Feature added

- `mark_notification_done: Mark a specific notification thread as done

#### Technical implementation

- I largely copied `mark_notification_read`
- For the param `threadId`, `WithNumber` ran into some issues, so I kept the param as a string. I'm open to suggestions.

### DIFF
--- a/pkg/github/server.go
+++ b/pkg/github/server.go
@@ -97,6 +97,7 @@ func NewServer(getClient GetClientFn, version string, readOnly bool, t translati
 	if !readOnly {
 		s.AddTool(MarkNotificationRead(getClient, t))
 		s.AddTool(MarkAllNotificationsRead(getClient, t))
+		s.AddTool(MarkNotificationDone(getClient, t))
 	}
 
 	return s
--- a/pkg/github/notifications.go
+++ b/pkg/github/notifications.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"strconv"
 	"time"
 
 	"github.com/github/github-mcp-server/pkg/translations"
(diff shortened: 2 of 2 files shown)


juruen:
@rfearing as having too many tools might become an issue, maybe it'd be worth exploring the possibility of  merging some of these tools into one. 

It seems that marking notifications might be a good candidate to having just one and using the parameters to decide what to do exactly.

And the same goes for the getters.

In general, we should probably explore providing higher level APIs instead of having a 1:1 mapping with the underlying API.

We are already doing that in some tools with a positive result.

In any case, these are just my two cents.

/cc @toby @SamMorrowDrums @williammartin

sridharavinash (approved):
Nice work! 💖

SamMorrowDrums:
Surprised this was merged without addressing @juruen 's comment.

Probably it makes more sense for the two mark notification tools to be one with an enum for what state to change it to.

Can consider as a follow-up though.

pkg/github/server.go
```diff
@@ -97,6 +97,7 @@ func NewServer(getClient GetClientFn, version string, readOnly bool, t translati
 	if !readOnly {
 		s.AddTool(MarkNotificationRead(getClient, t))
 		s.AddTool(MarkAllNotificationsRead(getClient, t))
```
SamMorrowDrums, line 99: Oh no it looks like the init code has been broken too. This is old.

SamMorrowDrums:
Oh, sorry I didn't realise this was merged into a branch!

rfearing:
@SamMorrowDrums, @juruen I targeted @sridharavinash's branch and that PR has not yet been merged

@sridharavinash I'm happy to invest some time into looking into @juruen's comment but in your branch.

SamMorrowDrums:
@rfearing thank you, really appreciate this too - I was just worried it landed into main without a final pass. There's a bit of a rebase involved to get it into main too :-D

Bu also happy to take a look at the next branch, I just missed it wasn't a merge to main!

Issue #192, opened 2025-04-09
Unable to download the docker image


akordowski:
I setup the MCP server as described in the `README`. But starting the server I get following errors:

```bash
[info] Starting server github
[info] Connection state: Starting
[info] Starting server from LocalProcess extension host
[info] Connection state: Starting
[info] Connection state: Running
[warning] [server stderr] Unable to find image 'ghcr.io/github/github-mcp-server:latest' locally
[warning] [server stderr] docker: Error response from daemon: Head "https://ghcr.io/v2/github/github-mcp-server/manifests/latest": denied: denied
[warning] [server stderr] 
[warning] [server stderr] Run 'docker run --help' for more information
[info] Connection state: Error Process exited with code 125
```

Any idea what I am doing wrong?

edgeq:
➕ - also unable to pull the docker image. Even when pulling manually: 

```
docker pull ghcr.io/github/github-mcp-server
Using default tag: latest
Error response from daemon: Head "https://ghcr.io/v2/github/github-mcp-server/manifests/latest": denied: denied
```

edgeq:
My current workaround is to use the official [anthropic mcp server](https://github.com/modelcontextprotocol/servers/tree/main/src/github). It's in deprecation, but it's a trusted source and still works.

juruen:
@akordowski @edgeq is there any chance that you were logged into the gh docker registry with an expired token?

I'm unable to reproduce this as I can successfully pull the image anonymously:

```
juruen@moonbase:~$ docker logout ghcr.io  # logout from gh docker registry
Removing login credentials for ghcr.io
juruen@moonbase:~$ docker pull ghcr.io/github/github-mcp-server
Using default tag: latest
latest: Pulling from github/github-mcp-server
3d78e577de35: Pull complete
bfb59b82a9b6: Pull complete
4eff9a62d888: Pull complete
a62778643d56: Pull complete
7c12895b777b: Pull complete
3214acf345c0: Pull complete
5664b15f108b: Pull complete
0bab15eea81d: Pull complete
4aa0ea1413d3: Pull complete
da7816fa955e: Pull complete
9aee425378d2: Pull complete
d00c3209d929: Pull complete
221438ca359c: Pull complete
e76e88e5b481: Pull complete
4d3ee8eeb83f: Pull complete
Digest: sha256:faa1dfed068fb1d9c866a5ce4909947a5f2a5a72130815fba56c32bb441ca83f
[...]

PR #282, opened 2025-04-15
Pre-reqs for docker use include auth to ghcr.io

ismith:
To pull the image, you need docker to be logged in to ghcr.io; this PR links in the README to instructions for doing that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ automation and interaction capabilities for developers and tools.
 ## Prerequisites
 
 1. To run the server in a container, you will need to have [Docker](https://www.docker.com/) installed.
-2. Once Docker is installed, you will also need to ensure Docker is running.
+2. Once Docker is installed, you will also need to ensure Docker is running, and that you are [logged in to the GitHub Container Registry (ghcr.io)](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry#authenticating-with-a-personal-access-token-classic).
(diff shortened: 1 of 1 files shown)


SamMorrowDrums:
Huh, is this correct? @juruen you validated that anonymous docker pull should just work right? 

I've also added this to other machines and scripts and had it pull without authentication. I'm wondering what issue you are experiencing here.

juruen:
> Huh, is this correct? @juruen you validated that anonymous docker pull should just work right?

@SamMorrowDrums  Yup. I validated it in two ways:

- brand new vm that never did a  `docker login` to any registries
- `docker logout ghcr.io` and `docker pull` on another vm

See  https://github.com/github/github-mcp-server/issues/192#issuecomment-2798798987

A `docker logout ghcr.io` helped that user. So maybe users who are logged into `ghcr.io` with an expired token are experiencing these issues.

@ismith could you try this:

```
docker logout ghcr.io
docker pull ghcr.io/github/github-mcp-server
```

SamMorrowDrums:
> users who are logged into ghcr.io with a expired token are experiencing these issues.

_Your docker registry was so preoccupied with whether or not they could validate your token, they didn't stop to think if they should._

Your scientists were so preoccupied with whether or not they could, they didn't stop to think if they should scene from Jurassic Park

williammartin:
@ismith if https://github.com/github/github-mcp-server/pull/282#issuecomment-2807098717 is correct, could you update this PR to the effect of "the image is public, so if you're getting errors, it's probably because your token is expired"

williammartin:
Gentle nudge @ismith.

PR #298, opened 2025-04-16
Notice for Docker engine

koke1997:
Docker engine needs to run in background before Claude Desktop App. Even if you run Claude Desktop as adminstrator , this will still not run docker engine .

Closes:

I think adding this into readme would reduce the number of incoming issues as there a Windows users which think Docker needs to be run inside WSL

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,8 @@ More about using MCP server tools in VS Code's [agent mode documentation](https:
 
 ### Usage with Claude Desktop
 
+Keep in mind Docker engine has to run in background before running Claude Desktop App if you are using Windows.
+
 ```json
 {
   "mcpServers": {


README.md
```diff
@@ -68,6 +68,8 @@ More about using MCP server tools in VS Code's [agent mode documentation](https:
 
 ### Usage with Claude Desktop
 
+Keep in mind that the Docker engine must be running in the background before starting the Claude Desktop App on Windows.
```
```suggestion
Keep in mind that the [Docker Engine](https://docs.docker.com/engine/)/[Docker Desktop](https://www.docker.com/products/docker-desktop/) must be running in the background before starting the Claude Desktop App on Windows.
```

I'd also add that this isn't specific to Claude Desktop, but just generally about running with Docker so perhaps it should be moved up to a more general place, or troubleshooting section.

koke1997 (commented):
I think this ok because Claude Desktop could be running in tray already even before running the docker image. So in this case for Claude it is important to run docker before Claude Desktop App, otherwise running it later won't make the Desktop App attach to process.

williammartin:
Do we know it's only relevant to Windows, or also Mac?

ghost:
@williammartin I am not sure, but i think after installing for the first time Docker on Mac this will not launch on startup with macOS. This option has to be ticked.

From my Claude Desktop app : 

image

ghost:
@williammartin @SamMorrowDrums  Any feedback? Can we merge this?

ghost:
@tonytrg Excuse me, but can i please get some feedback on this?

README.md
```diff
@@ -68,6 +68,8 @@ More about using MCP server tools in VS Code's [agent mode documentation](https:
 
 ### Usage with Claude Desktop
 
+Keep in mind Docker engine has to run in background before running Claude Desktop App if you are using Windows.
```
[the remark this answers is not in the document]
killers10: O

README.md
```diff
@@ -68,6 +68,8 @@ More about using MCP server tools in VS Code's [agent mode documentation](https:
 
 ### Usage with Claude Desktop
 
+Keep in mind that the [Docker Engine](https://docs.docker.com/engine/)/[Docker Desktop](https://www.docker.com/products/docker-desktop/) must be running in the background before starting the Claude Desktop App on Windows.
+
 ```json
 {
   "mcpServers": {
```
killers10, line 75: Evildead

README.md
```diff
@@ -68,6 +68,8 @@ More about using MCP server tools in VS Code's [agent mode documentation](https:
 
 ### Usage with Claude Desktop
```
killers10, line 69: Evildead

killers10: E

Issue #191, opened 2025-04-09
MCP Error -32603: 422 Invalid request - 'sha' wasn't supplied when updating file


evanmcgillivray:
## Description
When trying to update a file using the create_or_update_file tool, I'm encountering an error where the SHA parameter isn't being supplied, resulting in a 422 Invalid request from the GitHub API.

## Error Message
```
Failed to call tool create_or_update_file: Error: MCP error -32603: failed to https://api.github.com/repos/evanmcgillivray/Budget/contents/src/components
422 Invalid request. "sha" wasn't supplied. []
```

## Steps to Reproduce
1. Using the MCP server with an AI assistant
2. Attempting to update a file in a repository
3. The MCP server fails to include the required SHA parameter in the request

## Expected Behavior
The MCP server should properly include the SHA parameter when updating files, as required by the GitHub API.

## Additional Information
- GitHub requires the SHA parameter when updating a file to ensure you're updating the correct version
- This appears to be happening specifically when trying to update a file via the create_or_update_file function

PR #320, opened 2025-04-21
Fix: Specify sha is required for file update

crondinini:
Closes: #191

## The problem

When the LLM sends an `update_file` request, it doesn't include the `sha` of the file being replaced, leading to the error:
```
Failed to call tool create_or_update_file: Error: MCP error -32603: failed to https://api.github.com/repos/evanmcgillivray/Budget/contents/src/components
422 Invalid request. "sha" wasn't supplied. []
```

This happens because `sha` is an optional field since the same tool is used for both `create` and `update` file operations.

## The solution

I did some local testing and the issue seems to be fixed if we specify in the tool description that `sha` is required when updating a file.

However, a more robust solution might be to split the tool into 2 separate tools and and make a new `update_file` tool that has a required `sha` property. 

I wasn't sure which approach would be preferred, so I've decided to put forward the simplest solution that works to hear your feedback.


juruen:
> However, a more robust solution might be to split the tool into 2 separate tools and and make a new `update_file` tool that has a required `sha` property.
> 
> I wasn't sure which approach would be preferred, so I've decided to put forward the simplest solution that works to hear your feedback.

Thank you so much for the PR.

@crondinini Although having separate tools allows us to specify the required field directly in the schema, there have been some concerns about adding too many tools. That’s why we sometimes prefer to merge several tools into one. As far as I know, this isn’t set in stone yet and could definitely change.

Having said that, as this is a minor change to the description that might help to make sure that the update operation is called correctly, I think we should merge it. Please, feel free to correct my assumption!

I've successfully tested it:


Screenshot 2025-04-28 at 11 16 02

Screenshot 2025-04-28 at 11 16 13

crondinini:
@juruen Thank you for the review and for the explanation. That makes sense. Considering this is such a small change that fixes a bug, I agree it makes sense to merge. 

I had to solve a conflict so the review was automatically dismissed.

If you could give your approval again, that would be great! Thank you :)

SamMorrowDrums (approved):
Thank you for this 🙏

PR #330, opened 2025-04-22
Add simple e2e test

williammartin:
## Description

Whilst reviewing dependabot updates, I really wanted a fast way to build a Docker image, start the server, call a tool and validate the result, as opposed to [doing it manually](https://github.com/github/github-mcp-server/pull/323#pullrequestreview-2784138391).

The primary audience for the current e2e test is myself, and any other maintainer, and in future possibly other contributors. See the readme for some limitations on the approach, which is why I don't necessarily want to make all contributors first class audience, though there's no reason that it couldn't be used.

The implementation is pretty MVP, and that's also intentional, until we see if and where it provides value.

### DIFF
--- a/e2e/README.md
+++ b/e2e/README.md
@@ -1,0 +1,84 @@
+# End To End (e2e) Tests
+
+The purpose of the E2E tests is to have a simple (currently) test that gives maintainers some confidence in the black box behavior of our artifacts. It does this by:
+ * Building the `github-mcp-server` docker image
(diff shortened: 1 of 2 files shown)


toby (dismissed):

williammartin:
Bypassing codeowner review rule because the only change I made after review from @toby was to expand e2e to End to End in the README.

Samido254:
Impressed 😁

PR #335, opened 2025-04-23
Improve README by including config example for `gh-host` flag

rajbos:
I could figure out how to configure the MCP server against GitHub Enterprise Server, but it took a couple of tries and debugging as I forgot to include the http schema in the hostname. Adding an example makes it a bit more clear in my opinion, improving the user experience of the server 😄.


williammartin:
Hey @rajbos thanks for opening this. Instead of doing this (or in addition), do you think we should adjust the server to append `https://` if the scheme is unspecified?

rajbos:
Adding it automatically also works. For me it helped to be able to see the example of the extra container arguments, helpful if you do not run them all the time. So doing both would be my preference then 😁.

williammartin:
Yeh I just have a minor concern because the JSON is still bifurcated between VSCode and Insiders on whether the `input:` prompt is supported. Maybe we could adjust this to just demonstrate an amended `env` block?

rajbos:
> Yeh I just have a minor concern because the JSON is still bifurcated between VSCode and Insiders on whether the `input:` prompt is supported. Maybe we could adjust this to just demonstrate an amended `env` block?

I am using the input block in normal VS Code v1.99.3, so it already is available then?

SamMorrowDrums:
@rajbos the issue @williammartin was referring to is that the one-click install button doesn't support setting inputs automatically for VScode (it does for insiders) so for now people are getting a configuration that has the PAT in the env block, and not as an input, but that will be fixed after the next VScode release.

chidaozhi:
Could you please let me know how to config "-gh-host"?
Is this right?(use in cursor)
`"args": [
        "run",
        "-i",
        "--rm",
        "-e",
        "--gh-host=https://xxx.xxx.com/xxx",
        "GITHUB_PERSONAL_ACCESS_TOKEN",
        "ghcr.io/github/github-mcp-server"
      ],`

rajbos:
Check the changes files in the PR for the example. You now have it as 'name=value', which is not correct.

README.md
```diff
@@ -193,7 +193,27 @@ docker run -i --rm \
 ## GitHub Enterprise Server
 
 The flag `--gh-host` and the environment variable `GITHUB_HOST` can be used to set
-the GitHub Enterprise Server hostname.
+the GitHub Enterprise Server hostname inculding the `https` connection schema:
```
```suggestion
the GitHub Enterprise Server hostname.
Prefix the hostname with the `https://` URI scheme, as it otherwise defaults to `http://` which GitHub Enterprise Server does not support.
```

README.md
```diff
@@ -193,7 +193,27 @@ docker run -i --rm \
...
+    "--rm",
+    "-e",
+    "GITHUB_PERSONAL_ACCESS_TOKEN",
+    "-e",
+    "GITHUB_HOST",
+    "ghcr.io/github/github-mcp-server"
+    ],
+    "env": {
+        "GITHUB_PERSONAL_ACCESS_TOKEN": "${input:github_token}",
+        "GITHUB_HOST": "https://your_full_domain_name_dot_com"
```
```suggestion
        "GITHUB_HOST": "https://<your GHES domain name>"
```

SamMorrowDrums (commented):
A couple of notes, but I agree it's nice to expand the example.

rajbos:
LGTM!

SamMorrowDrums (approved):
🚀 thanks Rob!

Issue #350, opened 2025-04-25
get_file_contents: Add ability to select line(s) ranges


kaovilai:
### Describe the feature or problem you’d like to solve
get_file_contents - Get contents of a file or directory

owner: Repository owner (string, required)
repo: Repository name (string, required)
path: File path (string, required)
ref: Git reference (string, optional)
**begin: begin line# (number, optional)
end: end line# (number, optional)**

requested addition in bold
### Proposed solution
See above

### Additional context

This may help reduce model context consumed by parsing large files if we only want to request a subset..

aryasoni98:
@kaovilai I came across this issue and would love to contribute could you please assign it to me?

kaovilai:
I'm not a maintainer here so I cannot assign anyone.

Akshata990418:
I have integrated Github MCP to my application chainlit+langchain i m also not able to fetch the file content using get_file_contents tool. could anyone help with this @aryasoni98 @kaovilai 

Im facing this error while trying to use get_file_contents tool

**TypeError: Type is not msgpack serializable: ToolMessage**

PR #235, opened 2025-04-12
Document add_sub_issue tool functionality

aryasoni98:
Closes: #196

PR #351, opened 2025-04-26
feat: add line range support to get_file_contents tool

aryasoni98:
Closes: #350


README.md
```diff
@@ -15,6 +15,66 @@ automation and interaction capabilities for developers and tools.
 ## Prerequisites
 
 1. To run the server in a container, you will need to have [Docker](https://www.docker.com/) installed.
+
+## Required Token Permissions
```
williammartin, line 19: Please remove this section. I think it's interesting to have, but should have its own issue for discussion.

README.md
```diff
@@ -491,14 +551,16 @@ export GITHUB_MCP_TOOL_ADD_ISSUE_COMMENT_DESCRIPTION="an alternative description
...
-- **list_commits** - Get a list of commits of a branch in a repository
+- **list_commits** - Gets commits of a branch in a repository
   - `owner`: Repository owner (string, required)
   - `repo`: Repository name (string, required)
   - `sha`: Branch name, tag, or commit SHA (string, optional)
   - `path`: Only commits containing this file path (string, optional)
   - `page`: Page number (number, optional)
   - `perPage`: Results per page (number, optional)
 
+### Search
```
williammartin, line 590: Not sure why this is moved.

pkg/github/repositories.go
```diff
@@ -471,7 +488,39 @@ func GetFileContents(getClient GetClientFn, t translations.TranslationHelperFunc
 
 			var result interface{}
 			if fileContent != nil {
-				result = fileContent
+				if fileContent.Content != nil && (begin > 0 || end > 0) {
```
williammartin, line 491: I'm not sure about this implementation. Should it be possible for the LLM to only provide the beginning or the end? Whatever we choose should be documented so the LLM understands what is acceptable.

pkg/github/repositories.go
```diff
@@ -471,7 +488,39 @@ func GetFileContents(getClient GetClientFn, t translations.TranslationHelperFunc
...
+					ranged := lines[startIdx:endIdx]
+					joined := strings.Join(ranged, "\n")
+					result = &github.RepositoryContent{
+						Content: github.Ptr(base64.StdEncoding.EncodeToString([]byte(joined))),
+					}
+				} else {
+					result = fileContent
+				}
 			} else {
 				result = dirContent
```
williammartin, line 559: What should happen if the LLM provides begin and end but the contents returned are `dirContent`? How do we communicate what is acceptable to the LLM in the description?

pkg/github/repositories.go
williammartin, line 491: All of this needs to be unit tested.

pkg/github/repositories.go
```diff
@@ -471,7 +488,39 @@ func GetFileContents(getClient GetClientFn, t translations.TranslationHelperFunc
...
+					}
+					if startIdx < 0 {
+						startIdx = 0
+					}
+					if endIdx < startIdx {
+						endIdx = startIdx
+					}
+					ranged := lines[startIdx:endIdx]
+					joined := strings.Join(ranged, "\n")
+					result = &github.RepositoryContent{
```
williammartin: It seems like we should check the `.Encoding` of the file content before deciding that we need to base64 encode it.

williammartin (changes requested):
Thanks for this, added some notes. I haven't reviewed the line implementation because I'd rather see tests that pass first.

README.md
aryasoni98: Fix

README.md
aryasoni98: Fix

pkg/github/repositories.go
aryasoni98: Fix

pkg/github/repositories.go
aryasoni98: Fix

pkg/github/repositories.go
aryasoni98: Fix

pkg/github/repositories.go
aryasoni98: Fix

williammartin:
See https://github.com/github/github-mcp-server/pull/235#issuecomment-2892909687 for why I am closing this.

Issue #343, opened 2025-04-24
Remove comments from `create_pull_request_review`


toby:
Due to OpenAI, Windsurf and Gemini all handling JSON schemas in a conflicting manner, we need to remove commenting from `create_pull_request_review`. The LLM should still be able to leave comments on PRs using `add_pull_request_review_comment`. More info on [modelcontextprotocol/modelcontextprotocol/issues/410](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/410).

williammartin:
To do this we're going to need to adjust `add_pull_request_review_comment` to use the GraphQL API, because the REST API doesn't allow you to add new comments to an existing review: https://github.com/microsoft/vscode-pull-request-github/issues/908#issuecomment-541791329

williammartin:
I have a branch now that implements most of this in an MVP manner: https://github.com/github/github-mcp-server/commit/1e9fbbcf9731a24974542488d399bd0c904b10e7

Here's the results of the e2e test running:

```
➜  git:(343-remove-comments-from-create_pull_request_review) ✗ GITHUB_MCP_SERVER_E2E_TOKEN=$(gh auth token) go test -v -count=1 -run  "TestPullRequestReview" --tags e2e ./e2e

=== RUN   TestPullRequestReview
=== PAUSE TestPullRequestReview
=== CONT  TestPullRequestReview
    e2e_test.go:45: Building Docker image for e2e tests...
    e2e_test.go:118: Starting Stdio MCP client...
    e2e_test.go:221: Getting current user...
    e2e_test.go:250: Creating repository williammartin/github-mcp-server-e2e-TestPullRequestReview-1745611991474...
    e2e_test.go:274: Creating branch in williammartin/github-mcp-server-e2e-TestPullRequestReview-1745611991474...
    e2e_test.go:291: Creating commit with new file in williammartin/github-mcp-server-e2e-TestPullRequestReview-1745611991474...
    e2e_test.go:319: Creating pull request in williammartin/github-mcp-server-e2e-TestPullRequestReview-1745611991474...
    e2e_test.go:334: Creating pending review for pull request in williammartin/github-mcp-server-e2e-TestPullRequestReview-1745611991474...
    e2e_test.go:359: Adding review comment to pull request in williammartin/github-mcp-server-e2e-TestPullRequestReview-1745611991474...
[...]

PR #381, opened 2025-05-07
Split PR review creation, commenting and submission, and deletion

williammartin:
## Description

Relates to: https://github.com/github/github-mcp-server/issues/343

@toby @SamMorrowDrums before I go any further cleaning up the new and old tools, handling edge cases and adding unit, I want to get your feedback on this approach, since there have been quite a few pain points along the way.

The original issue suggested removing `comments` from  `create_pull_request_review` and depending on `add_pull_request_review_comment`, however, `add_pull_request_review_comment` uses the REST API which is only able to create a single comment and only if a review is not already in progress.

Thus, we need to lean on GraphQL here to:
 * Create a pending review (`mvp_create_pending_pull_request_review`)
 * Add a comment to that pending review (`mvp_add_pull_request_review_comment_to_pending_review`)
 * Submit the pending review (`mvp_submit_pull_request_review`)

The implicit requirement here is that a user may only have one pending review at a time, so if we get their most recent review on a PR, and it is pending, we can add a comment to that. This does not allow commenting on other reviews and it does not allow commenting on already submitted reviews, which could be further enhancements in the same or in different tools.

There is an additional challenge since we are now using GraphQL which is that for the most part, queries and mutations operate on GraphQL IDs (`node_id` in the REST API). So for example, when we create a pending review, there is a `reviewId` that is required for commenting. Within a single session, it would be easy to return the `id` in the `textContent` but this:
 * Leaks the API implementation (maybe not a big deal?)
 * Would require new sessions to request the Id in some way in order to provide it the next time

For now, I've decided with these very explicit tools (e.g. `mvp_add_pull_request_review_comment_to_pending_review`) to allow the model to be API implementation agnostic, and that the tool will internally do the lookup. For example, since we know there can only be one review for the user, knowing the `owner`, `repo` and `pr` number seems to be enough to look up the most recent review. The downside is that it's a bit more restrictive and requires more internal API lookups than an alternative which might be to expose a `get_pending_review` and allow the model to figure it out. If you asked me I would ship this and wait for feedback on when it sucks or should be enhanced later but I don't have much confidence in my intuition here.

Cheers.

---

Currently passes e2e tests:

```
➜  github-mcp-server git:(0ca07aa) ✗ GOMAXPROCS=1 GITHUB_MCP_SERVER_E2E_TOKEN=$(gh auth token) go test -v -count=1 --tags e2e ./e2e

=== RUN   TestGetMe
=== PAUSE TestGetMe
=== RUN   TestToolsets
=== PAUSE TestToolsets
=== RUN   TestTags
=== PAUSE TestTags
=== RUN   TestPullRequestReview
=== PAUSE TestPullRequestReview
=== CONT  TestGetMe
    e2e_test.go:49: Building Docker image for e2e tests...
    e2e_test.go:123: Starting Stdio MCP client...
--- PASS: TestGetMe (3.33s)
=== CONT  TestPullRequestReview
    e2e_test.go:123: Starting Stdio MCP client...
    e2e_test.go:384: Getting current user...
    e2e_test.go:413: Creating repository williammartin/github-mcp-server-e2e-TestPullRequestReview-1746631692347...
    e2e_test.go:437: Creating branch in williammartin/github-mcp-server-e2e-TestPullRequestReview-1746631692347...
    e2e_test.go:454: Creating commit with new file in williammartin/github-mcp-server-e2e-TestPullRequestReview-1746631692347...
    e2e_test.go:482: Creating pull request in williammartin/github-mcp-server-e2e-TestPullRequestReview-1746631692347...
    e2e_test.go:497: Creating pending review for pull request in williammartin/github-mcp-server-e2e-TestPullRequestReview-1746631692347...
    e2e_test.go:518: Adding review comment to pull request in williammartin/github-mcp-server-e2e-TestPullRequestReview-1746631692347...
    e2e_test.go:534: Submitting review for pull request in williammartin/github-mcp-server-e2e-TestPullRequestReview-1746631692347...
    e2e_test.go:548: Getting reviews for pull request in williammartin/github-mcp-server-e2e-TestPullRequestReview-1746631692347...
    e2e_test.go:422: Deleting repository williammartin/github-mcp-server-e2e-TestPullRequestReview-1746631692347...
--- PASS: TestPullRequestReview (9.26s)
=== CONT  TestTags
    e2e_test.go:123: Starting Stdio MCP client...
    e2e_test.go:246: Getting current user...
    e2e_test.go:275: Creating repository williammartin/github-mcp-server-e2e-TestTags-1746631701637...
    e2e_test.go:292: Creating tag williammartin/github-mcp-server-e2e-TestTags-1746631701637:v0.0.1...
    e2e_test.go:322: Listing tags for williammartin/github-mcp-server-e2e-TestTags-1746631701637...
    e2e_test.go:355: Getting tag williammartin/github-mcp-server-e2e-TestTags-1746631701637:v0.0.1...
    e2e_test.go:284: Deleting repository williammartin/github-mcp-server-e2e-TestTags-1746631701637...
--- PASS: TestTags (3.75s)
=== CONT  TestToolsets
    e2e_test.go:123: Starting Stdio MCP client...
--- PASS: TestToolsets (0.17s)
PASS
ok      github.com/github/github-mcp-server/e2e 16.767s
```


toby:
Thank you for this @williammartin! I would say that yes, this is worth shipping and getting feedback from. It's better to be a bit more limited (and have a more complex backend api dance) than to break on launch because of JSON schema conflicts imho.

williammartin:
Ok thanks, I'll move it forward to a place where I think it's acceptable and then ask for any final comments.

SamMorrowDrums:
I think the direction sounds great. I'd love to know how well models navigate it.

We could have simple `approve_pr`, `reject_pr` tools that only do top level comment, when a full review is not required but I expect the would cause more confusion not less.

I can't wait to try this out.

SamMorrowDrums:
@williammartin I think we need a cancel draft pull request review too, so there's a way to get past the fact you can only have one at a time issue.

williammartin (commented):
This is a test review of a tool in this PR itself.

williammartin:
@toby @SamMorrowDrums please play around with the changes in this PR. The delete review tool has been added. I haven't written unit tests yet, but you should look at the e2e tests to understand the flow.

internal/ghmcp/server.go
```diff
@@ -87,11 +89,21 @@ func NewMCPServer(cfg MCPServerConfig) (*server.MCPServer, error) {
 		return ghClient, nil // closing over client
 	}
 
+	getGQLClient := func(_ context.Context) (*githubv4.Client, error) {
```
williammartin, line 91: TODO: Make enterprise ready.

pkg/github/server.go
```diff
@@ -29,6 +30,27 @@ func NewServer(version string, opts ...server.ServerOption) *server.MCPServer {
 	return s
 }
 
+type constrainableValue interface {
+	Constrain(any) error
+}
+
+type constrainableInt32 int32
+
+func (ci *constrainableInt32) Constrain(param any) error {
```
williammartin: Ok, I'm not sure if this is way overengineered or not, but it facilitates constraining param values: https://github.com/github/github-mcp-server/pull/381/files#diff-c2e2a7250ef597d08ee5d429159bc89c4b862a7ed6cbcf30b8231a3ec599d190R1081

pkg/github/pullrequests.go
```diff
@@ -75,8 +76,123 @@ func GetPullRequest(getClient GetClientFn, t translations.TranslationHelperFunc)
 		}
 }
 
+// CreatePullRequest creates a tool to create a new pull request.
+func CreatePullRequest(getClient GetClientFn, t translations.TranslationHelperFunc) (mcp.Tool, server.ToolHandlerFunc) {
```
williammartin, line 83: This is just moved from below to a more sensible location in the file, putting the review tools together in one place.

pkg/github/tools.go
```diff
@@ -65,10 +67,16 @@ func InitToolsets(passedToolsets []string, readOnly bool, getClient GetClientFn,
 		AddWriteTools(
 			toolsets.NewServerTool(MergePullRequest(getClient, t)),
 			toolsets.NewServerTool(UpdatePullRequestBranch(getClient, t)),
-			toolsets.NewServerTool(CreatePullRequestReview(getClient, t)),
 			toolsets.NewServerTool(CreatePullRequest(getClient, t)),
 			toolsets.NewServerTool(UpdatePullRequest(getClient, t)),
 			toolsets.NewServerTool(AddPullRequestReviewComment(getClient, t)),
```
williammartin: TODO: decide what to do with this standalone tool. I think it can just be removed in favour of create pending + add comment.

pkg/github/pullrequests.go
```diff
@@ -1148,101 +1273,459 @@ func CreatePullRequest(getClient GetClientFn, t translations.TranslationHelperFu
...
+			mcp.WithString("side",
+				mcp.Description("The side of the diff to comment on"),
+				mcp.Enum("LEFT", "RIGHT"),
+			),
+			mcp.WithNumber("startLine",
+				mcp.Description("For multi-line comments, the first line of the range that the comment applies to"),
+			),
+			mcp.WithString("startSide",
+				mcp.Description("For multi-line comments, the starting side of the diff that the comment applies to"),
+				mcp.Enum("LEFT", "RIGHT"),
```
williammartin: It's not really clear to me whether this actually works with Gemini, but it matches the other existing Tool `AddPullRequestReviewComment` that doesn't seem to cause gemini to barf.

williammartin (commented):
this is a test from gemini model

williammartin (commented):
this is a test from gemini model

williammartin (commented):
this is a test from gemini model

williammartin:
Using gemini 2.5 Pro (Preview) in VS Code is basically unusable due to 500s (where the backend seems to be rate-limited). However, on the rare chance I got it to work I saw:

image

Comparatively, on `main` all I ever see is:

image

I tried this multiple times so my assumption is that it relates to the schema.

pkg/github/pullrequests.go
```diff
@@ -1148,101 +1273,459 @@ func CreatePullRequest(getClient GetClientFn, t translations.TranslationHelperFu
...
+			if err != nil {
+				return mcp.NewToolResultError(err.Error()), nil
 			}
 
-			newPR.Draft = github.Ptr(draft)
-			newPR.MaintainerCanModify = github.Ptr(maintainerCanModify)
+			client, err := getGQLClient(ctx)
+			if err != nil {
+				return nil, fmt.Errorf("failed to get GitHub GQL client: %w", err)
+			}
```
williammartin: TODO: A bit more e2e validation (and probably a manual check) of these working as expected

pkg/github/tools.go
```diff
@@ -65,10 +67,16 @@ func InitToolsets(passedToolsets []string, readOnly bool, getClient GetClientFn,
...
+
+			// Reviews
+			toolsets.NewServerTool(CreateAndSubmitPullRequestReview(getGQLClient, t)),
+			toolsets.NewServerTool(CreatePendingPullRequestReview(getGQLClient, t)),
+			toolsets.NewServerTool(AddPullRequestReviewCommentToPendingReview(getGQLClient, t)),
+			toolsets.NewServerTool(SubmitPendingPullRequestReview(getGQLClient, t)),
+			toolsets.NewServerTool(DeletePendingPullRequestReview(getGQLClient, t)),
```
williammartin, line 80: TODO: update README with new tools.

internal/ghmcp/server.go
SamMorrowDrums: And TODO add headers to these requests somehow too if possible, for same source identification purposes.

pkg/github/server.go
SamMorrowDrums: I will get an error calling this as a lib right? Because the type and interface are not exported so I cannot provide them.

I don't need to call it I suppose, just a thought that I would need to do that if I need to add similar functionality.

To your question, I think it makes sense - we want to avoid integer wrapping, and sanitising user input is generally a good thing. It is a little heavy handed perhaps, but I expect we will not touch it often. I've not had to touch any of these functions much, they largely just do what you'd expect, so I don't mind it - it's not a huge footprint.

SamMorrowDrums: I suppose one comment might be, why not allow the caller to pass the min and max, in case they wish to constrain a value futher? Then indeed you could lose the interface, and have a generic function with numeric argument with a min and max check based on user provided value?

🤔 hard to decide. I do get the desire for extending the type directly.

pkg/github/server.go
```diff
@@ -68,21 +90,32 @@ func requiredParam[T comparable](r mcp.CallToolRequest, p string) (T, error) {
...
+	if constrainableValue, ok := any(&parsableParamResult).(constrainableValue); ok {
+		if err := constrainableValue.Constrain(param); err != nil {
+			return zero, fmt.Errorf("failed to parse parameter %s: %w", p, err)
+		}
+		return parsableParamResult, nil
+	}
+
 	// Check if the parameter is of the expected type
-	if _, ok := r.Params.Arguments[p].(T); !ok {
+	typedParam, ok := r.Params.Arguments[p].(T)
```
SamMorrowDrums:
```suggestion
	typedParam, ok := param.(T)
```

Looks like you defined it and can do this here too.

pkg/github/server.go
```diff
@@ -106,16 +139,28 @@ func OptionalParam[T any](r mcp.CallToolRequest, p string) (T, error) {
...
+		if err := parseableParam.Constrain(param); err != nil {
+			return zero, fmt.Errorf("failed to parse parameter %s: %w", p, err)
+		}
+		return parsableParamResult, nil
+	}
+
 	// Check if the parameter is of the expected type
-	if _, ok := r.Params.Arguments[p].(T); !ok {
-		return zero, fmt.Errorf("parameter %s is not of type %T, is %T", p, zero, r.Params.Arguments[p])
+	typedParam, ok := param.(T)
```
SamMorrowDrums, line 137: Indeed you did do it here.

pkg/github/server_test.go
```diff
@@ -104,6 +107,70 @@ func Test_RequiredStringParam(t *testing.T) {
 	}
 }
 
+type testEnum string
+
+func (te *testEnum) Constrain(param any) error {
```
SamMorrowDrums: Ah, I see you used it a bunch.

pkg/github/tools.go
SamMorrowDrums: I think so, the one-shot version will get confusing otherwise.

pkg/github/pullrequests.go
```diff
@@ -947,199 +1066,205 @@ func CreatePullRequestReview(getClient GetClientFn, t translations.TranslationHe
...
-					if startLineFloat, ok := commentMap["start_line"].(float64); ok {
-						comment.StartLine = github.Ptr(int(startLineFloat))
-					}
-					if startSide, ok := commentMap["start_side"].(string); ok {
-						comment.StartSide = github.Ptr(startSide)
-					}
+// CreatePendingPullRequestReview creates a tool to create a pending review on a pull request.
+func CreatePendingPullRequestReview(getGQLClient GetGQLClientFn, t translations.TranslationHelperFunc) (mcp.Tool, server.ToolHandlerFunc) {
+	return mcp.NewTool("create_pending_pull_request_review",
+			mcp.WithDescription(t("TOOL_CREATE_PENDING_PULL_REQUEST_REVIEW_DESCRIPTION", "Create a pending review for a pull request.")),
```
SamMorrowDrums:
```suggestion
			mcp.WithDescription(t("TOOL_CREATE_PENDING_PULL_REQUEST_REVIEW_DESCRIPTION", "Create a pending review for a pull request. Call this first before attempting to add comments to a pending review, and ultimately submitting it. A pending pull request review means a pull request review, it is pending because you create it first and submit it later, and the PR author will not see it until it is submitted.")),
```

pkg/github/pullrequests.go
```diff
@@ -947,199 +1066,205 @@ func CreatePullRequestReview(getClient GetClientFn, t translations.TranslationHe
...
 }
 
-// CreatePullRequest creates a tool to create a new pull request.
-func CreatePullRequest(getClient GetClientFn, t translations.TranslationHelperFunc) (tool mcp.Tool, handler server.ToolHandlerFunc) {
-	return mcp.NewTool("create_pull_request",
-			mcp.WithDescription(t("TOOL_CREATE_PULL_REQUEST_DESCRIPTION", "Create a new pull request in a GitHub repository.")),
+// AddPullRequestReviewCommentToPendingReview creates a tool to add a comment to a pull request review.
+func AddPullRequestReviewCommentToPendingReview(getGQLClient GetGQLClientFn, t translations.TranslationHelperFunc) (mcp.Tool, server.ToolHandlerFunc) {
+	return mcp.NewTool("add_pull_request_review_comment_to_pending_review",
+			mcp.WithDescription(t("TOOL_ADD_PULL_REQUEST_REVIEW_COMMENT_TO_PENDING_REVIEW_DESCRIPTION", "Add a comment to the requester's latest pending pull request review.")),
```
SamMorrowDrums:
```suggestion
			mcp.WithDescription(t("TOOL_ADD_PULL_REQUEST_REVIEW_COMMENT_TO_PENDING_REVIEW_DESCRIPTION", "Add a comment to the requester's latest pending pull request review, a pending review needs to already exist to call this (check with the user if not sure).")),
```

pkg/github/pullrequests.go
```diff
@@ -1148,101 +1273,459 @@ func CreatePullRequest(getClient GetClientFn, t translations.TranslationHelperFu
...
+			// In future, we may want to return the review ID, but for the moment, we're not leaking
+			// API implementation details to the LLM.
+			return mcp.NewToolResultText(""), nil
+		}
+}
+
+// SubmitPendingPullRequestReview creates a tool to submit a pull request review.
+func SubmitPendingPullRequestReview(getGQLClient GetGQLClientFn, t translations.TranslationHelperFunc) (mcp.Tool, server.ToolHandlerFunc) {
+	return mcp.NewTool("submit_pending_pull_request_review",
+			mcp.WithDescription(t("TOOL_SUBMIT_PENDING_PULL_REQUEST_REVIEW_DESCRIPTION", "Submit the requester's latest pending pull request review.")),
```
SamMorrowDrums:
```suggestion
			mcp.WithDescription(t("TOOL_SUBMIT_PENDING_PULL_REQUEST_REVIEW_DESCRIPTION", "Submit the requester's latest pending pull request review, normally this is a final step after creating a pending review, adding comments first, unless you know that the user already did the first two steps, you should check before calling this.")),
```

pkg/github/pullrequests.go
```diff
@@ -1148,101 +1273,459 @@ func CreatePullRequest(getClient GetClientFn, t translations.TranslationHelperFu
...
 				return nil, fmt.Errorf("failed to marshal response: %w", err)
 			}
 
 			return mcp.NewToolResultText(string(r)), nil
 		}
 }
+
+func DeletePendingPullRequestReview(getGQLClient GetGQLClientFn, t translations.TranslationHelperFunc) (mcp.Tool, server.ToolHandlerFunc) {
+	return mcp.NewTool("delete_pending_pull_request_review",
+			mcp.WithDescription(t("TOOL_DELETE_PENDING_PULL_REQUEST_REVIEW_DESCRIPTION", "Delete the requester's latest pending pull request review.")),
```
SamMorrowDrums:
```suggestion
			mcp.WithDescription(t("TOOL_DELETE_PENDING_PULL_REQUEST_REVIEW_DESCRIPTION", "Delete the requester's latest pending pull request review. Use this after the user decides not to submit a pending review, if you don't know if they already created one then check first.")),
```

pkg/github/server.go
williammartin: I'm just going to remove it in favour of `ParseInt32`. It's simpler, and if we later discover we want this behaviour, we have evidence for it.

I have some ideas around parsing all args into structs that I think would be nice to explore first and are more general as a solution.

Thanks for the food for thought!

SamMorrowDrums (commented):
This is a test of the PR review feature. I've reviewed the mcp.WithDescription blocks in the pullrequests.go file and suggested improvements to make them more informative for LLMs by:

1. Adding more context about what the tools actually return
2. Using more descriptive language about the functionality
3. Including clear explanations of parameters and their effects
4. Highlighting relationships between related operations

These improvements will help LLMs better understand the purpose and capabilities of each GitHub tool when generating responses.

williammartin:
This has become pretty unwieldy as we were exploring the design space. I believe it is now ready to PR properly so I'm going to close this one in favour of a clean slate.

PR #389, opened 2025-05-09
Change code fence to neutral ```console

codingoutloud:
When ```bash is used for a code fence, bash keywords such as "for" and "in" get rendered with syntax highlighting. Switching to ```console avoids this distraction.

image

Closes:


williammartin (approved):
Thanks!

Issue #398, opened 2025-05-13
Add a tool to assign copilot to issues


williammartin:
### Describe the feature or problem you’d like to solve

With it becoming [possible to assign Copilot to issues](https://github.blog/news-insights/product-news/github-copilot-the-agent-awakens/), it would be nice to add a tool to allow this, in the same way copilot can be [requested for reviews](https://github.com/github/github-mcp-server/pull/387)

### What Prompt would I expect to use this tool

> Please assign copilot to <issue url>

aryasoni98:
@williammartin Would you like me to also help you draft a PR or suggest how to implement it if you're planning to code it out?

williammartin:
There's already a draft PR https://github.com/github/github-mcp-server/pull/399 thanks.

PR #399, opened 2025-05-13
Assign copilot to issues

williammartin:
## Description

Closes: #398

This PR drafts support for assigning copilot to issues as per project padawan. This is only available on github.com for users that are part of the preview.

image

image


internal/ghmcp/server.go
```diff
@@ -86,11 +88,21 @@ func NewMCPServer(cfg MCPServerConfig) (*server.MCPServer, error) {
 		return ghClient, nil // closing over client
 	}
 
+	getGQLClient := func(_ context.Context) (*githubv4.Client, error) {
```
williammartin, line 91: There will be a merge conflict here.

Issue #406, opened 2025-05-16
Add `golangci-lint-action` to the `lint` workflow


radar07:
### Describe the feature or problem you’d like to solve

Add [golangci-lint-action](https://github.com/golangci/golangci-lint-action) to the `lint` workflow

### Proposed solution

It reduces the complexity in configuring `golangci-lint` action in the GitHub workflow.

PR #407, opened 2025-05-16
feat: add `golangci-lint-action` to the `lint` workflow

radar07:
Closes: #406

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -17,34 +17,9 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v5
         with:
-          go-version-file: 'go.mod'
+          go-version-file: "go.mod"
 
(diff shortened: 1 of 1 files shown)


SamMorrowDrums (commented):
I love the simplicity just need to verify it  doesn't this miss a couple of the checks from the existing implementation before merging.

williammartin (changes requested):
@radar07 thanks for doing this but it loses the `go mod` checks. Please ensure this still does `go mod verify` and asserts nothing has changed after `go mod tidy`.

It's ok to lose `go fmt` because golangci-lint does include that check.

radar07:
@williammartin I missed this totally. I removed the `go fmt` check and added `golangci-lint` step after the **Run checks**.

SamMorrowDrums (approved):

PR #410, opened 2025-05-19
Split PR review creation, commenting, submission and deletion

williammartin:


internal/githubv4mock/githubv4mock.go
```diff
@@ -0,0 +1,235 @@
+// githubv4mock package provides a mock GraphQL server used for testing queries produced via
```
williammartin, line 1: This whole package should have its own tests but I want to get feedback on it before the extra effort.

pkg/github/pullrequests_test.go
```diff
@@ -1720,199 +1352,196 @@ func Test_CreatePullRequest(t *testing.T) {
...
-			name: "comment creation fails",
-			mockedClient: mock.NewMockedHTTPClient(
-				mock.WithRequestMatchHandler(
-					mock.PostReposPullsCommentsByOwnerByRepoByPullNumber,
-					http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
-						w.WriteHeader(http.StatusUnprocessableEntity)
-						w.Header().Set("Content-Type", "application/json")
-						_, _ = w.Write([]byte(`{"message": "Validation Failed"}`))
-					}),
+			name: "failure to get pull request",
```
williammartin, line 1434: I didn't do these GQL error cases in all the other tests intentionally yet, I just wanted to get a feel for whether it made sense. Every handler follows the same pattern on `if err := ... ; err != nil`

pkg/github/server_test.go
```diff
@@ -157,7 +164,7 @@ func Test_OptionalStringParam(t *testing.T) {
 	}
 }
 
-func Test_RequiredNumberParam(t *testing.T) {
+func Test_RequiredInt(t *testing.T) {
```
williammartin, line 167: These names just didn't match the actual functions.

pkg/github/tools.go
```diff
@@ -7,13 +7,15 @@ import (
...
+	"github.com/shurcooL/githubv4"
 )
 
 type GetClientFn func(context.Context) (*github.Client, error)
+type GetGQLClientFn func(context.Context) (*githubv4.Client, error)
 
 var DefaultTools = []string{"all"}
 
-func InitToolsets(passedToolsets []string, readOnly bool, getClient GetClientFn, t translations.TranslationHelperFunc) (*toolsets.ToolsetGroup, error) {
+func InitToolsets(passedToolsets []string, readOnly bool, getClient GetClientFn, getGQLClient GetGQLClientFn, t translations.TranslationHelperFunc) (*toolsets.ToolsetGroup, error) {
```
williammartin, line 18: This is going to be a breaking change for the remote server.

e2e/e2e_test.go
williammartin, line 1: I'm starting to feel the tension in the repetition in this file now, which is a good thing because it means we probably have sufficient examples to start teasing out shared abstractions, asserting the MCP result isn't an error and has text content etc etc.

Would prefer not to address in this PR since it's about behaviour rather than maintenance.

internal/ghmcp/server.go
```diff
@@ -44,25 +48,43 @@ type MCPServerConfig struct {
...
-		}
+	apiHost, err := parseAPIHost(cfg.Host)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse API host: %w", err)
 	}
 
+	// Construct our REST client
+	restClient := gogithub.NewClient(nil).WithAuthToken(cfg.Token)
+	restClient.UserAgent = fmt.Sprintf("github-mcp-server/%s", cfg.Version)
+	restClient.BaseURL = apiHost.baseRESTURL
```
williammartin, line 59: I chose to go away from `WithEnterpriseURLS` because there are hosts that are not enterprise but require different URLs e.g. review labs or development environments.

Sadly, `githubv4` doesn't offer quite the same options, so I commented below why that is. It also doesn't export the internal `graphql.Client` so we can't just set it.

internal/ghmcp/server.go
```diff
@@ -213,3 +240,141 @@ func RunStdioServer(cfg StdioServerConfig) error {
...
+	if s == "" {
+		return newDotcomHost()
+	}
+
+	url, err := url.Parse(s)
+	if err != nil {
+		return apiHost{}, fmt.Errorf("could not parse host as URL: %s", s)
+	}
+
+	if url.Scheme == "" {
```
williammartin: I'd be happy to loosen this later, but it is currently the expectation of the REST client.

pkg/github/helper_test.go
```diff
@@ -94,6 +94,12 @@ func mockResponse(t *testing.T, code int, body interface{}) http.HandlerFunc {
 	t.Helper()
 	return func(w http.ResponseWriter, _ *http.Request) {
 		w.WriteHeader(code)
+		s, ok := body.(string)
+		if ok {
+			_, _ = w.Write([]byte(s))
```
williammartin, line 101: Required for testing `get_pull_request_diff` which returns raw text, not JSON.

pkg/github/pullrequests.go
```diff
@@ -75,8 +77,123 @@ func GetPullRequest(getClient GetClientFn, t translations.TranslationHelperFunc)
 		}
 }
 
+// CreatePullRequest creates a tool to create a new pull request.
+func CreatePullRequest(getClient GetClientFn, t translations.TranslationHelperFunc) (mcp.Tool, server.ToolHandlerFunc) {
```
williammartin, line 81: This was just moved to a more natural place in the file.

pkg/github/pullrequests.go
```diff
@@ -936,210 +1076,274 @@ func CreatePullRequestReview(getClient GetClientFn, t translations.TranslationHe
...
+					PullRequest struct {
+						Reviews struct {
+							Nodes []struct {
+								ID    githubv4.ID
+								State githubv4.PullRequestReviewState
+								URL   githubv4.URI
+							}
+						} `graphql:"reviews(first: 1, author: $author)"`
+					} `graphql:"pullRequest(number: $prNum)"`
+				} `graphql:"repository(owner: $owner, name: $name)"`
```
williammartin, line 1151: Could we DRY this out?

williammartin: We could because a number of tools share it, but I'm keen to let it sit and soak in for a bit before choosing the wrong abstraction.

williammartin, line 1151: Could we DRY this out?

williammartin: Author note: this came from Gemini 2.5

williammartin: Author note: this came from GPT-4.1

internal/ghmcp/server.go
SamMorrowDrums: That's a shame! Might need to propose exporting it.

internal/githubv4mock/query.go
```diff
@@ -0,0 +1,157 @@
+// Ths contents of this file are taken from https://github.com/shurcooL/graphql/blob/ed46e5a4646634fc16cb07c3b8db389542cc8847/query.go
+// because they are not exported by the module, and we would like to use them in building the githubv4mock test utility.
+//
```
SamMorrowDrums, line 3: I feel like eventually we might want a mock library to exist in standalone form, so we could pull this out, but it's OK to coexist here for now

internal/ghmcp/server.go
```diff
@@ -213,3 +240,141 @@ func RunStdioServer(cfg StdioServerConfig) error {
...
+	if u.Scheme == "" {
+		return apiHost{}, fmt.Errorf("host must have a scheme (http or https): %s", s)
+	}
+
+	if strings.HasSuffix(u.Hostname(), "github.com") {
+		return newDotcomHost()
+	}
+
+	if strings.HasSuffix(u.Hostname(), "ghe.com") {
+		return newGHECHost(s)
```
SamMorrowDrums, line 354: Does proxima have any equivalent of this? Just curious.

pkg/github/pullrequests.go
```diff
@@ -75,8 +77,123 @@ func GetPullRequest(getClient GetClientFn, t translations.TranslationHelperFunc)
...
+				mcp.Required(),
+				mcp.Description("Branch to merge into"),
+			),
+			mcp.WithBoolean("draft",
+				mcp.Description("Create as draft PR"),
+			),
+			mcp.WithBoolean("maintainer_can_modify",
+				mcp.Description("Allow maintainer edits"),
+			),
+		),
```
SamMorrowDrums, line 117: I hope the GO SDK makes these definitions and parsing much closer to standard JSON parsing into structs as discussed!

SamMorrowDrums (approved):
amazing work, I think without the graphql parts it's not too bug, and those parts are going to be shared by other APIs, so I say shipit!

internal/ghmcp/server.go
williammartin: I don't understand this question sorry

williammartin: Discussed on Slack that this is addressing "proxima".

egaponenko-ma:
@SamMorrowDrums @williammartin Don't you think that removed `add_pull_request_review_comment` has a right to exist as standalone tool? As an example: I can fix a PR comment with AI agent and ask it to reply to those comments.

PR #411, opened 2025-05-19
Alter the description of get-me to reduce redundant repeat calls

SamMorrowDrums:
A lot of times the get-me tool is called multiple times in a row, which is redundant. This attempts to address this, by explaining to the model it will not be updated normally.


williammartin (approved):
LGTM with copilots suggestions (or at least, `includes` typo).

williammartin (approved):

PR #437, opened 2025-05-25
Expose ghmcp package

albertywu:
This PR makes two changes, so we can actually use this MCP server internally at my company:

1) Exposes a public package `pkg/ghmcp`. This public package simply delegates to the private `internal/ghmcp` package for implementation.

2) Adds ability for the client to specify a token provider function. Since the MCP server is a long-lived process, this provides a mechanism to update the token dynamically without needing to stop / restart the MCP server.

### DIFF
--- a/cmd/github-mcp-server/main.go
+++ b/cmd/github-mcp-server/main.go
@@ -5,7 +5,7 @@ import (
 	"fmt"
 	"os"
 
-	"github.com/github/github-mcp-server/internal/ghmcp"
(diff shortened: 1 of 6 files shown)


albertywu:
oops, created this PR against the public repo instead of my fork. Closing.

williammartin:
No worries @albertywu. If you have a request for some of this to be exported, feel free to open an issue. Since we expect breaking changes around our Go symbols, I put a bunch of this in `internal`. I don't expect I'll move it out soon, but a place for the conversation can still be useful. Cheers.

PR #444, opened 2025-05-27
docs: update link for non-insiders install

danielroe:
previously the link was the same for insiders vs non-insiders. I've simply changed `insiders.vscode.dev` to `vscode.dev` inside the link for non-insiders builds.

(I'm assuming this was an oversight.)


SamMorrowDrums:
This was actually not an oversight. The end of the query string for insiders `quality=insiders`, is what makes the difference currently.

@digitarald should this change merge? I will leave that up to you.

SamMorrowDrums:
Received no response however no installation issues have occurred so I will leave as is for now.

PR #445, opened 2025-05-27
Remove test that can panic

williammartin:


SamMorrowDrums (dismissed):

pkg/github/repository_resource_test.go
```diff
@@ -180,21 +180,6 @@ func Test_repositoryResourceContentsHandler(t *testing.T) {
...
-					mock.GetReposContentsByOwnerByRepoByPath,
-				),
-			),
-			requestArgs: map[string]any{
-				"owner": []string{"owner"},
-				"repo":  []string{"repo"},
-				"path":  []string{"src"},
-			},
-			expectedResult: nil,
-			expectError:    "no repository resource content found",
```
williammartin, line 196: Also, this `expectError` field is never actually used because all the tests set `expectedErrorMsg`.

williammartin: Fixed now in https://github.com/github/github-mcp-server/pull/445/files#diff-134064f0cd858eb16a795df5143e87363a3ac3bb4c6a58beddf2dbf1fd689ec3R235-R236

PR #447, opened 2025-05-28
Used typed tool handler for `get_me` tool

williammartin:
# Description

This PR is the first of many that moves handlers to use `NewTypedToolHandler` which is also a step towards the official Go SDK. While reworking this, I also took the opportunity to improve testing confidence.

## E2E Test

```
➜  github-mcp-server git:(wm/get-me-typed-tool) GOMAXPROCS=1 GITHUB_MCP_SERVER_E2E_HOST=https://github.com GITHUB_MCP_SERVER_E2E_TOKEN=$(gh auth token) go test -v -run 'GetMe' -count=1 --tags e2e ./e2e

=== RUN   TestGetMe
=== PAUSE TestGetMe
=== CONT  TestGetMe
    e2e_test.go:80: Building Docker image for e2e tests...
    e2e_test.go:163: Starting Stdio MCP client...
--- PASS: TestGetMe (7.14s)
PASS
ok      github.com/github/github-mcp-server/e2e 7.410s
```

### DIFF
--- a/pkg/github/context_tools.go
+++ b/pkg/github/context_tools.go
@@ -33,7 +33,7 @@ func GetMe(getClient GetClientFn, t translations.TranslationHelperFunc) (mcp.Too
 			return mcp.NewToolResultErrorFromErr("failed to get user", err), nil
 		}
 
-		return marshalledTextResult(user), nil
+		return MarshalledTextResult(user), nil
 	})
 
 	return tool, handler


pkg/toolsets/toolsets_test.go
```diff
@@ -4,14 +4,8 @@ import (
 	"testing"
 )
 
-func TestNewToolsetGroup(t *testing.T) {
+func TestNewToolsetGroupIsEmptyWithoutEverythingOn(t *testing.T) {
```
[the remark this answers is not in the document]
williammartin: Well that's not true because `len` won't panic and even if it was true, that would be fine because the test would reasonably fail.

pkg/github/__toolsnaps__/get_me.snap
```diff
@@ -0,0 +1,17 @@
...
+    "properties": {
+      "reason": {
+        "description": "Optional: the reason for requesting the user information",
+        "type": "string"
+      }
+    },
+    "type": "object"
+  },
+  "name": "get_me"
+}
```
SamMorrowDrums, line 17:
```suggestion
}

```

It doesn't really matter, but I do love EOF newlines being added by automation rather than omitted if it's possible.

pkg/github/context_tools.go
```diff
@@ -2,52 +2,39 @@ package github
...
-				return mcp.NewToolResultError(fmt.Sprintf("failed to get user: %s", string(body))), nil
-			}
-
-			r, err := json.Marshal(user)
-			if err != nil {
-				return nil, fmt.Errorf("failed to marshal user: %w", err)
-			}
+	type args struct{}
+	handler := mcp.NewTypedToolHandler(func(ctx context.Context, _ mcp.CallToolRequest, _ args) (*mcp.CallToolResult, error) {
+		client, err := getClient(ctx)
```
williammartin, line 26: I have some ideas about having a handler function that injects:

```go
type Clients struct {
 rest..
 gql..
}
```

Then all handlers don't have this `getClient` boilerplate. Added a lot of types though, would rather feel a bit more pain.

pkg/github/context_tools.go
```diff
@@ -2,52 +2,39 @@ package github
...
-	return mcp.NewTool("get_me",
-			mcp.WithDescription(t("TOOL_GET_ME_DESCRIPTION", "Get details of the authenticated GitHub user. Use this when a request includes \"me\", \"my\". The output will not change unless the user changes their profile, so only call this once.")),
-			mcp.WithToolAnnotation(mcp.ToolAnnotation{
-				Title:        t("TOOL_GET_ME_USER_TITLE", "Get my user profile"),
-				ReadOnlyHint: toBoolPtr(true),
-			}),
-			mcp.WithString("reason",
-				mcp.Description("Optional: the reason for requesting the user information"),
-			),
+func GetMe(getClient GetClientFn, t translations.TranslationHelperFunc) (mcp.Tool, server.ToolHandlerFunc) {
```
SamMorrowDrums, line 12: Do we want to return an mcp.Tool or have our own struct that gets converted later, I suppose it's an utter pain to make versions of all their structs.

pkg/github/context_tools.go
```diff
@@ -2,52 +2,39 @@ package github
...
-				}
-				return mcp.NewToolResultError(fmt.Sprintf("failed to get user: %s", string(body))), nil
-			}
-
-			r, err := json.Marshal(user)
-			if err != nil {
-				return nil, fmt.Errorf("failed to marshal user: %w", err)
-			}
+	type args struct{}
+	handler := mcp.NewTypedToolHandler(func(ctx context.Context, _ mcp.CallToolRequest, _ args) (*mcp.CallToolResult, error) {
```
SamMorrowDrums, line 25: Nice to see use of the typed handler (although this tool is not much of a showcase 😂). 

I was going to ask the same thing as above, do we want to use `mcp-go` structs here. I suppose we should for now.

A use case I can imagine is wrapping an existing handler with another one (to add something like chunking, or other aggregations like returning only the count). That sad part with the string output is that instead of Marshalling the data just-in-time, I would have to unmarshall the json again to process it in this form. 😓 this is a tough problem to solve without having to just re-write tools.

I had imagined we'd have a tool handler type that returns an interface (can be marshalled into json whatever that interface is called), and error. We would then give that a thin wrapper that binds it into `mcp-go`. Maybe overkill for now, just something that is coming onto my radar.

pkg/github/server.go
```diff
@@ -214,3 +215,12 @@ func OptionalPaginationParams(r mcp.CallToolRequest) (PaginationParams, error) {
 		perPage: perPage,
 	}, nil
 }
+
+func marshalledTextResult(v any) *mcp.CallToolResult {
```
SamMorrowDrums:
```suggestion
func MarshalledTextResult(v any) *mcp.CallToolResult {
```

I think we should export this.

pkg/github/context_tools.go
```diff
@@ -2,52 +2,39 @@ package github
...
+			return mcp.NewToolResultErrorFromErr("failed to get GitHub client", err), nil
+		}
 
-			return mcp.NewToolResultText(string(r)), nil
+		user, _, err := client.Users.Get(ctx, "")
+		if err != nil {
+			return mcp.NewToolResultErrorFromErr("failed to get user", err), nil
 		}
+
+		return marshalledTextResult(user), nil
```
SamMorrowDrums:
```suggestion
		return MarshalledTextResult(user), nil
```

SamMorrowDrums (dismissed):
I think you just left a comment that agrees with my comment about structs, broadly I agree with this direction, and some nits, and a general question about how much we extract our types from `mcp-go` types.

pkg/github/__toolsnaps__/get_me.snap
williammartin: I can add it, but I don't really understand why it's useful to you? Does it help in your IDE or something?

pkg/github/context_tools.go
williammartin: I want our own types that can be adapted for mcp-go/go-sdk but I'm not paying that price right now. I want to do the move over to the typed handlers and then iterate from there.

pkg/github/context_tools.go
williammartin: All good thoughts and as per https://github.com/github/github-mcp-server/pull/447#discussion_r2111969596 there's lots I do want to tackle but I want to do it piecemeal where we move over to having a struct for args, then we can look at the abstractions that make sense around that, rather than biting it all off at once.

SamMorrowDrums (approved):

PR #456, opened 2025-05-30
Bump go-github to v72.0.0

radar07:
Closes: #455

PR #467, opened 2025-06-03
Fix missing go-github v72 license bump

williammartin:
# Description

Backfills a license bump that should have happened as part of https://github.com/github/github-mcp-server/pull/456, not totally sure what happened there.


williammartin:
Bypassing required review because it's a license bump.

PR #484, opened 2025-06-05
feat: add type to issues

radar07:
Closes: #409

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -13,3 +13,7 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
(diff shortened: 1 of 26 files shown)


SamMorrowDrums:
Thank you for trying this.

I am concerned that there isn't a good story for how the LLM should discover what options are available for this field.

I'd propose an enum, but users can add additional types so that would be insufficient.

We will have to think about this.

SamMorrowDrums:
Thank you for trying this.

I am concerned that there isn't a good story for how the LLM should discover what options are available for this field.

I'd propose an enum, but users can add additional types so that would be insufficient.

We will have to think about this.

radar07:
I agree. We also might need to support `IssueType` in this server.

Supporting [CRUD operations of IssueType](https://docs.github.com/en/rest/orgs/issue-types?apiVersion=2022-11-28) would assist the model to discover the list of types and assigning them.

akenneth:
👋 

- This proposed PR adds issue types discoverability. Together with the PR proposed here it should add nice issue type support. https://github.com/github/github-mcp-server/pull/616

akenneth:
- Updating that my PR to add read issue types tool was merged https://github.com/github/github-mcp-server/pull/616

Maybe this one is ready to go now?

dsyme:
Duplicate of #869

PR #485, opened 2025-06-05
Reduce fields in `search_users` response

SamMorrowDrums:
There is little use in most of the fields returned by the `search_users` function, as they are reproducible easily by knowing the user handle, or not relevant. The profile URL is useful as it encourages the LLM to format the handle as a link, which is good, I think.

image

Here is an example showing that the "missing" data is easily re-constructed by the LLM:
 
image

### DIFF
--- a/pkg/github/search.go
+++ b/pkg/github/search.go
@@ -146,6 +146,19 @@ func SearchCode(getClient GetClientFn, t translations.TranslationHelperFunc) (to
 		}
 }
 
+type MinimalUser struct {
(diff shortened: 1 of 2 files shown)


SamMorrowDrums:
@williammartin and @toby I am going ahead with this for reasons I mentioned to Toby last night, but I checked and even if I ask for ordering by follower count we don't get it in the response, so think the published schema is maybe wrong for the search API? Other user APIs might well provide them.

PR #487, opened 2025-06-06
chore: separate toolset creation from init and use typed error

SamMorrowDrums:
This PR makes it easier to identify the error when a toolset does not exist, and also separates toolset creation from initialization, so it's possible to compose toolsets in more ways.


omgitsads (approved):
👍🏻

PR #488, opened 2025-06-06
fix: move defaulted 'state' param after non-default params in list_code_scanning_alerts

ph3t:
Fixes `SyntaxError: parameter without a default follows parameter with a default` error when using github-mcp-server with the Python MCP SDK

Closes: N/A


SamMorrowDrums (approved):

SamMorrowDrums:
Interesting, is this formally part of the spec or a quirk of the Python client SDK?

ph3t:
> Interesting, is this formally part of the spec or a quirk of the Python client SDK?

@SamMorrowDrums  Turns out the issue was in the ADK ([nerve](https://github.com/evilsocket/nerve)) I'm using, so this change didn't fix anything. But basically, Python expects functions with default args to be last. I have opened a new PR as an apology for this useless fix: https://github.com/github/github-mcp-server/pull/494 😅

PR #495, opened 2025-06-09
Export `ToBoolPtr` and `RequiredParam`

LuluBeatson:
1. Export `ToBoolPtr` and `RequiredParam` functions from the github package.
2. Clean up the type assertion in `RequiredParam`


SamMorrowDrums (approved):

williammartin:
Just a couple of notes on this @LuluBeatson @SamMorrowDrums 

I didn't export `toBoolPtr` because `github.ToBoolPtr` is pretty janky. It's not a meaningful abstraction. It would likely be better for us to expose a `ghmcp.Annotations` which uses types we care about and has a single adapter into the mcp server types.

Secondly, I _strongly_ encourage avoiding use of all the param parsing functions from this package. The jsonschema should be the source of truth. Using these takes the implementation away from where we should want it to go IMO.

That said, it's not like it's a huge cost to exporting and using these compared to where we are today. I just wanted to let you know my thinking on this.

PR #497, opened 2025-06-10
feat: add raw parameter to get_file_contents tool

### LINKED ISSUE #372: get_file_contents tool should have non-base64 option

The `get_file_contents` tool currently returns file contents in base64. This matches the backing API, but it'd be nice to have a non-base64 option for this. The corresponding `create_or_update_file` tool accept plaintext, even though the backing API only allows base64.

### Proposed solution

This would create interface consistency and probably lead to better model performance with the tool.

### Additional context

I'm happy to send a PR for this if there's a recommended approach I should take!

ko1ynnky:
## Summary

Adds an optional `raw` boolean parameter to the `get_file_contents` tool to return plain text content instead of base64 encoded content, addressing issue #372.

### Changes
- ✅ Added `raw` boolean parameter to `get_file_contents` tool schema
- ✅ When `raw=true`, fetches file content via GitHub's download URL and returns plain text
- ✅ When `raw=false` or unset, maintains existing base64 encoded behavior  
- ✅ Returns appropriate error for directories when `raw=true`
- ✅ Handles files without download URLs gracefully
- ✅ Added comprehensive test coverage for edge cases

### Benefits
- **Reduces token consumption** for LLM interactions by avoiding base64 decoding
- **Improves model performance** by providing direct access to file content
- **Maintains backward compatibility** with existing implementations
- **Provides interface consistency** with other tools like `create_or_update_file`

### Test Coverage
- Schema validation includes `raw` parameter
- Directory handling with `raw=true` returns appropriate error
- Files without download URL return appropriate error
- All existing tests continue to pass

### Example Usage
```bash
# Get base64 encoded content (existing behavior)
get_file_contents --owner github --repo docs --path README.md

# Get raw text content (new feature)  
get_file_contents --owner github --repo docs --path README.md --raw true
```

Closes #372

### DIFF
--- a/pkg/github/repositories.go
+++ b/pkg/github/repositories.go
@@ -431,6 +431,9 @@ func GetFileContents(getClient GetClientFn, t translations.TranslationHelperFunc
 			mcp.WithString("branch",
 				mcp.Description("Branch to get contents from"),
 			),
+			mcp.WithBoolean("raw",
+				mcp.Description("Return raw file contents instead of base64 encoded (only applies to files, not directories)"),
+			),
 		),
 		func(ctx context.Context, request mcp.CallToolRequest) (*mcp.CallToolResult, error) {
 			owner, err := RequiredParam[string](request, "owner")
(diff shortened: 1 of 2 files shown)


SamMorrowDrums:
Thank you for the contribution. The file API just doesn't work well for this, so I've made this: https://github.com/github/github-mcp-server/pull/505

I'm going to close out this one in favour of mine, as I think it is simpler, and was much needed. :pray:

Issue #482, opened 2025-06-05
Notifications and Secret Protection missing from toolsets docs


SamMorrowDrums:
We have two toolsets that are not advertised. We should add them. Ultimately, we should automate these docs.

PR #499, opened 2025-06-10
Add context toolset and adjust readme

SamMorrowDrums:
Adds the missing toolsets to docs, and also makes the context tools a toolset in their own right - as the behaviour is a little strange that it sits outside the toolsets user controls.

As the namespaces concept appears to have more traction that toolsets, we will expand the concept to encompass resources for now, so that you can run multiple servers with different toolsets without a break in functionality.

This change partially aligns us with the [namespaces spec proposal](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/334).

Closes: https://github.com/github/github-mcp-server/issues/482


pkg/github/search.go
```diff
@@ -209,6 +209,7 @@ func SearchUsers(getClient GetClientFn, t translations.TranslationHelperFunc) (t
 			}
 
 			client, err := getClient(ctx)
+
```
SamMorrowDrums:
```suggestion
```

PR #505, opened 2025-06-12
use better raw file handling and return resources

SamMorrowDrums:
Closes https://github.com/github/github-mcp-server/issues/372

This PR attempts to address some initially poor choices for how do handle files and resources.

- Adds an API adaptor for raw file API
- uses text response for text files, and base64 response for binary
- Adds URL parsing so it works in all supported environments
- Uses resource responses (getting much better results in VSCode for example), we already have the resource handles for these and resources are the right pattern for this.
- Drops directory handling from the template resources (it just doesn't work well)
- Keeps the directory handling as a fallback for the tool version of get repo content.
- Adds mocks for the new URI patterns
- Tests it all

image

image

image

https://github.com/github/github-mcp-server/pull/497
CC @ko1ynnky I decided to go a bit further and getting very good results now. Thanks for the encouragement.


tonytrg (dismissed):

ko1ynnky:
@SamMorrowDrums 
Thanks Sam! Great work taking this much further - your solution is way better than mine. Glad I could provide some encouragement to get the ball rolling!

PR #534, opened 2025-06-17
Update README.md

PierreGode:
Fix Uncaught Exception issue, trailing comma in documentation.


SamMorrowDrums (approved):
I will accept the change for cleaner json, but VSCode itself supports jsonc with comments and trailing commas.

PR #538, opened 2025-06-17
Change Anthropic to Github in mcpurl README

dgiacomo:
Noticed mcpurl README referenced Anthropic's MCP when I think it's intended to be Github's MCP. Maybe a copy pasta - super easy change.


SamMorrowDrums (approved):

SamMorrowDrums:
Anthropic built the initial implementation, and I think at first we compared outputs using this to ensure we weren't diverging. Probably a legacy of that.

PR #558, opened 2025-06-20
get_file_contents accepts a ref instead of just branch

tonytrg:
Closes:

### DIFF
--- a/pkg/github/repositories.go
+++ b/pkg/github/repositories.go
@@ -8,6 +8,7 @@ import (
 	"io"
 	"net/http"
 	"net/url"
+	"strconv"
 	"strings"
 
 	"github.com/github/github-mcp-server/pkg/raw"
@@ -432,8 +433,11 @@ func GetFileContents(getClient GetClientFn, getRawClient raw.GetRawClientFn, t t
 				mcp.Required(),
 				mcp.Description("Path to file/directory (directories must end with a slash '/')"),
 			),
-			mcp.WithString("branch",
-				mcp.Description("Branch to get contents from"),
+			mcp.WithString("ref",
+				mcp.Description("Accepts optional git refs such as `refs/tags/<tag>`, `refs/heads/<branch>` or `refs/pull/<pr_number>/head`"),
+			),
+			mcp.WithString("sha",
+				mcp.Description("Accepts optional git sha, if sha is specified it will be used instead of ref"),
 			),
 		),
 		func(ctx context.Context, request mcp.CallToolRequest) (*mcp.CallToolResult, error) {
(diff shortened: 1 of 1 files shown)


almaleksia:
Howdy 👋 

Can we make the change backwards compatible because we depend on this tool: https://github.com/github/sweagentd/pull/3357

pkg/github/__toolsnaps__/get_file_contents.snap
```diff
@@ -18,9 +14,17 @@
         "description": "Path to file/directory (directories must end with a slash '/')",
         "type": "string"
       },
+      "ref": {
+        "description": "Accepts optional git refs such as `refs/tags/\u003ctag\u003e`, `refs/heads/\u003cbranch\u003e` or `refs/pull/\u003cpr_number\u003e/head`",
```
SamMorrowDrums:
```suggestion
        "description": "Accepts optional git refs such as `refs/tags/<tag>`, `refs/heads/<branch>` or `refs/pull/<pr_number>/head`",
```
This seems broken, these should just be regular angled brackets no? They shouldn't need Unicode markup.

tonytrg: I updated the snapshot generation/json marshalling to not escape `<>`, it seems we already get the escaped strings from the tool. So I think there is nothing we can do.

williammartin: @tonytrg I think we maybe want to be setting https://pkg.go.dev/encoding/json#Encoder.SetEscapeHTML to `false` since this is unrelated to html? Totally not a usecase I thought about.

williammartin: That said, it's probably worth thinking about what the MCP server implementation is actually doing here. Maybe the clients really are getting escaped JSON.

pkg/github/__toolsnaps__/get_file_contents.snap
```diff
@@ -6,10 +6,6 @@
   "description": "Get the contents of a file or directory from a GitHub repository",
   "inputSchema": {
     "properties": {
-      "branch": {
```
williammartin, line 9: Just as an aside, I think it's really nice to look at the snapshot to see the structural changes that have occurred without looking at the imperative tool handler.

pkg/github/__toolsnaps__/get_file_contents.snap
tonytrg: @williammartin I changed the encoder to set it to false, it somehow keeps escaping it, which is weird.
We changed the symbol from `<>` to `{}`.

pkg/github/__toolsnaps__/get_file_contents.snap
tonytrg: Agree, just looking at the files commit history is pretty great way to check user faced changes.

pkg/github/__toolsnaps__/get_file_contents.snap
tonytrg: thanks to @williammartin for digging https://github.com/mark3labs/mcp-go/blob/8f5b048218f6d044c3322f16b8cb0b08e10bf5d0/mcp/tools.go#L538

pkg/github/__toolsnaps__/get_file_contents.snap
SamMorrowDrums: I think we will use this for 1p clients to help them understand the diff.

SamMorrowDrums (approved):

PR #570, opened 2025-06-23
Add ctx propagation to raw api request

kerobbi:
Update `newRequest` to accept a `context.Context` param and attach it to the generated `http.Request` using `withContext`.

This change ensures that trace contexts are correctly propagated, supporting distributed tracing in the remote MCP server.

### DIFF
--- a/pkg/raw/raw.go
+++ b/pkg/raw/raw.go
@@ -25,9 +25,13 @@ func NewClient(client *gogithub.Client, rawURL *url.URL) *Client {
 	return &Client{client: client, url: rawURL}
 }
 
-func (c *Client) newRequest(method string, urlStr string, body interface{}, opts ...gogithub.RequestOption) (*http.Request, error) {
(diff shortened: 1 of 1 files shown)


SamMorrowDrums (commented):
Just to validate, does this work if you try it locally? I assume so, but an alternative is to take the current request context and then extend it and add it back to the request.

kerobbi:
> Just to validate, does this work if you try it locally? I assume so, but an alternative is to take the current request context and then extend it and add it back to the request.

I did test this locally and confirmed that ctx propagation is working as expected, trace info and parent-child spans are all showing up correctly.

I'm a bit confused about the suggested alternative. Since we are creating a new outbound request here (not working with an incoming HTTP request), I don't think there's an existing req ctx to extend? Let me know if I'm missing something.

SamMorrowDrums (approved):

PR #571, opened 2025-06-23
feat: add support for sse

ruthakur:
Description: Fix vulnerability and add support for SSE

Changes:
1. Adds supports for SSE along with STDIO. The SSE server runs on port 8080.
2. Upgrades golang version to 1.23.8 to fix the OSS vulnerability in the stdlib package of go1.23.7

Tests:
1. Unit tests have not been added for the change
2. The change is functionally tested and working fine.


tonytrg:
@ruthakur thank you for the contribution, but `SSE` mode is not supported, we might look into supporting `streamable http` on local.

PR #605, opened 2025-06-28
feat: add include_sha parameter to get_file_contents tool

### LINKED ISSUE #595: Tool `get_file_contents` is missing the required `sha` in its response

yonaka15:
## Summary

This PR addresses issue #595 by adding an optional `include_sha` parameter to the `get_file_contents` tool, providing a more flexible and performant solution than the previous approach in PR #599.

## Problem

The `get_file_contents` tool was missing the SHA hash in its response, which is essential for subsequent file update operations using `create_or_update_file`. The SHA is required by the GitHub API for any update operation.

## Solution

Instead of always returning metadata (as in PR #599), this PR adds an `include_sha` boolean parameter that allows users to explicitly choose between:
- **Raw content** (default): Fast retrieval via GitHub's Raw API
- **File metadata**: Complete file information including SHA, size, type via Contents API

## Key Features

- ✅ **Backward compatible**: Default behavior unchanged
- ✅ **Performance optimized**: Raw API used when metadata not needed
- ✅ **Flexible**: Works for both files and directories
- ✅ **Comprehensive**: Returns complete GitHub API metadata when requested
- ✅ **Well-tested**: Full test coverage for all scenarios

## Usage Examples

```javascript
// Get file content (existing behavior)
get_file_contents({ owner: "user", repo: "repo", path: "file.md" })
// Returns: raw file content

// Get file metadata including SHA
get_file_contents({ 
  owner: "user", 
  repo: "repo", 
  path: "file.md", 
  include_sha: true 
})
// Returns: { name, path, sha, size, type, content, encoding, ... }
```

## Changes Made

- Added `include_sha` boolean parameter to tool schema
- Updated implementation to conditionally use Contents API vs Raw API
- Enhanced tool description to document new parameter
- Added comprehensive test coverage for all scenarios
- Updated tool snapshots to reflect schema changes

## Testing

- ✅ All existing tests pass
- ✅ New test cases cover include_sha functionality
- ✅ Edge cases and error scenarios tested
- ✅ Both file and directory metadata retrieval verified

## Supersedes

This PR supersedes #599 by providing a more user-friendly and performant solution to the same problem.

Closes #595
Supersedes #599


LuluBeatson (changes requested):
Hi @yonaka15, thank you for your PR, returning the SHA will be useful! :)

We recommend the following changes:
- Not adding a new argument `include_sha`, but always returning a file's SHA
- For files, always return a resource. The SHA can be included with the text output of the tool result but the file should be a resource. This is done on [repositories.go#L566-L578](https://github.com/github/github-mcp-server/blob/ea7304769f9d48c1e603cac589dfebbc2c641ef4/pkg/github/repositories.go#L566-L578)

So overall, the flow for files can look like:
1. First, get the SHA if not provided as a tool input. For this you can use a cheaper method, like we do for PRs ([repositories.go#L516](https://github.com/github/github-mcp-server/blob/ea7304769f9d48c1e603cac589dfebbc2c641ef4/pkg/github/repositories.go#L516)) instead of the file API.
2. Then, download the file using `rawClient.GetRawContent`, providing the SHA (to avoid a race condition). This is already done on [repositories.go#L531](https://github.com/github/github-mcp-server/blob/ea7304769f9d48c1e603cac589dfebbc2c641ef4/pkg/github/repositories.go#L531)
3. Return a `NewToolResultResource` with the file as a resource and a json message containing the SHA. i.e. You can update the "successfully downloaded text file" / "successfully downloaded binary file" text part of the tool result but keep the `TextResourceContents` / `BlobResourceContents` in [repositories.go#L566-L578](https://github.com/github/github-mcp-server/blob/ea7304769f9d48c1e603cac589dfebbc2c641ef4/pkg/github/repositories.go#L566-L578)

<details><summary>Example: Tool call result with both text and resource output</summary>

Image

</details>


<details><summary>Example: Tool call result with file content encoded as text (avoid this)</summary>

Image

</details>


Thank you, we look forward to your response!

yonaka15:
Based on the feedback received, I've implemented a better approach in #676 that addresses the concerns about the `include_sha` parameter.

The new implementation automatically includes SHA information in all responses without requiring users to specify an additional parameter, making it simpler and more intuitive to use.

Closing this PR in favor of #676.

Issue #608, opened 2025-06-29
Enhancement: Add log truncation options to `get_job_logs` tool for improved LLM context efficiency


kehao95:
## Summary

The `get_job_logs` tool currently returns complete job logs, which can be extremely long and inefficient when used as MCP tool context for LLMs. Adding log truncation options like `tail_lines`  parameters would help focus on relevant failure information.

## Problem

When debugging CI/CD failures using the `get_job_logs` tool in LLM conversations:

1. **Context Overflow**: Complete job logs can easily exceed LLM context limits
2. **Information Overload**: Most debugging scenarios only need the final portion where failures occur
4. **Poor User Experience**: Long logs make it harder to identify specific failure points

## Proposed Solution

Add optional truncation parameters to the `get_job_logs` tool:

- `tail_lines`: Return last N lines (most useful for failure debugging)

Example usage:
```json
{
  "name": "get_job_logs",
  "arguments": {
    "owner": "example",
    "repo": "repo", 
    "job_id": 12345,
    "tail_lines": 100,
    "return_content": true
  }
}
```

This enhancement aligns with existing concerns about LLM context efficiency, similar to Issue #142 regarding excessive context size.

SamMorrowDrums:
We really agree with this. We have been working on similar, and I had a similar though on recreating the `tail` command as the most common place to find the actual failure reason. You can raise a PR, but we will hopefully fix this ASAP, because it is a significant issue. Thanks for reporting.

923417809125:
Tanveerhussain

PR #615, opened 2025-06-30
Add tail logs option

JoannaaKL:
This PR adds a new option to tail logs. 
Closes: https://github.com/github/github-mcp-server/issues/608

### DIFF
--- a/pkg/github/actions.go
+++ b/pkg/github/actions.go
@@ -640,10 +640,10 @@ func GetJobLogs(getClient GetClientFn, t translations.TranslationHelperFunc) (to
 
 			if failedOnly && runID > 0 {
 				// Handle failed-only mode: get logs for all failed jobs in the workflow run
-				return handleFailedJobLogs(ctx, client, owner, repo, int64(runID), returnContent, tailLines)
+				return handleFailedJobLogs(ctx, client, owner, repo, int64(runID), returnContent, int16(tailLines))
 			} else if jobID > 0 {
 				// Handle single job mode
-				return handleSingleJobLogs(ctx, client, owner, repo, int64(jobID), returnContent, tailLines)
+				return handleSingleJobLogs(ctx, client, owner, repo, int64(jobID), returnContent, int16(tailLines))
 			}
 
 			return mcp.NewToolResultError("Either job_id must be provided for single job logs, or run_id with failed_only=true for failed job logs"), nil
@@ -651,7 +651,7 @@ func GetJobLogs(getClient GetClientFn, t translations.TranslationHelperFunc) (to
 }
 
 // handleFailedJobLogs gets logs for all failed jobs in a workflow run
-func handleFailedJobLogs(ctx context.Context, client *github.Client, owner, repo string, runID int64, returnContent bool, tailLines int) (*mcp.CallToolResult, error) {
+func handleFailedJobLogs(ctx context.Context, client *github.Client, owner, repo string, runID int64, returnContent bool, tailLines int16) (*mcp.CallToolResult, error) {
 	// First, get all jobs for the workflow run
 	jobs, resp, err := client.Actions.ListWorkflowJobs(ctx, owner, repo, runID, &github.ListWorkflowJobsOptions{
 		Filter: "latest",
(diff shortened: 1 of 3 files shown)


pkg/github/actions.go
```diff
@@ -775,6 +787,16 @@ func downloadLogContent(logURL string) (string, *http.Response, error) {
...
 	logContent := strings.TrimSpace(string(content))
+
+	// Truncate to tail_lines if specified
+	if tailLines > 0 {
+		lines := strings.Split(logContent, "\n")
+		if len(lines) > tailLines {
+			lines = lines[len(lines)-tailLines:]
+			logContent = strings.Join(lines, "\n")
+		}
+	}
```
SamMorrowDrums: Nerdy, but you could build from reverse parsing the string maybe, for efficiency perhaps?

```
if tailLines > 0 {
    lineCount := 0
    lastNewlinePos := len(logContent)
    
    // Count backwards to find the nth newline from the end
    for i := len(logContent) - 1; i >= 0 && lineCount < tailLines; i-- {
        if logContent[i] == '\n' {
            lineCount++
            if lineCount == tailLines {
                logContent = logContent[i+1:]
                break
            }
        }
    }
}
```

SamMorrowDrums (commented):
This already looks really close. I think we might want to return the total lines anyway, just to contextualize the value, but we can look at that later.

pkg/github/actions.go
SamMorrowDrums: Could also add an offset from the bottom, so repeat calls could search upwards without getting the same data twice?

JoannaaKL: Yeah it's a good call not to split the log content and move from backwards. Regarding `lastNewlinePos` - I am not sure how that could be used, right now the function keeps no state, I am not sure how adding this variable could help.

pkg/github/actions.go
```diff
@@ -584,6 +584,10 @@ func GetJobLogs(getClient GetClientFn, t translations.TranslationHelperFunc) (to
 			mcp.WithBoolean("return_content",
 				mcp.Description("Returns actual log content instead of URLs"),
 			),
+			mcp.WithNumber("tail_lines",
+				mcp.Description("Number of lines to return from the end of the log"),
+				mcp.DefaultNumber(50),
```
kehao95: We can encourage LLM using this option in the description. But I think the default behavior should be return all if `tail_lines` not specified?

kehao95:
Also in my personal test default to 50 seems too low that usually only reach to post-action steps.
Is there better way that we can return only the tail log from failed step? ( might not be the scope of this issue though)

JoannaaKL:
> Also in my personal test default to 50 seems too low that usually only reach to post-action steps. Is there better way that we can return only the tail log from failed step? ( might not be the scope of this issue though)

I bumped it to 500, which now might be too high but let's see. We could think of making the search more powerful and return only meaningful log lines, but it's out of scope for this pr. :)

SamMorrowDrums:
> Also in my personal test default to 50 seems too low that usually only reach to post-action steps.
> 
> Is there better way that we can return only the tail log from failed step? ( might not be the scope of this issue though)

Yeah I tried quite a few interesting options, but last n lines is better than nothing as we plan to release tomorrow. We can iterate further though.

pkg/github/actions.go
```diff
@@ -757,25 +770,46 @@ func getJobLogData(ctx context.Context, client *github.Client, owner, repo strin
...
+func trimContent(content string, tailLines int) (string, int) {
+	// Truncate to tail_lines if specified
+	lineCount := 0
+	if tailLines > 0 {
+
+		// Count backwards to find the nth newline from the end
+		for i := len(content) - 1; i >= 0 && lineCount < tailLines; i-- {
+			if content[i] == '\n' {
+				lineCount++
+				if lineCount == tailLines {
```
SamMorrowDrums, line 806: It might be worth a comment here explaining that there isn't an exit condition because we want the total line count.

pkg/github/actions_test.go
```diff
@@ -885,6 +885,64 @@ func Test_GetJobLogs(t *testing.T) {
...
+				"repo":        "repo",
+				"run_id":      float64(456),
+				"failed_only": true,
+				"tail_lines":  float64(1),
+			},
+			expectError: false,
+			checkResponse: func(t *testing.T, response map[string]any) {
+				assert.Equal(t, float64(456), response["run_id"])
+				assert.Equal(t, float64(3), response["total_jobs"])
+				assert.Equal(t, float64(2), response["failed_jobs"])
```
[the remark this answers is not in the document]
Ebaoone: G

Ebaoone (changes requested):
Mo

PR #628, opened 2025-07-02
Cleanup

JoannaaKL:
This pr:
- removes redundant function 
- adds `script/test` and calls it from the main go workflow


avin022:
hai..

mendozafelipe1021-a11y:
> This pr:
> 
> * removes redundant function
> * adds `script/test` and calls it from the main go workflow

PR #666, opened 2025-07-11
fix: get_file_contents use "/" for root

LuluBeatson:
Problem: To get root directory, agent was calling `get_file_contents` with path `""` which is not allowed for a required param.

Fix: Updated `path` param description:
> Path to file/directory. Directories, including root, must end with '/'

<details><summary>Screenshot</summary>

Claude Sonnet 3.5
```
get the contents of the github/github-mcp-server repo
```

image

</details>

### DIFF
--- a/README.md
+++ b/README.md
@@ -899,7 +899,7 @@ export GITHUB_MCP_TOOL_ADD_ISSUE_COMMENT_DESCRIPTION="an alternative description
 
 - **get_file_contents** - Get file or directory contents
   - `owner`: Repository owner (username or organization) (string, required)
-  - `path`: Path to file/directory (directories must end with a slash '/') (string, required)
+  - `path`: Path to file/directory. Directories, including root, must end with '/' (string, required)
   - `ref`: Accepts optional git refs such as `refs/tags/{tag}`, `refs/heads/{branch}` or `refs/pull/{pr_number}/head` (string, optional)
   - `repo`: Repository name (string, required)
   - `sha`: Accepts optional commit SHA. If specified, it will be used instead of ref (string, optional)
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -899,7 +899,7 @@ export GITHUB_MCP_TOOL_ADD_ISSUE_COMMENT_DESCRIPTION="an alternative description
 
 - **get_file_contents** - Get file or directory contents
   - `owner`: Repository owner (username or organization) (string, required)
-  - `path`: Path to file/directory (directories must end with a slash '/') (string, required)
+  - `path`: Path to file/directory. Directories, including root, must end with '/' (string, required)
```
tonytrg:
```suggestion
  - `path`: Path to file/directory. Directories must end with '/'. Root equals '/' (string, required)
```

Not a great opinion on that, feel free to disregard.

pkg/github/repositories.go
```diff
@@ -464,7 +464,7 @@ func GetFileContents(getClient GetClientFn, getRawClient raw.GetRawClientFn, t t
 			),
 			mcp.WithString("path",
 				mcp.Required(),
-				mcp.Description("Path to file/directory (directories must end with a slash '/')"),
+				mcp.Description("Path to file/directory. Directories, including root, must end with '/'"),
```
tonytrg: What do you think about:
- a manual substitution in code of "" with "/".
  - Which means we make assumptions on an empty input that the caller probably means root path, but would that be preferable to an error. 
-  path optional and default is / == root.
- a descriptive error message:
  - "Empty path not allowed, did you mean root path:/"

LuluBeatson:
Thank you @tonytrg. I've decided to go your suggestion of making the path optional with "/" default - I think this most reliable

Issue #818, opened 2025-08-04
Add org-level discussions support to list_discussion_categories tool


tommaso-moro:
At the moment, the `list_discussion_categories` tool doesn't support org-level discussions (e.g. "What are the discussion categories in modelcontextprotocol?" won't work as expected).

A change similar to what was introduced in https://github.com/github/github-mcp-server/pull/775 for `list_discussions` is needed.

Without it, the model will often guess a random repo value.

Example of the problem:

<img width="357" height="1061" alt="Image" src="https://github.com/user-attachments/assets/35d33d33-9287-4bb4-837a-74e92e96ecb5" />

PR #775, opened 2025-07-28
Add support for org-level discussions in list_discussions tool

tommaso-moro:
### Overview
This PR adds support for org-level discussions to the `list_discussions`, thus closing issue #671 

For the sake of not bloating the MCP with too many (similar) tools, I made the design decision to update the existing `list_discussions` tool to add support for org-level discussions: if a `repo` is provided, the repo-level discussions are queried, otherwise the org-level discussions are.

**Changes**

- The `repo` parameter is made optional. When it is not passed, `repo` will be set to `.github` which will query discussions at the organisation level. (This is because organization discussions appear to be stored in a special `.github` repository within the organization).
- The tool descriptions are updated to accommodate for the above change
- The tests are updated to test for the scenario where the tool is used to query org-level discussions

**Screenshots**
Screenshot 2025-07-28 at 16 44 41
Screenshot 2025-07-28 at 16 45 02
Screenshot 2025-07-28 at 16 47 09
Screenshot 2025-07-28 at 18 15 54
Screenshot 2025-07-28 at 18 16 18

PR #819, opened 2025-08-04
Add support for org-level discussions in list_discussion_categories tool

tommaso-moro:
### Overview
Closes: https://github.com/github/github-mcp-server/issues/818

This PR improves the existing `list_discussion_categories` tool, by adding logic to handle organisation-level discussions. (It implements similar logic to what I implemented in [PR #775](https://github.com/github/github-mcp-server/pull/775) ) 

### Changes introduced
Before, the "repo" argument was required, so when the user would ask to query org-level discussions the model would either (i) hallucinate random values for `repo` or (ii) it would have to know the implementation detail about the Github API which is that org-level discussions are stored in a hidden `.github` repo.
With these changes:
- the `repo` argument is made optional
- the tool description is updated so that the model knows to omit `repo` if discussions are to be queried at the org-level
- the `repo` value is deterministically set to `.github` when needed
- the tests are updated to reflect these changes 

**Before (the model hallucinates repo values)**
Screenshot 2025-08-04 at 17 14 44

**After (the model correctly omits the repo argument, and the tool uses the deterministic flow to query org-level discussions)**
Screenshot 2025-08-04 at 17 18 12

### DIFF
--- a/pkg/github/discussions.go
+++ b/pkg/github/discussions.go
@@ -443,7 +443,7 @@ func GetDiscussionComments(getGQLClient GetGQLClientFn, t translations.Translati
 
 func ListDiscussionCategories(getGQLClient GetGQLClientFn, t translations.TranslationHelperFunc) (tool mcp.Tool, handler server.ToolHandlerFunc) {
 	return mcp.NewTool("list_discussion_categories",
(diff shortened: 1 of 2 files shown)


mattdholloway (dismissed):
lgtm

PR #833, opened 2025-08-07
Migrate list_issues tool from REST to GraphQL API

### LINKED ISSUE #790: Provide a top level function for fetching non-PR issues

Currently, when I ask to list most recent issues, it uses "List issues".
Since my repo has many recent PRs, it only fetches PRs (whole first page) and never finds any actual issues. 
It doesn't seem to know how to fetch non-PR issues, so it seems I'd have to get it to paginate through list issues several times.
I think it'd benefit folks to separate Issues from PRs, as many of us would like to fetch issues separately.. and also in the UI, only issues are called "Issues".

### Proposed solution

Either break "List issues" into "List issues" and "List pull requests" or add an argument for type issue/pr.

### Example prompts or workflows (for tools/toolsets only)

"List most recent issue in [repo]"

"Summarize 5 recent issues in [repo]"

### Additional context

Image

Image

MattBabbage:
## Summary

This PR migrates the `list_issues` tool from using GitHub's REST API to GraphQL API to improve performance and provide better pagination support.

## Changes Made

- **Migrated list_issues tool to GraphQL**: Replaced REST API calls with GraphQL queries for better performance and more efficient data fetching
- **Added dynamic label filtering**: Implemented conditional GraphQL queries that support optional label filtering
- **Updated pagination**: Moved from page-based pagination to cursor-based pagination using GraphQL
- **Added new data structures**: 
  - `IssueFragment` for GraphQL response mapping
  - `IssueQueryResult` interface for query type abstraction
  - `ListIssuesQueryType` and `ListIssuesQueryNoLabelsType` for conditional queries
- **Improved data mapping**: Added `fragmentToIssue` function to convert GraphQL responses to GitHub API format

## Key Improvements

1. **Performance**: GraphQL allows fetching only required fields, reducing payload size
2. **Better pagination**: Cursor-based pagination is more reliable than page-based for large datasets
3. **Conditional queries**: Dynamic query generation based on whether labels are provided
4. **Consistent API**: Maintains compatibility with existing response format while using GraphQL under the hood

## Files Changed

- `pkg/github/issues.go`: Main implementation of GraphQL-based list_issues tool
- `pkg/github/queries.json`: Added GraphQL query definitions
- `pkg/github/tools.go`: Updated tool registration

### GraphQL Query Structure
The implementation uses two different GraphQL query types:
- `ListIssuesQueryType`: For queries with label filtering
- `ListIssuesQueryNoLabelsType`: For queries without label filtering

This conditional approach ensures optimal query structure and avoids GraphQL errors when labels parameter is not provided.

### Pagination Migration
- **Before**: REST API with page-based pagination (`page`, `per_page`)
- **After**: GraphQL with cursor-based pagination (`after`, `first`)
- **Backward compatibility**: Tool description updated to guide users on new pagination approach

### Data Transformation
The `fragmentToIssue` function ensures that GraphQL responses are converted to the same format as the original REST API responses, maintaining compatibility with existing tooling and workflows.

## Testing

- Maintains backward compatibility with existing tool interface
- Supports both labeled and unlabeled issue queries
- Handles cursor-based pagination correctly
- Preserves all issue metadata including labels, state, timestamps, and author information

Fixes #790

### DIFF
--- a/pkg/github/issues.go
+++ b/pkg/github/issues.go
@@ -18,6 +18,34 @@ import (
 	"github.com/shurcooL/githubv4"
 )
 
+var ListIssuesQuery struct {
+	Repository struct {
+		Issues struct {
+			Nodes []struct {
+				Number githubv4.Int
+				Title  githubv4.String
+				Body   githubv4.String
+				Author struct {
+					Login githubv4.String
+				}
+				CreatedAt githubv4.DateTime
+				Labels    struct {
+					Nodes []struct {
+						Name githubv4.String
+					}
+				} `graphql:"labels(first: 10)"`
+			}
+			PageInfo struct {
+				HasNextPage     githubv4.Boolean
+				HasPreviousPage githubv4.Boolean
+				StartCursor     githubv4.String
+				EndCursor       githubv4.String
+			}
+			TotalCount int
+		} `graphql:"issues(first: $first, after: $after, states: $states, orderBy: {field: $orderBy, direction: $direction})"`
+	} `graphql:"repository(owner: $owner, name: $repo)"`
+}
+
 // GetIssue creates a tool to get details of a specific issue in a GitHub repository.
 func GetIssue(getClient GetClientFn, t translations.TranslationHelperFunc) (tool mcp.Tool, handler server.ToolHandlerFunc) {
 	return mcp.NewTool("get_issue",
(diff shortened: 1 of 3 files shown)


pkg/github/issues.go
```diff
@@ -18,6 +18,126 @@ import (
...
+	PageInfo struct {
+		HasNextPage     githubv4.Boolean
+		HasPreviousPage githubv4.Boolean
+		StartCursor     githubv4.String
+		EndCursor       githubv4.String
+	}
+	TotalCount int
+}
+
+// ListIssuesQueryNoLabels is the query structure for fetching issues without label filtering.
```
tommaso-moro: nit but I think the comment meant to say `ListIssuesQuery is the query structure.. etc`?

Issue #728, opened 2025-07-22
MCP server not getting Permissions even they were given (for Organization)


sandy124356:
### Describe the bug

Classic PAT token created and Authorised with Org's SSO.
Given the token all permissions while creating.
Launched remote MCP server in VScode, allowed all MCP tools permissions in Vs code.

Now when we ask to view a specific repo's speciifc Branch's file/source code under the Org's Repos,
MCP still complains "I am unable to fetch the contents of <code>.py from it due to a reference resolution issue. This could be a temporary API problem or a repository configuration detail." 

It can able to list all repos in my Org.
can list branches in the repo.
But can never been able to access any of the content.

### Affected version

Please run ` docker run -i --rm ghcr.io/github/github-mcp-server ./github-mcp-server --version` and paste the output below

https://github.com/github/github-mcp-server/main

### Steps to reproduce the behavior

Try any prompt to access/summerise a code file under a Organisation repo.

my sample prompt:
summerize the code under <org> in the <repo>, within <branch> and <code_file> 

(or )

whats the code under this file doing ?

https://github.com/<ORG>/<REPO>/blob/<BRANCH>/<PATH>/<file>

### Expected vs actual behavior

Should have ability to read files and respond accordingly under Organization as all permissions were given.
But LLM's always complain some times saying 'branch doesnt exist, or repo doesnt exist or no MCP tool to read file contents, or 'a reference resolution issue' or ' I do not have the direct code contents from this repo'

### Logs

have seen traces of using
"failed to resolve git reference: failed to get reference: GET https://api.github.com/repos/<ORG>/<REPO>/git/ref/master: 404 Not Found []"

snemmaluri-tmx:
Also can you please help me understand which tool in this MCP server will help to read code content ? I believe its get_file_contents.  when we use github remote server , it comes with all tools by default (https://api.githubcopilot.com/mcp/) as per documentation.

When i ask to create a branch in the Organization repo, its getting branch created, but its never able to read a code file or summarize it.


more details that might be helpful

prompt: "use get_file_contents and read the file. the path where the file exists is [https://github.com/<org>/<repo>/blob/develop/code/<code.py>

Response:
Thinking:

input: 
{
  "owner": "<org>",
  "repo": "<repo>",
  "path": "code/code.py",
  "ref": "develop"
}

output:

failed to resolve git reference: failed to get reference: GET https://api.github.com/repos/<org>/<repo>/git/ref/develop: 404 Not Found []

[...]

PR #851, opened 2025-08-11
Tommy/improve-ref-handling

tommaso-moro:
Closes: [#728](https://github.com/github/github-mcp-server/issues/728)

### Overview
This PR significantly improves how we handle refs in the `get_file_contents` tool, after users [reported](https://github.com/github/github-mcp-server/issues/728) the tool being often broken from bad ref handling. 

### The problem
Despite the tool description, models can pass as argument to the tool a `ref` formatted either as fully qualified (e.g. `refs/heads/beanch_name`) or as short user-friendly names (e.g. `branch_name`). 

The [get_repository_content](https://docs.github.com/en/rest/repos/contents?apiVersion=2022-11-28#get-repository-content) endpoint from the Github API is able to handle both flexibly:
Screenshot 2025-08-11 at 11 47 24
Screenshot 2025-08-11 at 11 47 39

However, the `get_reference` endpoint doesn't (it requires a `ref` formatted in a specific way):
Screenshot 2025-08-11 at 11 51 42
Screenshot 2025-08-11 at 11 51 33

Our previous logic would fail often because it uses both endpoints but it was not handling user-friendly refs (e.g. `branch_name`, `heads/branch_name`, `tag_name`), which are often provided by models to the tool. This would thus break the tool.

### The Solution
This PR introduces new, robust logic to resolve a `ref` by identifying the format in which it was provided and constructing a fully qualified `ref` accordingly. In short, the new `resolveGitReference` function now handles:
- Fully-Qualified refs: `refs/heads/main`
- Partially-Qualified refs: `heads/main`
- Short Name refs: `main` (discovering automatically if it's a branch or a tag)

This ensures that the tool can handle the most common ways in which the `ref` input is provided by the user.

Additionally, I added:
- Comprehensive tests that cover all new logic paths.
- I have added quite detailed comments to the function itself to make the resolution logic, which can be tedious, much easier to follow

**Demo**
Screenshot 2025-08-11 at 11 43 38
Screenshot 2025-08-11 at 11 39 53
Screenshot 2025-08-07 at 17 55 20

### DIFF
--- a/pkg/github/repositories.go
+++ b/pkg/github/repositories.go
@@ -1407,7 +1407,6 @@ func resolveGitReference(ctx context.Context, githubClient *github.Client, owner
 	var resp *github.Response
 	var err error
 
-	// Only enter the resolution logic if the ref is NOT already fully qualified.
 	switch {
 	case strings.HasPrefix(ref, "refs/"):
 		// 2b) Already fully qualified. The reference will be fetched at the end.
--- a/pkg/github/repositories.go
+++ b/pkg/github/repositories.go
@@ -1390,6 +1390,8 @@ func resolveGitReference(ctx context.Context, githubClient *github.Client, owner
 		return &raw.ContentOpts{Ref: "", SHA: sha}, nil
 	}
 
+	originalRef := ref // Keep original ref for clearer error messages down the line.
+
 	// 2) If no SHA is provided, we try to resolve the ref into a fully-qualified format.
 	// 2a) If ref is empty, determine the default branch.
 	if ref == "" {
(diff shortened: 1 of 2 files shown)


pkg/github/repositories.go
```diff
@@ -1358,36 +1358,96 @@ func filterPaths(entries []*github.TreeEntry, path string, maxResults int) []str
...
 		if err != nil {
 			_, _ = ghErrors.NewGitHubAPIErrorToCtx(ctx, "failed to get repository info", resp, err)
 			return nil, fmt.Errorf("failed to get repository info: %w", err)
 		}
-		ref = fmt.Sprintf("refs/heads/%s", repoInfo.GetDefaultBranch())
+		ref = repoInfo.GetDefaultBranch()
 	}
 
-	// 3. Get the SHA from the ref
+	originalRef := ref // Keep original ref for clearer error messages down the line.
```
almaleksia: you assign `ref` to `originalRef` after you assigned it default branch (in case empty) which makes it not quite original. Besides, it means you're running ref resolution logic on the default branch which is not necessary because a) we know it is a branch and b) we know it exists. 

I propose we do this assign before line 1395 and use `originalRef` in checks on line 1407 and 1408.

pkg/github/repositories.go
```diff
@@ -1358,36 +1358,96 @@ func filterPaths(entries []*github.TreeEntry, path string, maxResults int) []str
...
+					// The branch lookup failed for a different reason.
+					_, _ = ghErrors.NewGitHubAPIErrorToCtx(ctx, "failed to get reference (branch)", resp, err)
+					return nil, fmt.Errorf("failed to get reference for branch '%s': %w", originalRef, err)
+				}
+			}
+		}
+	}
+
+	// Now that 'ref' is fully qualified, we get the definitive reference object.
 	reference, resp, err := githubClient.Git.GetRef(ctx, owner, repo, ref)
```
almaleksia: It might be that you already retrieved a reference on line 1430 or 1439. In this case this call is redundant. We probably would want to avoid that.

pkg/github/repositories.go
```diff
@@ -1358,36 +1358,96 @@ func filterPaths(entries []*github.TreeEntry, path string, maxResults int) []str
...
-	// 1. If SHA is provided, use it directly
+	// 1) If SHA explicitly provided, it's the highest priority.
 	if sha != "" {
 		return &raw.ContentOpts{Ref: "", SHA: sha}, nil
 	}
 
-	// 2. If neither provided, use the default branch as ref
+	// 2) If no SHA is provided, we try to resolve the ref into a fully-qualified format.
+	// 2a) If ref is empty, determine the default branch.
 	if ref == "" {
```
almaleksia: I'd make conditions in this method a bit easier to follow using switch, something like:

```go
switch {
	case originalRef == "":
		// 2a) If ref is empty, determine the default branch.
		repoInfo, resp, err := githubClient.Repositories.Get(ctx, owner, repo)
		if err != nil {
			_, _ = ghErrors.NewGitHubAPIErrorToCtx(ctx, "failed to get repository info", resp, err)
			return nil, fmt.Errorf("failed to get repository info: %w", err)
		}
		ref = repoInfo.GetDefaultBranch()
	
	case strings.HasPrefix(originalRef, "refs/"):
		// 2b) Already fully qualified, use as-is.
		ref = originalRef
	
	case strings.HasPrefix(originalRef, "heads/") || strings.HasPrefix(originalRef, "tags/"):
		// 2c) It's partially qualified. Make it fully qualified.
		ref = "refs/" + originalRef
	
	default:
        // 2d) It's a short name; try to resolve it as a branch or tag.
       ....
```

almaleksia (commented):
I really like this logic! 
Left some comments, mostly flow and readability related.

pkg/github/repositories.go
tommaso-moro: Not sure how I missed this! Ty 🙌

pkg/github/repositories.go
```diff
@@ -1358,36 +1358,102 @@ func filterPaths(entries []*github.TreeEntry, path string, maxResults int) []str
...
 		repoInfo, resp, err := githubClient.Repositories.Get(ctx, owner, repo)
 		if err != nil {
 			_, _ = ghErrors.NewGitHubAPIErrorToCtx(ctx, "failed to get repository info", resp, err)
 			return nil, fmt.Errorf("failed to get repository info: %w", err)
 		}
-		ref = fmt.Sprintf("refs/heads/%s", repoInfo.GetDefaultBranch())
+		ref = repoInfo.GetDefaultBranch()
```
almaleksia: you might want to keep removed logic here:
```
ref = fmt.Sprintf("refs/heads/%s", repoInfo.GetDefaultBranch())
```

 We do not need to run ref resolution logic in this case.

pkg/github/repositories.go
```diff
@@ -1358,36 +1358,102 @@ func filterPaths(entries []*github.TreeEntry, path string, maxResults int) []str
...
 	if sha != "" {
 		return &raw.ContentOpts{Ref: "", SHA: sha}, nil
 	}
 
-	// 2. If neither provided, use the default branch as ref
+	originalRef := ref // Keep original ref for clearer error messages down the line.
+
+	// 2) If no SHA is provided, we try to resolve the ref into a fully-qualified format.
+	// 2a) If ref is empty, determine the default branch.
 	if ref == "" {
```
almaleksia: this block can also be included as part of switch statement.

pkg/github/repositories.go
```diff
@@ -1358,36 +1358,102 @@ func filterPaths(entries []*github.TreeEntry, path string, maxResults int) []str
...
-	reference, resp, err := githubClient.Git.GetRef(ctx, owner, repo, ref)
-	if err != nil {
-		_, _ = ghErrors.NewGitHubAPIErrorToCtx(ctx, "failed to get reference", resp, err)
-		return nil, fmt.Errorf("failed to get reference: %w", err)
+	var reference *github.Reference
+	var resp *github.Response
+	var err error
+
+	switch {
+	case strings.HasPrefix(ref, "refs/"):
```
almaleksia: it might be beneficial and less error prone to use `originalRef` in this condition and on line 1413.

pkg/github/repositories.go
```diff
@@ -1358,36 +1358,100 @@ func filterPaths(entries []*github.TreeEntry, path string, maxResults int) []str
...
+	case originalRef == "":
+		// 2a) If ref is empty, determine the default branch.
 		repoInfo, resp, err := githubClient.Repositories.Get(ctx, owner, repo)
 		if err != nil {
 			_, _ = ghErrors.NewGitHubAPIErrorToCtx(ctx, "failed to get repository info", resp, err)
 			return nil, fmt.Errorf("failed to get repository info: %w", err)
 		}
 		ref = fmt.Sprintf("refs/heads/%s", repoInfo.GetDefaultBranch())
+	case strings.HasPrefix(originalRef, "refs/"):
+		// 2b) Already fully qualified. The reference will be fetched at the end.
```
almaleksia, line 1410: here you need to add:

```go
ref := originalRef
```

tommaso-moro: If we're in that case, `ref` is already the same as `originalRef` (as we have `originalRef := ref` at the top of the function) and because it is fully qualified we can exit the switch statement. Or am I missing something?

almaleksia: nvm! it's all good

PR #866, opened 2025-08-13
Add actions job log buffer and profiler

mattdholloway:
Extends the `get_job_logs` tool and other parts of the MCP server that rely on downloading Actions job log data. The new functionality ensures memory leaks do not occur by using a 'sliding window' buffer to only have 5MB of the log in memory at a time as it reads the entire log. This then allows for a tail step after this that retains the correct amount of lines within the final window before completion.

Closes: https://github.com/github/copilot-agent-services/issues/375


cmd/github-mcp-server/main.go
```diff
@@ -55,6 +55,7 @@ var (
 				ExportTranslations:   viper.GetBool("export-translations"),
 				EnableCommandLogging: viper.GetBool("enable-command-logging"),
 				LogFilePath:          viper.GetString("log-file"),
+				ContentWindowSize:    viper.GetInt("content_window_size"),
```
omgitsads:
```suggestion
				ContentWindowSize:    viper.GetInt("content-window-size"),
```

cmd/github-mcp-server/main.go
```diff
@@ -84,6 +86,7 @@ func init() {
 	_ = viper.BindPFlag("enable-command-logging", rootCmd.PersistentFlags().Lookup("enable-command-logging"))
 	_ = viper.BindPFlag("export-translations", rootCmd.PersistentFlags().Lookup("export-translations"))
 	_ = viper.BindPFlag("host", rootCmd.PersistentFlags().Lookup("gh-host"))
+	_ = viper.BindPFlag("content_window_size", rootCmd.PersistentFlags().Lookup("content-window-size"))
```
omgitsads:
```suggestion
	_ = viper.BindPFlag("content-window-size", rootCmd.PersistentFlags().Lookup("content-window-size"))
```

pkg/profiler/profiler.go
```diff
@@ -0,0 +1,215 @@
+package profiler
```
omgitsads: Maybe we should move this into `internal/` rather than `pkg/`.

pkg/buffer/buffer.go
```diff
@@ -0,0 +1,67 @@
...
+// The function uses a ring buffer to efficiently store only the last maxJobLogLines lines.
+// If the response contains more lines than maxJobLogLines, only the most recent lines are kept.
+func ProcessResponseAsRingBufferToEnd(httpResp *http.Response, maxJobLogLines int) (string, int, *http.Response, error) {
+	lines := make([]string, maxJobLogLines)
+	validLines := make([]bool, maxJobLogLines)
+	totalLines := 0
+	writeIndex := 0
+
+	scanner := bufio.NewScanner(httpResp.Body)
+	scanner.Buffer(make([]byte, 0, 64*1024), 1024*1024)
```
omgitsads, line 35: Let's reduce this per line size to something more reasonable.

omgitsads (approved):

PR #873, opened 2025-08-13
fix: Skip CodeQL scanning on forked repositories

kaovilai:
## Summary
Adds a condition to the CodeQL workflow to only run on the main repository, preventing failures on forks.

## Problem
The CodeQL code scanning workflow fails on forked repositories because:
- Forks don't have `security-events: write` permission to upload results
- The workflow uses internal GitHub secrets (`GITHUB_REGISTRIES_PROXY`) not available in forks
- It references internal GitHub CodeQL packs that may not be accessible from forks

## Solution
Added an `if` condition to the analyze job to only run when `github.repository == 'github/github-mcp-server'`

This ensures the workflow:
- ✅ Runs on the main repository as intended
- ✅ Skips execution on forks, preventing unnecessary failures
- ✅ Still allows forks to run other workflows that don't require special permissions


SamMorrowDrums (commented):
Thank you for your contribution, I think it's reasonable.

Issue #684, opened 2025-07-16
Feature req - repository security advisories


leonard520:
### Describe the feature or problem you’d like to solve

Is there a plan to support query repository security advisories?

### Proposed solution

As a user, I have a need to access repository security advisories through the MCP server.

### Example prompts or workflows (for tools/toolsets only)

Please list repository security advisories
I would like to query a CVE <ghsa_id>

Since the agent was not trained on the latest codebase, its awareness of known vulnerabilities is limited. By integrating the GitHub Advisory Database, the agent can perform real-time CVE lookups, fix vulnerable dependencies, and provide improved security and compliance.

### Additional context

N/A

leonard520:
Kindly ping on the feature request

D1M1TR10S:
Hey @leonard520. We shipped this toolset with [this release](https://github.com/github/github-mcp-server/releases/tag/v0.13.0). The tools are called: list_repository_security_advisories, list_org_repository_security_advisories , list_global_security_advisories and get_global_security_advisory.

PR #919, opened 2025-08-19
Add Global Security Advisories Toolset

jurre:
This implements the Global Security Advisories tool, pulling from [these endpoints](https://docs.github.com/en/rest/security-advisories/global-advisories?apiVersion=2022-11-28), this allows users to query for vulnerabilities in their dependencies and figure out patched versions.

I'll follow this up with repo-level advisories as well, although I think for most use-cases users would probably want to use the existing Dependabot Alerts tool.

image

image

Addresses part of: https://github.com/github/github-mcp-server/issues/684

### DIFF
--- a/README.md
+++ b/README.md
@@ -244,7 +244,7 @@ For other MCP host applications, please refer to our installation guides:
 - **[Cursor](docs/installation-guides/install-cursor.md)** - Installation guide for Cursor IDE
 - **[Windsurf](docs/installation-guides/install-windsurf.md)** - Installation guide for Windsurf IDE
 
-For a complete overview of all installation options, see our **[Installation Guides Index](docs/installation-guides/installation-guides.md)**.
+For a complete overview of all installation options, see our **[Installation Guides Index](docs/installation-guides)**.
 
 > **Note:** Any host application that supports local MCP servers should be able to access the local GitHub MCP server. However, the specific configuration process, syntax and stability of the integration will vary by host application. While many may follow a similar format to the examples above, this is not guaranteed. Please refer to your host application's documentation for the correct MCP configuration syntax and setup process.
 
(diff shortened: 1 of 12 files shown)


pkg/github/security_advisories.go
```diff
@@ -0,0 +1,229 @@
...
+			mcp.WithDescription(t("TOOL_LIST_GLOBAL_SECURITY_ADVISORIES_DESCRIPTION", "List global security advisories from GitHub.")),
+			mcp.WithToolAnnotation(mcp.ToolAnnotation{
+				Title:        t("TOOL_LIST_GLOBAL_SECURITY_ADVISORIES_USER_TITLE", "List global security advisories"),
+				ReadOnlyHint: ToBoolPtr(true),
+			}),
+			mcp.WithString("ghsaId",
+				mcp.Description("Filter by GitHub Security Advisory ID (format: GHSA-xxxx-xxxx-xxxx)."),
+			),
+			mcp.WithString("type",
+				mcp.Required(),
```
tommaso-moro: Should this be required? It seems like it's optional in the endpoint [docs](https://docs.github.com/en/rest/security-advisories/global-advisories?apiVersion=2022-11-28)

tommaso-moro (changes requested):
Lgtm but I think it would be good to make the `type` param optional in the `ListGlobalSecurityAdvisories` tool, as per the [endpoint docs](https://docs.github.com/en/rest/security-advisories/global-advisories?apiVersion=2022-11-28)

pkg/github/security_advisories.go
jurre: Yeah you are right!

jurre: Updated 👍

PR #929, opened 2025-08-20
Fix installation guides path

GuyPago:
The existing link directs to a non existing `.md` file.


rusly2002ms-cmd:
# A first-level heading
## A second-level heading
### A third-level heading

Issue #887, opened 2025-08-14
Use a valid role for prompts: assistant instead of system


victorRadu:
### Describe the bug

Prompts return role: system. This is not valid according to the current MCP spec:
https://modelcontextprotocol.io/specification/2025-06-18/server/prompts#promptmessage.
This causes for example the MCP inspector to throw an error

### Affected version

GitHub MCP Server
Version: v0.11.0
Commit: 8aa3379feacb388289379d27b27692e91dd3c2ed
Build Date: 2025-08-08T14:34:15Z

#### Path 1:
1. Check the specs https://modelcontextprotocol.io/specification/2025-06-18/server/prompts#promptmessage.
2. Check the response of the Prompts, they claim: role: system

#### Path 2:
1. Launch the MCPInspector (https://github.com/modelcontextprotocol/inspector)
2. List Prompts
3. Try to get any prompt
4. Error response: [ { "received": "system", "code": "invalid_enum_value", "options": [ "user", "assistant" ], "path": [ "messages", 0, "role" ], "message": "Invalid enum value. Expected 'user' | 'assistant', received 'system'" } ]

### Expected vs actual behavior

The expectation would be to have "role": "assistant"

jkos-studio:
Go

PR #941, opened 2025-08-21
Fix role from "system" to "user" in message prompt

tommaso-moro:
Closes: #887 

The `system` prompt is not valid according to the MCP [specs](https://modelcontextprotocol.io/specification/2025-06-18/server/prompts#promptmessage.), so we use `user` instead


jkos-studio:
Go

jkos-studio:
Go

JaclynCodes (commented):
Ok

Issue #949, opened 2025-08-22
Add ability to specify reason why issue is being closed, when closing issue


tommaso-moro:
The `update_issue` tool allows to mark an issue as closed, but doesn't allow for specifying a reason (e.g. closed as "Not Planned")

C-Ross:
➕ I would find this useful, as one of the use cases I find the Copilot and the GitHub MCP particularly good at is consolidating and obsoleting issues, as well as obsoleting classes of issues.

PR #1073, opened 2025-09-10
Add specifying state change reason to `update_issue` tool

kerobbi:
This PR adds support for a `state_reason` parameter to the `update_issue` tool, enabling users to specify the reason when changing an issue's state (particularly when closing issues).

Valid values are: `completed`, `not_planned`, `duplicate`

When closing an issue as duplicate, the current implementation mandates the presence of `duplicate_of` (new parameter), specifying which issue the target issue is a duplicate of, replicating the user experience for duplicate management in the UI.

## Status

- The [REST API](https://docs.github.com/en/enterprise-cloud@latest/rest/issues/issues?apiVersion=2022-11-28#update-an-issue) allows closing an issue as duplicate, but does not give the ability to specify the issue the target issue is a duplicate of.
- The GraphQL API [offers the above functionality](https://docs.github.com/en/graphql/reference/input-objects#closeissueinput); however, the underlying `shurcooL/githubv4` library does not fully align with what the API offers.

## Chosen approach

- REST API for base functionality, GraphQL API to close as duplicate.
- **Limitation:** The underlying `githubv4` library does not fully align with what the GraphQL API offers, custom structs need to be created to get this working.

## Alternatives considered

<details><summary><b>REST API + automatic closing comment</b></summary>

- Automatically post a comment (i.e., "Duplicate of X") before closing the issue.
 
- Discarded because this introduces complexity for a feature that could be manually replicated using the `add_issue_comment` tool.
</details>

<details>
<summary><b>Contribute to <code>shurcooL/githubv4</code></b></summary>

- It is possible to open a PR in `shurcooL/githubv4` and make the necessary changes for this work (i.e., update `CloseIssue` and `CloseIssueInput` structs).
- Although a better approach, I've discarded this mainly due to the repo activity - it honestly seems like contributions have not been considered by the maintainer in a while

</details>

## Future improvements
- If and when up-to-date support for closing issues as duplicates is added in `shurcooL/githubv4`, we could remove the custom structs created for this PR.

Closes: #949

### DIFF
--- a/pkg/github/issues.go
+++ b/pkg/github/issues.go
@@ -88,17 +88,14 @@ func fetchIssueIDs(ctx context.Context, gqlClient *githubv4.Client, owner, repo 
 }
 
 // getCloseStateReason converts a string state reason to the appropriate enum value
-func getCloseStateReason(stateReason string) *IssueClosedStateReason {
+func getCloseStateReason(stateReason string) IssueClosedStateReason {
 	switch stateReason {
 	case "not_planned":
-		reason := IssueClosedStateReasonNotPlanned
-		return &reason
+		return IssueClosedStateReasonNotPlanned
 	case "duplicate":
-		reason := IssueClosedStateReasonDuplicate
-		return &reason
+		return IssueClosedStateReasonDuplicate
 	default: // Default to "completed" for empty or "completed" values
-		reason := IssueClosedStateReasonCompleted
-		return &reason
+		return IssueClosedStateReasonCompleted
 	}
 }
 
@@ -1259,7 +1256,6 @@ func UpdateIssue(getClient GetClientFn, getGQLClient GetGQLClientFn, t translati
 
 			// Create the issue request with only provided fields
 			issueRequest := &github.IssueRequest{}
-			hasNonStateUpdates := false
 
 			// Set optional parameters if provided
 			title, err := OptionalParam[string](request, "title")
@@ -1268,7 +1264,6 @@ func UpdateIssue(getClient GetClientFn, getGQLClient GetGQLClientFn, t translati
 			}
 			if title != "" {
 				issueRequest.Title = github.Ptr(title)
-				hasNonStateUpdates = true
 			}
 
 			body, err := OptionalParam[string](request, "body")
@@ -1277,7 +1272,6 @@ func UpdateIssue(getClient GetClientFn, getGQLClient GetGQLClientFn, t translati
 			}
 			if body != "" {
 				issueRequest.Body = github.Ptr(body)
-				hasNonStateUpdates = true
 			}
 
 			// Get labels
@@ -1287,7 +1281,6 @@ func UpdateIssue(getClient GetClientFn, getGQLClient GetGQLClientFn, t translati
 			}
 			if len(labels) > 0 {
 				issueRequest.Labels = &labels
-				hasNonStateUpdates = true
 			}
 
 			// Get assignees
@@ -1297,7 +1290,6 @@ func UpdateIssue(getClient GetClientFn, getGQLClient GetGQLClientFn, t translati
 			}
 			if len(assignees) > 0 {
 				issueRequest.Assignees = &assignees
-				hasNonStateUpdates = true
 			}
 
 			milestone, err := OptionalIntParam(request, "milestone")
@@ -1307,7 +1299,6 @@ func UpdateIssue(getClient GetClientFn, getGQLClient GetGQLClientFn, t translati
 			if milestone != 0 {
 				milestoneNum := milestone
 				issueRequest.Milestone = &milestoneNum
-				hasNonStateUpdates = true
 			}
 
 			// Get issue type
@@ -1317,7 +1308,6 @@ func UpdateIssue(getClient GetClientFn, getGQLClient GetGQLClientFn, t translati
 			}
 			if issueType != "" {
 				issueRequest.Type = github.Ptr(issueType)
-				hasNonStateUpdates = true
 			}
 
 			// Handle state, state_reason and duplicateOf parameters
@@ -1330,9 +1320,6 @@ func UpdateIssue(getClient GetClientFn, getGQLClient GetGQLClientFn, t translati
 			if err != nil {
 				return mcp.NewToolResultError(err.Error()), nil
 			}
-			if stateReason != "" && state == "" {
-				return mcp.NewToolResultError("state_reason can only be used when state is also provided"), nil
-			}
 
 			duplicateOf, err := OptionalIntParam(request, "duplicate_of")
 			if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/github/issues.go
```diff
@@ -1183,14 +1212,45 @@ func UpdateIssue(getClient GetClientFn, t translations.TranslationHelperFunc) (t
...
+			}
+
+			// Validate state_reason usage
+			if stateReason != "" && state == "" {
+				return mcp.NewToolResultError("state_reason can only be used when state is also provided"), nil
+			}
+			if state == "open" && stateReason != "" && stateReason != "reopened" {
+				return mcp.NewToolResultError("when state is 'open', state_reason can only be 'reopened'"), nil
+			}
+			if state == "closed" && stateReason != "" && stateReason != "completed" && stateReason != "not_planned" && stateReason != "duplicate" {
```
tonytrg: This seems unnecessarily complicated.

Does this check the specific state closed & statereason reopened.

tonytrg (commented):
Sorry couldn't do a full review today.

Thanks for tackling this complicate case 💯 , it is probably owed due to limitations of apis, but we are introducing a hefty amount of conditional branches. Can this maybe be simplified?

From what I am understanding from the pr description only state_reason `duplicate` require the graphql api. Can the graphql client relevant code be more isolated from the rest client code.

pkg/github/issues.go
LuluBeatson: Does the REST/GQL API do any of this validation on their side and provide a helpful error message? If so, we could skip it in the tool.

pkg/github/issues.go
```diff
@@ -1183,14 +1212,45 @@ func UpdateIssue(getClient GetClientFn, t translations.TranslationHelperFunc) (t
...
+				return mcp.NewToolResultError(err.Error()), nil
+			}
+
+			// Validate state_reason usage
+			if stateReason != "" && state == "" {
+				return mcp.NewToolResultError("state_reason can only be used when state is also provided"), nil
+			}
+			if state == "open" && stateReason != "" && stateReason != "reopened" {
+				return mcp.NewToolResultError("when state is 'open', state_reason can only be 'reopened'"), nil
+			}
```
LuluBeatson: Here if state:open means stateReason has to be `reopened`, could we just set it as such and continue? instead of responding with error

pkg/github/issues.go
```diff
@@ -1227,34 +1290,117 @@ func UpdateIssue(getClient GetClientFn, t translations.TranslationHelperFunc) (t
...
+					}
+					return mcp.NewToolResultError(fmt.Sprintf("failed to update issue: %s", string(body))), nil
+				}
 			}
-			defer func() { _ = resp.Body.Close() }()
 
-			if resp.StatusCode != http.StatusOK {
-				body, err := io.ReadAll(resp.Body)
+			// Handle GraphQL API updates (state_reason = "duplicate")
+			if gqlUpdateNeeded {
```
LuluBeatson: You could consider extracting into 2 separate functions for the REST and GraphQL updaters, which could make unit testing a bit easier.

It also seems like the GraphQL capabilities are a superset of the REST API, so could we use only GraphQL?

pkg/github/issues.go
kerobbi: Main purpose of this was just to ensure that the `state` - `state_reason` values were used properly. For instance, if `state` is open, the only acceptable value for `state_reason` is `reopened`. Addressed it in the latest changes though as I moved the entire state / state reason handling to use the GraphQL API.

pkg/github/issues.go
kerobbi: That makes perfect sense! It did become redundant though as I've moved the entire `state` / `state_reason` handling to make use of the gql api rather than the rest api as suggested by @tonytrg. The only acceptable values for `state_reason` now are `completed`, `not_planned` and `duplicate`.

pkg/github/issues.go
kerobbi: Looked better into this and we could indeed use only GQL (i.e., [`updateIssue`](https://docs.github.com/en/graphql/reference/mutations#updateissue) / [`updateIssueInput`](https://docs.github.com/en/graphql/reference/input-objects#updateissueinput)). However, I believe the current hybrid approach is actually more optimal, mainly due to the amount of ID resolution overhead that would be needed if we go full GQL. I did refactor everything for better readability/separation of concerns, but more than happy to review this further if needed.

pkg/github/issues.go
```diff
@@ -1168,6 +1259,7 @@ func UpdateIssue(getClient GetClientFn, t translations.TranslationHelperFunc) (t
 
 			// Create the issue request with only provided fields
 			issueRequest := &github.IssueRequest{}
+			hasNonStateUpdates := false
```
tonytrg: Does a check at the end not work more cleanly:

`if *issueRequest == github.IssueRequest{} {
    fmt.Println("issueRequest is empty")
}`

To avoid these multiple hasNonStateUpdates checks.

tonytrg: But I would not test that, I would simply forward the payload and execute the request.
Any extra logic to check if there are really changes could potentially lead to unwanted behaviour.

If the tool call has no params set, that would change the issue, the api call is equal to a noop.

We simply forward the users request and return the apis output. No extra logic.
The LLM should get the same payload as if executing itself via curl.

pkg/github/issues.go
```diff
@@ -1228,30 +1317,151 @@ func UpdateIssue(getClient GetClientFn, t translations.TranslationHelperFunc) (t
...
+
+				// Get target issue ID (and duplicate issue ID if needed)
+				issueID, duplicateIssueID, err := fetchIssueIDs(ctx, gqlClient, owner, repo, issueNumber, duplicateOf)
+				if err != nil {
+					return ghErrors.NewGitHubGraphQLErrorResponse(ctx, "Failed to find issues", err), nil
+				}
+
+				// Do all logic
+				switch state {
+				case "open":
```
tonytrg, line 1375: That is much simpler 👍

pkg/github/issues.go
```diff
@@ -18,6 +18,90 @@ import (
...
+	}
+
+	return query.Repository.Issue.ID, query.Repository.DuplicateIssue.ID, nil
+}
+
+// getCloseStateReason converts a string state reason to the appropriate enum value
+func getCloseStateReason(stateReason string) *IssueClosedStateReason {
+	switch stateReason {
+	case "not_planned":
+		reason := IssueClosedStateReasonNotPlanned
```
tonytrg:
```suggestion
		return IssueClosedStateReasonNotPlanned
```

tonytrg: nit oneliner

pkg/github/issues.go
```diff
@@ -1228,30 +1317,151 @@ func UpdateIssue(getClient GetClientFn, t translations.TranslationHelperFunc) (t
...
+					err = gqlClient.Mutate(ctx, &mutation, closeInput, nil)
+					if err != nil {
+						return ghErrors.NewGitHubGraphQLErrorResponse(ctx, "Failed to close issue", err), nil
+					}
 				}
-				return mcp.NewToolResultError(fmt.Sprintf("failed to update issue: %s", string(body))), nil
 			}
 
+			// Get the final state of the issue to return
+			client, err := getClient(ctx)
```
tonytrg: we have the first api call with the information.

that third call seems unnecessary, we simply return the minimal response, and for that we have the id and url.

PR #1085, opened 2025-09-13
docs: Change `trust` for Gemini CLI configuration from true to false

vorburger:
See https://github.com/google-gemini/gemini-cli/issues/5512.


vorburger:
> See https://github.com/google-gemini/gemini-cli/issues/5512.

TL;DR According to https://github.com/google-gemini/gemini-cli/blob/main/docs/tools/mcp-server.md#optional:

> `trust` (boolean): When true, bypasses all tool call confirmations for this server (default: false)

It's probably not a good idea to recommend this as the default for new users on this initial set-up example documentation.

LuluBeatson:
Thank you very much for this @vorburger. I've included this change in https://github.com/github/github-mcp-server/pull/1223 so I have now closed this PR.

PR #1067, opened 2025-09-09
Fix environment variable mapping for read-only mode

pinotelio:
Fix environment variable mapping for read-only mode configuration

Add SetEnvKeyReplacer to viper configuration to properly map environment variables with underscores to flag names with dashes. This enables the documented GITHUB_READ_ONLY=1 environment variable to work correctly.

Without this fix, viper was looking for GITHUB_READ-ONLY (with dash) but the documentation and standard convention use GITHUB_READ_ONLY (with underscore).

Fixes issue where read-only mode was not being activated when using GITHUB_READ_ONLY=1 in Docker containers.

**Incorrect Read-Only Tools when `GITHUB_READ_ONLY=1` passed:**
Incorrect Read-Only Tools

**Test with `GITHUB_READ-ONLY=1` (Dash):**
Test With Dash

**Correct Result After The Patch:**
Patched

| Requirement | Command | Status |
|-------------|---------|--------|
| ✅ Tests pass | `go test -v ./...` | **PASSED** |
| ✅ Linter passes | `golangci-lint run` | **PASSED** |
| ✅ Project linter | `script/lint` | **PASSED** |
| ✅ Update snapshots | `UPDATE_TOOLSNAPS=true go test ./...` | **COMPLETED** |
| ✅ Update docs | `script/generate-docs` | **COMPLETED** |

<br class="Apple-interchange-newline">

Closes:

PR #1102, opened 2025-09-19
Update README.md

cheftako:
The viper binding at https://github.com/github/github-mcp-server/blob/2a0c7bff82ea8d8f1d4999b3f222c8e8c13b05cc/cmd/github-mcp-server/main.go#L84 means the environment flag is GITHUB_READ-ONLY not GITHUB_READ_ONLY. To be backward compatible I think we need to honor the flag as it works. So probably best to just fix the docs here.


JoannaaKL:
Hey @cheftako 👋  Viper configuration was updated yesterday and `GITHUB_READ_ONLY` will be honored [link](https://github.com/github/github-mcp-server/pull/1067). In case of any questions please open an issue.

Issue #1108, opened 2025-09-21
Add toolset configuration support for remote GitHub MCP server


nick-youngblut:
### Describe the feature or problem you'd like to solve

The remote GitHub MCP server at `https://api.githubcopilot.com/mcp/` currently exposes many dozen tools by default with no documented way to limit or configure which toolsets are available. While the Docker-based server supports toolset configuration via the `GITHUB_TOOLSETS` environment variable, the remote server lacks equivalent functionality.

<img width="537" height="866" alt="Image" src="https://github.com/user-attachments/assets/bd2ec67c-5a55-4edf-8bdb-71cf562e3efd" />

This creates several problems:
- AI agents can become overwhelmed by too many tool options, leading to poor tool selection
- Some MCP hosts (like Cursor) only send the first 40 tools to agents, meaning 60+ tools are inaccessible
- Users who only need specific GitHub functionality (e.g., read-only code review, issue management only) must expose unnecessary and potentially risky tools
- Organizations wanting to limit scope for security/compliance have no granular control beyond the all-or-nothing `X-MCP-Readonly` header

### Proposed solution

Add support for toolset configuration in the remote server via headers or URL parameters, similar to how the Docker version uses environment variables. This could be implemented as:

**Option 1: Header-based configuration**
```json
{
  "headers": {
    "Authorization": "Bearer TOKEN",
    "X-MCP-Toolsets": "context,repos,issues,pull_requests"
  }
}
```

**Option 2: URL parameter**
```json
{
  "url": "https://api.githubcopilot.com/mcp/?toolsets=context,repos,issues"
}
```

**Option 3: Dynamic toolset discovery** (already supported in Docker)
```json
{
  "headers": {
    "X-MCP-Dynamic-Toolsets": "true"
  }
}
```

This would benefit users by:
- Improving AI agent accuracy through reduced tool confusion
- Ensuring all desired tools fit within host limitations (40-tool limit)
- Providing security-conscious organizations with granular access control
- Maintaining feature parity between Docker and remote server implementations

### Example prompts or workflows (for tools/toolsets only)

1. **Code Review Workflow** - Enable only: `context,repos,pull_requests`
   - "Review the latest PR and check if it follows our coding standards"
   - "Show me all open PRs that modify authentication logic"
   - "What changes were made in PR #123?"

2. **Issue Triage Workflow** - Enable only: `context,issues`
   - "List all high-priority bugs from this week"
   - "Create an issue for the bug we just discussed"
[...]

PR #1111, opened 2025-09-22
docs: Update optional headers and path params for remote server

LuluBeatson:
Closes: https://github.com/github/github-mcp-server/issues/1108

Adds description of optional headers and URL path params that can be used to restrict the Remote GitHub MCP server's toolsets and read-only mode.


tommaso-moro (approved):
lgtm!

PR #1143, opened 2025-09-26
Add container label & improve registry script

MattBabbage:
This pull request updates the release workflow and configuration to improve version management, automate validation, and align metadata with MCP requirements. The most important changes are grouped below.

All PRs will be squashed into one, _lots of trial and error here_.


.github/workflows/registry-releaser.yml
```diff
@@ -16,33 +16,37 @@ jobs:
...
+        with:
+          go-version: "stable"
+        
       - name: Fetch tags
         run: git fetch --tags
 
       - name: Install MCP Publisher
         run: |
-          curl -L "https://github.com/modelcontextprotocol/registry/releases/download/v1.0.0/mcp-publisher_1.0.0_$(uname -s | tr '[:upper:]' '[:lower:]')_$(uname -m | sed 's/x86_64/amd64/;s/aarch64/arm64/').tar.gz" | tar xz mcp-publisher
+          git clone --quiet https://github.com/modelcontextprotocol/registry publisher-repo
```
JoannaaKL, line 29: Why did you decide to clone and build it instead of pulling the package? :)

MattBabbage: Had issue where v1.0.0 or the curl build would cause a known bug in workflows. camelCase would get converted to snake_case then get rejected by API. Looked at other MCP servers and this was their method, which actually seemed fixed the problem

server.json
```diff
@@ -62,11 +61,5 @@
         }
       ]
     }
-  ],
-  "remotes": [
```
almaleksia, line 66: can you explain why you deleted this?

.github/workflows/registry-releaser.yml
```diff
@@ -16,33 +16,37 @@ jobs:
...
-            echo "Using triggered tag: ${{ github.ref_name }}"
           else
-            # Fallback to latest tag (for manual triggers)
             LATEST_TAG=$(git tag --sort=-version:refname | grep -E '^v[0-9]+\.[0-9]+\.[0-9]+(-.*)?$' | head -n 1)
-            if [ -z "$LATEST_TAG" ]; then
-              echo "❌ No release tag found. Cannot determine version."
-              exit 1
-            fi
+            [ -z "$LATEST_TAG" ] && { echo "No release tag found"; exit 1; }
             TAG_VERSION=$(echo "$LATEST_TAG" | sed 's/^v//')
```
almaleksia, line 40: wouldn't this be useful to keep?

.github/workflows/registry-releaser.yml
```diff
@@ -16,33 +16,37 @@ jobs:
...
-            echo "Using latest tag: $LATEST_TAG"
```
almaleksia, line 40: wouldn't this be useful to keep?

server.json
```diff
@@ -12,9 +12,8 @@
     {
       "registryType": "oci",
       "registryBaseUrl": "https://ghcr.io",
-      "identifier": "ghcr.io/github/github-mcp-server",
```
almaleksia, line 15: is there a reason you removed this prefix?

server.json
```diff
@@ -12,9 +12,8 @@
...
+      "identifier": "github/github-mcp-server",
       "version": "${VERSION}",
-      "runtimeHint": "docker",
```
almaleksia, line 17: is it not needed?

server.json
MattBabbage: When publishing to the OSS MCP Registry OIDC is used for auth. Which has the token for our publisher domain embedded in the JWT, this only allows us to publish to github.github.io.

Publishing remote at current url results in:
{"title":"Bad Request","status":400,"detail":"Failed to publish server","errors":[{"message":"remote URL https://api.githubcopilot.com/mcp/ does not match namespace io.github.github/github-mcp-server: remote URL host api.githubcopilot.com does not match publisher domain github.github.io"}]}

I will investigate adding permissions for remote repository, and probably publish it separately as they are not version synced. _i.e. remote is always the newest version so publishing V0.14.0 as remote with that url, would result in a misleading entry to their db_

server.json
MattBabbage: Their api seems to construct a composite from registryBaseUrl + identifier

server.json
MattBabbage: Nope, not in current json spec and doesnt seem used

.github/workflows/registry-releaser.yml
MattBabbage: Added back

almaleksia (approved):

PR #1164, opened 2025-09-30
Add Repository Tree Navigation Tool

natagdunbar:
Closes: https://github.com/github/copilot-agent-services/issues/651

This PR adds a new `get_repository_tree` tool to the GitHub MCP Server, enabling natural language exploration of repository file structures.

### What's Added
**New Tool**: `get_repository_tree` - Retrieves repository file/directory structure using GitHub's Git Tree API
**Response Types**: TreeResponse and TreeEntryResponse structs for structured data
**Tool Registration**: Added to the "repos" toolset alongside existing repository tools
**Recursive Exploration**: Optional recursive directory traversal
**Path Filtering**: Filter results by directory prefix (e.g., src/, [pkg](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html))
**Branch/SHA Support**: Query specific branches, tags, or commit SHAs
**Rich Metadata**: Returns file types, modes, SHAs, and URLs

### Impact
This addresses the navigation capability gap between VS Code agent mode (which has CLI access for ls) and dotcom chat, enabling AI assistants to naturally explore and understand repository structures through conversational interfaces.

I was able to use this tool in VS code: 
Screenshot 2025-09-30 at 4 33 14 PM
Screenshot 2025-09-30 at 4 33 29 PM
Screenshot 2025-09-30 at 4 33 36 PM
Screenshot 2025-09-30 at 4 34 36 PM


pkg/github/repositories.go
```diff
@@ -677,6 +677,147 @@ func GetFileContents(getClient GetClientFn, getRawClient raw.GetRawClientFn, t t
...
+			if treeSHA == "" {
+				repoInfo, _, err := client.Repositories.Get(ctx, owner, repo)
+				if err != nil {
+					return mcp.NewToolResultError(fmt.Sprintf("failed to get repository info: %s", err)), nil
+				}
+				treeSHA = *repoInfo.DefaultBranch
+			}
+
+			// Get the tree using the GitHub Git Tree API
+			tree, resp, err := client.Git.GetTree(ctx, owner, repo, treeSHA, recursive)
```
iancanderson: Does the client handle `recursive: false` such that it _won't_ return recursive entries? The [docs](https://docs.github.com/en/rest/git/trees?apiVersion=2022-11-28#get-a-tree) suggest that we'll need to omit the `recursive` parameter completely in order to prevent recursion:

> Setting this parameter to any value returns the objects or subtrees referenced by the tree specified in :tree_sha. For example, setting recursive to any of the following will enable returning objects or subtrees: 0, 1, "true", and "false". Omit this parameter to prevent recursively returning objects or subtrees.

iancanderson: Ah, yep, it looks like the client will only include the recursive parameter if it's `true`: https://github.com/google/go-github/blob/46f1bf23e6f9659d04f9eaebff5d25902cddfd8e/github/git_trees.go#L102-L104

👍

pkg/github/repositories.go
```diff
@@ -677,6 +677,147 @@ func GetFileContents(getClient GetClientFn, getRawClient raw.GetRawClientFn, t t
...
+					err,
+				), nil
+			}
+			defer func() { _ = resp.Body.Close() }()
+
+			// Filter tree entries if path_filter is provided
+			var filteredEntries []*github.TreeEntry
+			if pathFilter != "" {
+				for _, entry := range tree.Entries {
+					if entry.Path != nil && strings.HasPrefix(*entry.Path, pathFilter) {
```
iancanderson: I wonder, will models try to use filter values like `src/*`?
I think it's something we can keep an eye out for. If we see it, it could be worth adding some negative prompting in the tool description.

pkg/github/repositories.go
```diff
@@ -677,6 +677,147 @@ func GetFileContents(getClient GetClientFn, getRawClient raw.GetRawClientFn, t t
...
+				treeEntries[i] = TreeEntryResponse{
+					Path: *entry.Path,
+					Type: *entry.Type,
+					Mode: *entry.Mode,
+					SHA:  *entry.SHA,
+					URL:  *entry.URL,
+				}
+				if entry.Size != nil {
+					treeEntries[i].Size = entry.Size
+				}
```
iancanderson: The [response schema](https://docs.github.com/en/rest/git/trees?apiVersion=2022-11-28#get-a-tree) for the api says that path, mode, type, and sha are required, but not `URL` or `Size`, so would be good to check for non-nil `URL` as well.

```json
    "tree": {
      "description": "Objects specifying a tree structure",
      "type": "array",
      "items": {
        "type": "object",
        "required": [
          "path",
          "mode",
          "type",
          "sha"
        ],
```

pkg/github/repositories.go
```diff
@@ -677,6 +677,146 @@ func GetFileContents(getClient GetClientFn, getRawClient raw.GetRawClientFn, t t
...
+				Recursive: recursive,
+				Count:     len(filteredEntries),
+			}
+
+			r, err := json.Marshal(response)
+			if err != nil {
+				return nil, fmt.Errorf("failed to marshal response: %w", err)
+			}
+
+			return mcp.NewToolResultText(string(r)), nil
```
omgitsads: We have support for Output Schemas now, so you could move `TreeResponse` and `TreeResponseEntry` struct type definitions out of this func, and make it a structured output for the tool, rather than passing back escaped JSON e.g. https://github.com/mark3labs/mcp-go/blob/main/examples/structured_input_and_output/README.md.

pkg/github/git.go
```diff
@@ -0,0 +1,156 @@
...
+			client, err := getClient(ctx)
+			if err != nil {
+				return mcp.NewToolResultError("failed to get GitHub client"), nil
+			}
+
+			// If no tree_sha is provided, use the repository's default branch
+			if treeSHA == "" {
+				repoInfo, _, err := client.Repositories.Get(ctx, owner, repo)
+				if err != nil {
+					return mcp.NewToolResultError(fmt.Sprintf("failed to get repository info: %s", err)), nil
```
omgitsads:
```suggestion
					return ghErrors.NewGitHubAPIErrorResponse(ctx,
					  "failed to get repository info",
					  resp,
					  err,
					), nil
```

omgitsads (dismissed):

PR #1180, opened 2025-10-06
Add note for GitHub Enterprise Cloud Data Residency configuration

mortenp1337:
Clarifies that users of GitHub Enterprise Cloud with Data Residency need to update the API URL in their JSON configuration, providing a concrete example for clarity consistent with examples in other documentation: https://docs.github.com/en/enterprise-cloud@latest/admin/data-residency/about-github-enterprise-cloud-with-data-residency#api-access


mortenp1337:
Fixed in #1210

Issue #44, opened 2025-03-27
Add support for Projects


toby:
From @ladynaggaga:

Add support for get_project_board or analyze_project_board tool where devs , leads, PMs can look at the status of work item in copilot.

vongohren:
Just to connect the two tasks: https://github.com/github/github-mcp-server/issues/117

moonray:
I have a basic implementation working locally based on the original modelcontextprotocol repo. I'll try to work on getting it ported to this new repo.

deweyjose:
I'd like the ability to add issues to a project. Right now when I create issues via the MCP server project is not specified. Is this a completely different request? Happy to create a different issue.

hillaryke:
At Arclio, we're excited to share Github Projects MCP Server to manage projects. We also needed this functionality for our own projects, so we created this MCP Server which is open-source:
https://github.com/Arclio/github-projects-mcp

This MCP server lets you:
- List and manage Projects V2 for users/orgs
- Work with project items (issues, PRs, drafts)
- Create issues and add them to projects
- Update project item fields i.e moving a ticket across the board.

Feel free to check it out and give a star if you find it helpful or interesting.

sirkitree:
possibly related since i understand Projects to be on GraphQL: https://github.com/github/github-mcp-server/issues/338

bczegeny:
Will this support status for items in a project, start and end dates?

Statuses very from project to project, and are not shared across projects at an org level. 
Dates tend to also have a variety of labeling names
Labels: Start, End, Start date, End Date, Target Date, Delivery Date

Will the tool calls be flexible enough to service prompts like:
- Tell me whats over due this week?
- Write me a weekly update on progress this week?
- Tell me what team member has too much work on their plate? (this should look for # of assigned issues, any custom fields around priority, story points, number of sub issues and progress etc?

tynes:
> At Arclio, we're excited to share Github Projects MCP Server to manage projects. We also needed this functionality for our own projects, so we created this MCP Server which is open-source: https://github.com/Arclio/github-projects-mcp
> 
> This MCP server lets you:
> 
> * List and manage Projects V2 for users/orgs
> * Work with project items (issues, PRs, drafts)
> * Create issues and add them to projects
> * Update project item fields i.e moving a ticket across the board.
> 
> Feel free to check it out and give a star if you find it helpful or interesting.

Thanks for sharing this! It would be awesome if this functionality could be upstreamed

[...]

PR #1194, opened 2025-10-08
Add update project item tool

JoannaaKL:
This pr:
- adds tool to update project item for an user or org
- formats all project tools to follow formatting standard
Closes: https://github.com/github/github-mcp-server/issues/44#issuecomment-3377640352

### DIFF
--- a/pkg/github/projects.go
+++ b/pkg/github/projects.go
@@ -691,13 +691,9 @@ func UpdateProjectItem(getClient GetClientFn, t translations.TranslationHelperFu
 				mcp.Required(),
 				mcp.Description("The unique identifier of the project item. This is not the issue or pull request ID."),
 			),
-			mcp.WithNumber("field_id",
+			mcp.WithObject("updated_field",
 				mcp.Required(),
-				mcp.Description("The unique identifier of the project field to be updated."),
-			),
-			mcp.WithObject("field_value",
-				mcp.Required(),
-				mcp.Description("The new value for the field: For text, number, and date fields, provide the new value directly. For single select and iteration fields, provide the ID of the option or iteration. To clear the field, set this to null. Example: {\"field_value\": \"Done\"}"),
+				mcp.Description("Object consisting of the ID of the project field to update and the new value for the field. To clear the field, set \"value\" to null. Example: {\"id\": 123456, \"value\": \"New Value\"}"),
 			),
 		), func(ctx context.Context, req mcp.CallToolRequest) (*mcp.CallToolResult, error) {
 			owner, err := RequiredParam[string](req, "owner")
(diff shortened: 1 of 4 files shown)


tonytrg:
I am having a hard time understanding `new_field`, not too familiar with the api of projects but it seems hard to understand to me only from the mcps side of description.

https://docs.github.com/en/rest/projects/items?apiVersion=2022-11-28#update-project-item-for-organization--parameters

fields array of objects Required
A list of field updates to apply.

id integer Required
The ID of the project field to update.

value null or string or number Required
The new value for the field:

For text, number, and date fields, provide the new value directly.
For single select and iteration fields, provide the ID of the option or iteration.
To clear the field, set this to null.

pkg/github/projects.go
```diff
@@ -558,14 +668,135 @@ func AddProjectItem(getClient GetClientFn, t translations.TranslationHelperFunc)
...
+				mcp.Required(),
+				mcp.Description("The project's number."),
+			),
+			mcp.WithNumber("item_id",
+				mcp.Required(),
+				mcp.Description("The unique identifier of the project item. This is not the issue or pull request ID."),
+			),
+			mcp.WithObject("new_field",
+				mcp.Required(),
+				mcp.Description("Object consisting of the ID of the project field to update and the new value for the field. To clear the field, set \"value\" to null. Example: {\"id\": 123456, \"value\": \"New Value\"}"),
```
tonytrg, line 696: seems like we are looking for a map struct.

Should it look something like this

				mcp.Items(
					map[string]interface{}{
						"type": "string",
					},
				),

tonytrg: I am not sure if we can rely on the llm to always understand and format a correct parseable struct

tonytrg: ah ok i see it needs to be more flexible
```
			mcp.WithObject("new_field",
				mcp.Required(),
				mcp.Description("Object consisting of the ID of the project field to update and the new value for the field. To clear the field, set \"value\" to null."),
				mcp.Properties(map[string]any{
					"id": map[string]any{
						"type":        "number",
						"description": "The ID of the project field to update (required)",
					},
					"value": map[string]any{
						"description": "The new value for the field. Can be a string for text fields, a number for number/single_select/iteration fields, or null to clear the field (required)",
					},
				}),
```

would something like that help?

almaleksia: I'd propose we split it into 2 fields which is field_id and field_value

JoannaaKL: How the types would look like then? We can use `WithNumber` for the field_id but field value needs to stay an object

JoannaaKL: We can do it 👍

tonytrg: @almaleksia @JoannaaKL i updated the struct the one before was wrong

pkg/github/projects.go
```diff
@@ -558,14 +668,146 @@ func AddProjectItem(getClient GetClientFn, t translations.TranslationHelperFunc)
...
+				mcp.Required(),
+				mcp.Description("The unique identifier of the project item. This is not the issue or pull request ID."),
+			),
+			mcp.WithNumber("field_id",
+				mcp.Required(),
+				mcp.Description("The unique identifier of the project field to be updated."),
+			),
+			mcp.WithObject("field_value",
+				mcp.Required(),
+				mcp.Description("The new value for the field: For text, number, and date fields, provide the new value directly. For single select and iteration fields, provide the ID of the option or iteration. To clear the field, set this to null. Example: {\"id\": 123456, \"value\": \"Done\"}"),
```
almaleksia: I think example should be changed

pkg/github/projects.go
```diff
@@ -558,14 +668,146 @@ func AddProjectItem(getClient GetClientFn, t translations.TranslationHelperFunc)
...
+			if err != nil {
+				return mcp.NewToolResultError(err.Error()), nil
+			}
+
+			rawFieldValue, exists := req.GetArguments()["field_value"]
+			if !exists {
+				return mcp.NewToolResultError("missing required parameter: field_value"), nil
+			}
+
+			fieldValue, ok := rawFieldValue.(map[string]any)
```
almaleksia: I think this needs to be changed

tonytrg:
💯 thanks for making this complicated one work

note for future as i dont want to block this merge:
- correct parsing of the iteration ids is missing right now, but as that actually should be part of the sdk we can fix that later on

PR #1202, opened 2025-10-10
Allow unknown toolsets to be ignored

omgitsads:
Allow unknown toolsets to be optionally ignored, rather than raising an error.


pkg/toolsets/toolsets.go
```diff
@@ -203,23 +203,33 @@ func (tg *ToolsetGroup) IsEnabled(name string) bool {
...
 
-func (tg *ToolsetGroup) EnableToolsets(names []string) error {
+type EnableToolsetsOptions struct {
+	IgnoreUnknown bool
+}
+
+func (tg *ToolsetGroup) EnableToolsets(names []string, options *EnableToolsetsOptions) error {
+	if options == nil {
+		options = &EnableToolsetsOptions{
+			IgnoreUnknown: false,
```
almaleksia: shall we set it to true by default?

PR #1210, opened 2025-10-13
restructure readme for better ghes ghec visibility

tonytrg:
We had multiple complaints that the docs were lacking ghes ghec documentation. This refactor highlights their configuration and fixes a error in the example url.

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,6 +98,27 @@ The default configuration is:
 
 See [Remote Server Documentation](docs/remote-server.md) for full details on remote server configuration, toolsets, headers, and advanced usage. This file provides comprehensive instructions and examples for connecting, customizing, and installing the remote GitHub MCP Server in VS Code and other MCP hosts.
 
+#### Enterprise Cloud with data residency (ghe.com)
+
+GitHub Enterprise Cloud can also make use of the remote server.
+
+Example for `https://octocorp.ghe.com`:
+```
+{
+    ...
+    "proxima-github": {
+      "type": "http",
+      "url": "https://copilot-api.octocorp.ghe.com/mcp",
+      "headers": {
+        "Authorization": "Bearer ${input:github_mcp_pat}"
+      }
+    },
+    ...
+}
+```
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -160,6 +160,33 @@ To keep your GitHub PAT secure and reusable across different MCP hosts:
 
 </details>
 
+## GitHub Enterprise Server and Enterprise Cloud with data residency (ghe.com)
+
+The flag `--gh-host` and the environment variable `GITHUB_HOST` can be used to set
+the hostname for GitHub Enterprise Server or GitHub Enterprise Cloud with data residency.
+
+- For GitHub Enterprise Server, prefix the hostname with the `https://` URI scheme, as it otherwise defaults to `http://`, which GitHub Enterprise Server does not support.
+- For GitHub Enterprise Cloud with data residency, use `https://copilot-api.YOURSUBDOMAIN.ghe.com` as the hostname.
```
omgitsads: This should not be `copilot-api` right? just `api.YOURSUBDOMAIN.ghe.com`?

omgitsads: Or not even `api.`?

tonytrg: fixed

tonytrg: api is not needed we infer all paths from the url, so we just check for the ghe.com pattern.

README.md
```diff
@@ -160,6 +181,33 @@ To keep your GitHub PAT secure and reusable across different MCP hosts:
 
 </details>
 
+### GitHub Enterprise Server and Enterprise Cloud with data residency (ghe.com)
```
omgitsads, line 184: Should this be moved under VSCode? The config is specifically for VSCode. Also if we're going to use `${input:github_token}` perhaps that also needs to be provided in the config, though it's somewhat redundant doing it everywhere.

tonytrg: i want to clearly seperate it - so its clear these settings are client agnostic.
I will do a later review of the whole readme structure, as we keep getting complaints about settings/questions which  are already in the readme docs.

README.md
hahaha1082: > This should not be `copilot-api` right? just `api.YOURSUBDOMAIN.ghe.com`?

PR #1217, opened 2025-10-14
chore(mcp/server.json): improve the OCI package specification

axel7083:
## Description

Since this repository has a `server.json` it is listed in the MCP registry[^1], however some fields could be improved to specify which arguments are mandatory, and using the synthax `-e <name>=<value>` for environment variables.

We also need to specify the `format` of arguments, as reading from the spec (https://static.modelcontextprotocol.io/schemas/2025-09-29/server.schema.json) the default is `string` and `-it` and `--rm` are booleans

[^1]: https://registry.modelcontextprotocol.io/v0/servers/io.github.github%2Fgithub-mcp-server


axel7083:
cc @MattBabbage (I saw that you were the one to introduce the server.json)

PR #1227, opened 2025-10-15
Update readme for remote only tools

omgitsads:
Update the readme to call out remote only toolsets and add them to the remote doc


LuluBeatson:
Updated sections

- README.md > [Additional Tools in Remote GitHub MCP Server](https://github.com/github/github-mcp-server/blob/update-readme-for-remote-only-tools/README.md#additional-tools-in-remote-github-mcp-server)
- docs/remote-server.md > [Additional Remote Server Toolsets](https://github.com/github/github-mcp-server/blob/update-readme-for-remote-only-tools/docs/remote-server.md#additional-remote-server-toolsets)

PR #1232, opened 2025-10-15
Add Gemini CLI extension

marzvrover:
Adds a `gemini-extension.json` manifest file so that the remote MCP server can be installed with the `gemini extensions install https://github.com/github/github-mcp-server` command.

Also updates the env var name for the PAT in all Gemini CLI cases to be `GITHUB_MCP_PAT`. Gemini CLI uses environment variables plus those found in the `.env` file, so I caution using a more explicitly named variable to prevent a user accidentally using a PAT they have in their environment that contains permissions that they may not want their MCP instance to have access to.

### DIFF
--- a/docs/installation-guides/install-gemini-cli.md
+++ b/docs/installation-guides/install-gemini-cli.md
@@ -30,7 +30,7 @@ After securely storing your PAT, you can add the GitHub MCP server configuration
 
 > **Note:** For the most up-to-date configuration options, see the [main README.md](../../README.md).
 
-### Method 1: Remote Server (Recommended)
+### Method 1: Gemini Extension (Recommended)
 
 The simplest way is to use GitHub's hosted MCP server via our gemini extension.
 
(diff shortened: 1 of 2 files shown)


gemini-extension.json
```diff
@@ -0,0 +1,12 @@
+{
+	"name": "github",
+	"version": "1.0.0",
```
siminapasat, line 3: can you also add a description pls?

tonytrg: `Connect AI assistants to GitHub - manage repos, issues, PRs, and workflows through natural language.` is the description we use for the registry can we stick to the same description here too.

docs/installation-guides/install-gemini-cli.md
```diff
@@ -32,21 +32,13 @@ After securely storing your PAT, you can add the GitHub MCP server configuration
 
 ### Method 1: Remote Server (Recommended)
```
tonytrg: Can we turn that into two seperate methods.
Id prefer not to delete the manual instruction, as there are prob some people who like to configure it themselves

### Method 1: Extension (Recommended)
...

### Method 2: Manual config

tonytrg:
left some comments, but tested the flow and it works nicely great stuff!

gemini-extension.json
marzvrover: There is no extension description in the docs or [their code](https://github.com/google-gemini/gemini-cli/blob/main/packages/core/src/config/config.ts#L135-L144). But there is one for MCP servers, so I'll add one there.

pachecocordovamoiseseduardo-byte (approved):

PR #1273, opened 2025-10-21
Split up tooling utility

tonytrg:
Closes:

Splits up the toolset functionality into sub functions, as they were not really reusable.

### DIFF
--- a/pkg/github/tools.go
+++ b/pkg/github/tools.go
@@ -502,7 +502,7 @@ func CleanToolsets(enabledToolsets []string) ([]string, []string) {
 }
 
 func RemoveToolset(tools []string, toRemove string) []string {
-	result := []string{}
+	result := make([]string, 0, len(tools))
 	for _, tool := range tools {
 		if tool != toRemove {
 			result = append(result, tool)


pkg/github/tools.go
```diff
@@ -445,3 +445,77 @@ func GenerateToolsetsHelp() string {
...
+	result := make([]string, 0, len(tools))
+	for _, tool := range tools {
+		if tool != toRemove {
+			result = append(result, tool)
+		}
+	}
+	return result
+}
+
+func ContainsToolset(tools []string, toCheck string) bool {
```
JoannaaKL, line 514: Can we actually operate on maps everywhere? Right now this function is O(N) because it needs to iterate over all tools in the array, with a map it'd be an O(1). You'd need to refactor `CleanToolsets` to achieve that

JoannaaKL: Or at least `CleanToolsets` could return a map and an array

tonytrg: we are passing lists everywhere, also the libraries expect lists too.

As the tool lists are pretty small we have a neglible performance gain and would have to transform the lists into maps which is more costly probably.

But the biggest hit would be having now to manage transformations between lists and maps everywhere in the codebase where these functions are used, if the utils are maps then they couldnt be used in all the places currently(they are lists everywhere).

tonytrg: https://github.com/github/github-mcp-server/pull/1273/files#diff-efffdb5f6c0e96b376f9f02d7a887ea629058c21a95e161981c5c08e41f12e0cR109 example we would transform it into a map and then back to a list. iterating through the list is the lesser evil

StefanXhunga (commented):
Hello everyone, thank you for support

PR #1282, opened 2025-10-22
projects: add item field support

tmelliottjr:
Towards: https://github.com/github/github-mcp-server/issues/44

This PR adds an optional `fields` parameter to `list_project_items` and `get_project_item` tools, allowing users to specify which custom field IDs to include in responses. Previously, only the title field was returned by default.

The implementation includes a new `ManageProjectItemsPrompt` that provides step-by-step guidance for working with GitHub Projects, including discovering projects, viewing available fields, querying items with specific field values, and updating field values on project items. 

The type system is updated to distinguish between `projectV2Field` (field definitions) and `projectV2FieldValues` (field values on items), ensuring proper serialization of field data. Comprehensive test coverage validates field selection functionality across both organization and user-owned projects.

### DIFF
--- a/pkg/github/projects.go
+++ b/pkg/github/projects.go
@@ -76,12 +76,11 @@ func ListProjects(getClient GetClientFn, t translations.TranslationHelperFunc) (
 			projects := []github.ProjectV2{}
 			minimalProjects := []MinimalProject{}
 
-			opts := listProjectsOptions{}
-			opts.PerPage = perPage
-
-			if queryStr != "" {
-				opts.Query = queryStr
+			opts := listProjectsOptions{
+				paginationOptions:  paginationOptions{PerPage: perPage},
+				filterQueryOptions: filterQueryOptions{Query: queryStr},
 			}
+
 			url, err = addOptions(url, opts)
 			if err != nil {
 				return nil, fmt.Errorf("failed to add options to request: %w", err)
@@ -258,8 +257,7 @@ func ListProjectFields(getClient GetClientFn, t translations.TranslationHelperFu
 			}
 			projectFields := []projectV2Field{}
 
-			opts := paginationOptions{}
-			opts.PerPage = perPage
+			opts := paginationOptions{PerPage: perPage}
 
 			url, err = addOptions(url, opts)
 			if err != nil {
@@ -448,15 +446,10 @@ func ListProjectItems(getClient GetClientFn, t translations.TranslationHelperFun
 			}
 			projectItems := []projectV2Item{}
 
-			opts := listProjectItemsOptions{}
-			opts.PerPage = perPage
-
-			if queryStr != "" {
-				opts.Query = queryStr
-			}
-
-			if len(fields) > 0 {
-				opts.Fields = fields
+			opts := listProjectItemsOptions{
+				paginationOptions:     paginationOptions{PerPage: perPage},
+				filterQueryOptions:    filterQueryOptions{Query: queryStr},
+				fieldSelectionOptions: fieldSelectionOptions{Fields: fields},
 			}
 
 			url, err = addOptions(url, opts)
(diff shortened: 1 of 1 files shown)


pkg/github/projects.go
```diff
@@ -877,21 +916,53 @@ type projectV2Field struct {
 	UpdatedAt *github.Timestamp `json:"updated_at,omitempty"` // The time when this field was last updated.
 }
 
+type projectV2ItemFieldValue struct {
```
stephenotalora, line 912: small oversight but types in Go start with a capital

```suggestion
type ProjectV2ItemFieldValue struct {
```

we will need to go with a vscode refactor to change this properly, so please don't commit my suggestion 😄

stephenotalora: Also for us to reuse this in the future I think it would be best if we commit it directly into [google/go-github](https://github.com/google/go-github/blob/master/github/projects.go) at the same level of `ProjectV2` and `ProjectV2Field` although I can take care of that as part of the refactor in https://github.com/github/memex/issues/20995 if my understanding is correct, as to not block this PR 👍🏼

pkg/github/projects.go
```diff
@@ -877,21 +916,53 @@ type projectV2Field struct {
...
+	ContentNodeID *string                    `json:"content_node_id,omitempty"`
+	ProjectURL    *string                    `json:"project_url,omitempty"`
+	ContentType   *string                    `json:"content_type,omitempty"`
+	Creator       *github.User               `json:"creator,omitempty"`
+	CreatedAt     *github.Timestamp          `json:"created_at,omitempty"`
+	UpdatedAt     *github.Timestamp          `json:"updated_at,omitempty"`
+	ArchivedAt    *github.Timestamp          `json:"archived_at,omitempty"`
+	ItemURL       *string                    `json:"item_url,omitempty"`
+	Fields        []*projectV2ItemFieldValue `json:"fields,omitempty"`
+}
```
stephenotalora, line 934: similar to this [comment](https://github.com/github/github-mcp-server/pull/1282/files#r2453585305) for the remaining structs, we should also capitalize the first letter as previously mentioned imo as to limit the diff from `google/go-github`

pkg/github/projects.go
```diff
@@ -958,3 +1021,56 @@ func addOptions(s string, opts any) (string, error) {
...
+			mcp.WithArgument("owner", mcp.ArgumentDescription("The owner of the project (user or organization name)"), mcp.RequiredArgument()),
+			mcp.WithArgument("owner_type", mcp.ArgumentDescription("Type of owner: 'user' or 'org'"), mcp.RequiredArgument()),
+		), func(_ context.Context, request mcp.GetPromptRequest) (*mcp.GetPromptResult, error) {
+			owner := request.Params.Arguments["owner"]
+			ownerType := request.Params.Arguments["owner_type"]
+
+			messages := []mcp.PromptMessage{
+				{
+					Role:    "user",
+					Content: mcp.NewTextContent("You are an assistant helping users work with GitHub Projects (Projects V2). Your role is to help them discover projects, understand project fields, query items, and update field values on project items."),
```
JoannaaKL:
```suggestion
					Content: mcp.NewTextContent("You are an assistant helping users work with GitHub Projects V2. Your role is to help them discover projects, understand project fields, query items, and update field values on project items."),
```

pkg/github/projects.go
JoannaaKL: Starting with a capital will make this struct exported, is there particular reason for that? I think we can stick to unexported struct

pkg/github/projects.go
JoannaaKL: Above comment :)

pkg/github/projects.go
tmelliottjr: @stephenotalora, I'm going to leave this as is and the `go-github` work can follow suite afterwards. My goal here is unblock this functionality with the current implementation until we the `go-github` changes can be made, released, and incorporated back here.

pkg/github/projects.go
tmelliottjr: Replied above.

pkg/github/projects.go
tmelliottjr: 👍🏻

JoannaaKL (approved):

pkg/github/projects.go
stephenotalora: yes thank you both agreed!

PR #1305, opened 2025-10-25
Add discussion metadata fields to get_discussion tool

### LINKED ISSUE #1303: get_discussion not returning whether discussion is open/closed
The get_discussion tool is not correctly returning discussion state information. The implementation is missing key fields that indicate whether a discussion is open, closed, or answered.

No Discussion Status: The current implementation doesn't include fields that would indicate:

* Whether the discussion is answered (isAnswered)
* The discussion state (open/closed)
* Answer information (answeredAt, answerChosenAt)

Inconsistent with Issues: The issues.go implementation properly includes a State field in its GraphQL queries and data structures, but discussions do not.

### Affected version

5e5e80ac4cf3bbde93946234e530b4f0211833c6

get_discussion should be consistent with get_issue and report if the discussion is open/closed

Higangssh:
## Problem
  The `get_discussion` tool lacked important metadata fields already available in `get_issue`, making it difficult
  to determine discussion status.

  ## Solution
  Added discussion metadata fields:
  - **closed** - Boolean indicating if the discussion is closed
  - **isAnswered** - Whether discussion has an accepted answer
  - **answerChosenAt** - Timestamp when answer was selected (optional)

  ## Implementation Details
  Modified `GetDiscussion` to return `map[string]interface{}` instead of `github.Discussion` struct because
  go-github v74's Discussion type lacks `isAnswered` and `answerChosenAt` fields. This approach mirrors existing
  functions: `ListDiscussions` and `GetDiscussionComments` use similar patterns.

  ## Changes Made
  - Updated Discussion field mappings to match GitHub GraphQL API schema
  - Replaced non-existent `State` field with `Closed` (Boolean)
  - Removed non-existent `AnsweredAt` field
  - Using `AnswerChosenAt` for answer timestamp
  - Added nil checks for optional timestamp fields
  - Updated all test cases and mock data
  - All tests passing ✅

  ## Testing
  All tests pass with updated snapshots. Linter reports no issues. Backward compatibility verified.

  Fixes #1303


pkg/github/discussions.go
```diff
@@ -44,11 +44,15 @@ type DiscussionFragment struct {
...
-	Number    githubv4.Int
-	Title     githubv4.String
-	CreatedAt githubv4.DateTime
-	UpdatedAt githubv4.DateTime
-	Author    struct {
+	Number         githubv4.Int
+	Title          githubv4.String
+	CreatedAt      githubv4.DateTime
+	UpdatedAt      githubv4.DateTime
+	State          githubv4.String
```
tommaso-moro: I believe `State`, `AnsweredAt` and `IsAnswered` are not valid fields for the `Discussion` type

tommaso-moro (changes requested):
Hi @Higangssh thank you for your work on this. I have tested it locally and it is not working (screenshot below), it looks like some of the fields are invalid. For example, `State`, `IsAnswered` and `AnsweredAt` don't exist on `Discussion`. 

Instead of `State` you can use the `closed` field and then update the `response` map to use `"closed": bool(d.Closed)`. As to the other two, I would recommend using `answerChosenAt`, and deriving `isAnswered` from it (specifically, it will be equivalent to `answerChosenAt != nil`). 

I also recommend using the official [GitHub GraphQL API Explorer](https://docs.github.com/en/graphql/overview/explorer) to play around with the query and see which fields are available and what they return.

If you can, it would be great if you could update this PR. Otherwise, I am also happy to do it if you prefer. Thank you!

Image

Higangssh:
Hi @tommaso-moro, thank you so much for the detailed review and suggestions!

I've updated the PR following your recommendations:
- ✅ `State` → `Closed` (Boolean)
- ✅ `AnsweredAt` → Removed (using `answerChosenAt` instead)

### Regarding `isAnswered`:

Following your suggestion to use the GraphQL API Explorer, I tested the available fields and found that `isAnswered` is actually a valid field in the Discussion type.

**Evidence:**
1. **Official Documentation**: The [GitHub GraphQL Reference](https://docs.github.com/en/graphql/reference/objects#discussion) lists `isAnswered (Boolean)` as a field for Discussion objects.

2. **GraphQL Query Test**: I tested with both answered and unanswered discussions:
```graphql
   query {
     repository(owner: "github", name: "github-mcp-server") {
       discussion(number: 1248) {
         isAnswered
         answerChosenAt
       }
     }
   }
```
   - Answered discussion (1248): `{ "isAnswered": true, "answerChosenAt": "2025-10-20T09:57:37Z" }`
   - Unanswered discussion (1088): `{ "isAnswered": false, "answerChosenAt": null }`

Both queries executed successfully without errors, confirming that `isAnswered` is valid.

Given this, I've kept `isAnswered` as a direct field query rather than deriving it. This approach uses the field that GitHub's API provides and is more efficient than client-side calculation.

Would you prefer I derive it from `answerChosenAt != nil` instead, or is this approach acceptable?

Thank you again for your help! 🙏


---
image
---
image

tommaso-moro (approved):
Nice one! Thank you for your work on this 🚀

PR #1352, opened 2025-11-03
projects: add server instructions

tmelliottjr:
This PR updates the tool/argument descriptions for projects related tools to better enable tool selection and project item filter.

Additionally, this PR introduces pagination for `list_*` operations to enable effective querying.

### DIFF
--- a/pkg/github/__toolsnaps__/list_project_fields.snap
+++ b/pkg/github/__toolsnaps__/list_project_fields.snap
@@ -6,6 +6,14 @@
   "description": "List Project fields for a user or org",
   "inputSchema": {
     "properties": {
+      "after": {
+        "description": "Forward pagination cursor from previous pageInfo.nextCursor.",
+        "type": "string"
+      },
+      "before": {
+        "description": "Backward pagination cursor from previous pageInfo.prevCursor (rare).",
+        "type": "string"
+      },
       "owner": {
         "description": "If owner_type == user it is the handle for the GitHub user account. If owner_type == org it is the name of the organization. The name is not case sensitive.",
         "type": "string"
@@ -19,7 +27,7 @@
         "type": "string"
       },
       "per_page": {
-        "description": "Number of results per page (max 100, default: 30)",
+        "description": "Results per page (max 50)",
         "type": "number"
       },
       "project_number": {
--- a/pkg/github/instructions.go
+++ b/pkg/github/instructions.go
@@ -62,6 +62,75 @@ Check 'list_issue_types' first for organizations to use proper issue types. Use 
 		return `## Discussions
 		
 Use 'list_discussion_categories' to understand available categories before creating discussions. Filter by category for better organization.`
+	case "projects":
+		return `## Projects
+
+Workflow: 1) list_project_fields (get field IDs), 2) list_project_items (with pagination), 3) optional updates.
+
+Field usage:
+	- Call list_project_fields first to understand available fields and get IDs/types before filtering.
+	- Use EXACT returned field names (case-insensitive match). Don't invent names or IDs.
+	- Iteration synonyms (sprint/cycle) only if that field exists; map to the actual name (e.g. sprint:@current).
+	- Only include filters for fields that exist and are relevant.
+
+Pagination (mandatory):
+	- Loop while pageInfo.hasNextPage=true using after=pageInfo.nextCursor.
+	- Keep query, fields, per_page IDENTICAL on every page.
+	- Use before=pageInfo.prevCursor only when explicitly navigating to a previous page.
+
+Counting rules:
+	- Count items array length after full pagination.
+	- Never count field objects, content, or nested arrays as separate items.
+
+Summary vs list:
+	- Summaries ONLY if user uses verbs: analyze | summarize | summary | report | overview | insights.
+	- Listing verbs (list/show/get/fetch/display/enumerate) → enumerate + total.
+
+Self-check before returning:
+	- Paginated fully
+	- Correct IDs used
+	- Field names valid
+	- Summary only if requested.
+
+Return COMPLETE data or state what's missing (e.g. pages skipped).
+
+list_project_items query rules:
+Query string - For advanced filtering of project items using GitHub's project filtering syntax:
+
+MUST reflect user intent; strongly prefer explicit content type if narrowed:
+	- "open issues" → state:open is:issue
+	- "merged PRs" → state:merged is:pr
+	- "items updated this week" → updated:>@today-7d (omit type only if mixed desired)
+	- "list all P1 priority items" → priority:p1 (omit state if user wants all, omit type if user specifies "items")
+	- "list all open P2 issues" → is:issue state:open priority:p2 (include state if user wants open or closed, include type if user specifies "issues" or "PRs")
+	- "all open issues I'm working on" → is:issue state:open assignee:@me
+
+Query Construction Heuristics:
+	a. Extract type nouns: issues → is:issue | PRs, Pulls, or Pull Requests → is:pr | tasks/tickets → is:issue (ask if ambiguity)
+	b. Map temporal phrases: "this week" → updated:>@today-7d
+	c. Map negations: "excluding wontfix" → -label:wontfix
+	d. Map priority adjectives: "high/sev1/p1" → priority:high OR priority:p1 (choose based on field presence)
+	e. When filtering by label, always use wildcard matching to account for cross-repository differences or emojis: (e.g. "bug 🐛" → label:*bug*)
+	f. When filtering by milestone, always use wildcard matching to account for cross-repository differences: (e.g. "v1.0" → milestone:*v1.0*)
+
+Syntax Essentials (items):
+   AND: space-separated. (label:bug priority:high).
+   OR: comma inside one qualifier (label:bug,critical).
+   NOT: leading '-' (-label:wontfix).
+   Hyphenate multi-word field names. (team-name:"Backend Team", story-points:>5).
+   Quote multi-word values. (status:"In Review" team-name:"Backend Team").
+   Ranges: points:1..3, updated:<@today-30d.
+   Wildcards: title:*crash*, label:bug*.
+	 Assigned to User: assignee:@me | assignee:username | no:assignee
+
+Common Qualifier Glossary (items):
+   is:issue | is:pr | state:open|closed|merged | assignee:@me|username | label:NAME | status:VALUE |
+   priority:p1|high | sprint-name:@current | team-name:"Backend Team" | parent-issue:"org/repo#123" |
+   updated:>@today-7d | title:*text* | -label:wontfix | label:bug,critical | no:assignee | has:label
+
+Never:
+   - Infer field IDs; fetch via list_project_fields.
+   - Drop 'fields' param on subsequent pages if field values are needed.`
 	default:
 		return ""
 	}
--- a/pkg/github/projects.go
+++ b/pkg/github/projects.go
@@ -23,11 +23,12 @@ const (
 	ProjectAddFailedError    = "failed to add a project item"
 	ProjectDeleteFailedError = "failed to delete a project item"
 	ProjectListFailedError   = "failed to list project items"
+	MaxProjectsPerPage       = 50
 )
 
 func ListProjects(getClient GetClientFn, t translations.TranslationHelperFunc) (tool mcp.Tool, handler server.ToolHandlerFunc) {
 	return mcp.NewTool("list_projects",
-			mcp.WithDescription(t("TOOL_LIST_PROJECTS_DESCRIPTION", "List Projects for a user or org")),
+			mcp.WithDescription(t("TOOL_LIST_PROJECTS_DESCRIPTION", `List Projects for a user or organization`)),
 			mcp.WithToolAnnotation(mcp.ToolAnnotation{
 				Title:        t("TOOL_LIST_PROJECTS_USER_TITLE", "List projects"),
 				ReadOnlyHint: ToBoolPtr(true),
@@ -40,28 +41,38 @@ func ListProjects(getClient GetClientFn, t translations.TranslationHelperFunc) (
 				mcp.Description("If owner_type == user it is the handle for the GitHub user account. If owner_type == org it is the name of the organization. The name is not case sensitive."),
 			),
 			mcp.WithString("query",
-				mcp.Description("Filter projects by a search query (matches title and description)"),
+				mcp.Description(`Filter projects by title text and open/closed state; permitted qualifiers: is:open, is:closed; examples: "roadmap is:open", "is:open feature planning".`),
 			),
 			mcp.WithNumber("per_page",
-				mcp.Description("Number of results per page (max 100, default: 30)"),
+				mcp.Description(fmt.Sprintf("Results per page (max %d)", MaxProjectsPerPage)),
+			),
+			mcp.WithString("after",
+				mcp.Description("Forward pagination cursor from previous pageInfo.nextCursor."),
+			),
+			mcp.WithString("before",
+				mcp.Description("Backward pagination cursor from previous pageInfo.prevCursor (rare)."),
 			),
 		), func(ctx context.Context, req mcp.CallToolRequest) (*mcp.CallToolResult, error) {
 			owner, err := RequiredParam[string](req, "owner")
 			if err != nil {
 				return mcp.NewToolResultError(err.Error()), nil
 			}
+
 			ownerType, err := RequiredParam[string](req, "owner_type")
 			if err != nil {
 				return mcp.NewToolResultError(err.Error()), nil
 			}
+
 			queryStr, err := OptionalParam[string](req, "query")
 			if err != nil {
 				return mcp.NewToolResultError(err.Error()), nil
 			}
-			perPage, err := OptionalIntParamWithDefault(req, "per_page", 30)
+
+			pagination, err := extractPaginationOptions(req)
 			if err != nil {
 				return mcp.NewToolResultError(err.Error()), nil
 			}
+
 			client, err := getClient(ctx)
 			if err != nil {
 				return mcp.NewToolResultError(err.Error()), nil
@@ -77,7 +88,7 @@ func ListProjects(getClient GetClientFn, t translations.TranslationHelperFunc) (
 
 			minimalProjects := []MinimalProject{}
 			opts := &github.ListProjectsOptions{
-				ListProjectsPaginationOptions: github.ListProjectsPaginationOptions{PerPage: &perPage},
+				ListProjectsPaginationOptions: pagination,
 				Query:                         queryPtr,
 			}
 
@@ -100,14 +111,12 @@ func ListProjects(getClient GetClientFn, t translations.TranslationHelperFunc) (
 				minimalProjects = append(minimalProjects, *convertToMinimalProject(project))
 			}
 
-			if resp.StatusCode != http.StatusOK {
-				body, err := io.ReadAll(resp.Body)
-				if err != nil {
-					return nil, fmt.Errorf("failed to read response body: %w", err)
-				}
-				return mcp.NewToolResultError(fmt.Sprintf("failed to list projects: %s", string(body))), nil
+			response := map[string]any{
+				"projects": minimalProjects,
+				"pageInfo": buildPageInfo(resp),
 			}
-			r, err := json.Marshal(minimalProjects)
+
+			r, err := json.Marshal(response)
 			if err != nil {
 				return nil, fmt.Errorf("failed to marshal response: %w", err)
 			}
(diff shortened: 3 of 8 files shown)


pkg/github/projects.go
```diff
@@ -23,11 +23,12 @@ const (
 	ProjectAddFailedError    = "failed to add a project item"
 	ProjectDeleteFailedError = "failed to delete a project item"
 	ProjectListFailedError   = "failed to list project items"
+	MaxProjectsPerPage       = 50
 )
 
 func ListProjects(getClient GetClientFn, t translations.TranslationHelperFunc) (tool mcp.Tool, handler server.ToolHandlerFunc) {
 	return mcp.NewTool("list_projects",
-			mcp.WithDescription(t("TOOL_LIST_PROJECTS_DESCRIPTION", "List Projects for a user or org")),
+			mcp.WithDescription(t("TOOL_LIST_PROJECTS_DESCRIPTION", `List Projects for a user or organization   `)),
```
JoannaaKL:
```suggestion
			mcp.WithDescription(t("TOOL_LIST_PROJECTS_DESCRIPTION", `List Projects for a user or organization`)),
```

pkg/github/projects.go
```diff
@@ -363,7 +392,49 @@ func GetProjectField(getClient GetClientFn, t translations.TranslationHelperFunc
 
 func ListProjectItems(getClient GetClientFn, t translations.TranslationHelperFunc) (tool mcp.Tool, handler server.ToolHandlerFunc) {
 	return mcp.NewTool("list_project_items",
-			mcp.WithDescription(t("TOOL_LIST_PROJECT_ITEMS_DESCRIPTION", "List Project items for a user or org")),
+			mcp.WithDescription(t("TOOL_LIST_PROJECT_ITEMS_DESCRIPTION", `GitHub Projects V2 - List Project Items with advanced filtering and field selection
```
JoannaaKL: Maybe this can be added to [server instructions](https://github.com/github/github-mcp-server/blob/304f29a2fd336f6092e816b7b3cbf67729abe937/pkg/github/instructions.go#L10C6-L10C26)?

pkg/github/projects.go
```diff
@@ -381,37 +408,89 @@ func ListProjectItems(getClient GetClientFn, t translations.TranslationHelperFun
 				mcp.Description("The project's number."),
 			),
 			mcp.WithString("query",
-				mcp.Description("Search query to filter items"),
+				mcp.Description(`Query string - For advanced filtering of project items using GitHub's search syntax:
```
kerobbi: Could this be added to [server instructions](https://github.com/github/github-mcp-server/blob/304f29a2fd336f6092e816b7b3cbf67729abe937/pkg/github/instructions.go#L10C6-L10C26)?

pkg/github/projects.go
```diff
@@ -208,25 +228,35 @@ func ListProjectFields(getClient GetClientFn, t translations.TranslationHelperFu
...
 			),
 			mcp.WithNumber("per_page",
-				mcp.Description("Number of results per page (max 100, default: 30)"),
+				mcp.Description(fmt.Sprintf("Results per page (max %d). Keep constant across paginated requests; changing mid-sequence can complicate page traversal.", MaxProjectsPerPage)),
+			),
+			mcp.WithString("after",
+				mcp.Description("Forward pagination cursor. Use when the previous response's pageInfo.hasNextPage=true. Supply pageInfo.nextCursor as 'after' and immediately request the next page. LOOP UNTIL pageInfo.hasNextPage=false (don't stop early). Keep per_page identical for every page."),
+			),
+			mcp.WithString("before",
+				mcp.Description("Backward pagination cursor (rare): supply to move to the preceding page using pageInfo.prevCursor. Not needed for normal forward iteration."),
```
kerobbi: Same as above.

pkg/github/projects.go
```diff
@@ -40,28 +41,45 @@ func ListProjects(getClient GetClientFn, t translations.TranslationHelperFunc) (
...
 			),
 			mcp.WithNumber("per_page",
-				mcp.Description("Number of results per page (max 100, default: 30)"),
+				mcp.Description(fmt.Sprintf("Results per page (max %d). Keep constant across paginated requests; changing mid-sequence can complicate page traversal.", MaxProjectsPerPage)),
+			),
+			mcp.WithString("after",
+				mcp.Description("Forward pagination cursor. Use when the previous response's pageInfo.hasNextPage=true. Supply pageInfo.nextCursor as 'after' and immediately request the next page. LOOP UNTIL pageInfo.hasNextPage=false (don't stop early). Keep query and per_page identical for every page."),
+			),
+			mcp.WithString("before",
+				mcp.Description("Backward pagination cursor (rare): supply to move to the preceding page using pageInfo.prevCursor. Not needed for normal forward iteration."),
```
kerobbi: Could we make these descriptions more focused on param purpose and move the more detailed usage instructions to server instructions?

pkg/github/projects.go
kerobbi: Same as above.

pkg/github/projects.go
kerobbi: On a similar topic as the above comments, this description feels quite verbose - any chance we could consolidate everything into server instructions and keep this contained to purpose + brief examples if necessary?

pkg/github/projects.go
```diff
@@ -381,37 +408,89 @@ func ListProjectItems(getClient GetClientFn, t translations.TranslationHelperFun
...
 			),
 			mcp.WithNumber("per_page",
-				mcp.Description("Number of results per page (max 100, default: 30)"),
+				mcp.Description(fmt.Sprintf("Results per page (max %d). Keep constant across paginated requests; changing mid-sequence can complicate page traversal.", MaxProjectsPerPage)),
+			),
+			mcp.WithString("after",
+				mcp.Description("Forward pagination cursor. Use when the previous response's pageInfo.hasNextPage=true. Supply pageInfo.nextCursor as 'after' and immediately request the next page. LOOP UNTIL pageInfo.hasNextPage=false (don't stop early). Keep query, fields, and per_page identical for every page."),
+			),
+			mcp.WithString("before",
+				mcp.Description("Backward pagination cursor (rare): supply to move to the preceding page using pageInfo.prevCursor. Not needed for normal forward iteration."),
```
kerobbi: Same as the first two comments.

pkg/github/projects.go
tmelliottjr: Updated in https://github.com/github/github-mcp-server/pull/1352/commits/2ba80b83c227b88c544c7e79ef3189ad9e49242a

pkg/github/instructions.go
```diff
@@ -62,6 +62,104 @@ Check 'list_issue_types' first for organizations to use proper issue types. Use
...
+	- list_projects
+	- get_project
+	- list_project_fields
+	- get_project_field
+	- list_project_items
+	- get_project_item
+Write Tools:
+	- add_project_item
+	- update_project_item
+	- delete_project_item
```
kerobbi:
```suggestion
Workflow: 1) list_project_fields (get field IDs), 2) list_project_items (with pagination), 3) optional updates.
```

pkg/github/instructions.go
```diff
@@ -62,6 +62,104 @@ Check 'list_issue_types' first for organizations to use proper issue types. Use
...
+
+Field usage:
+	- Call list_project_fields first to understand available fields and get IDs/types before filtering.
+	- Use EXACT returned field names (case-insensitive match). Don't invent names or IDs.
+	- Iteration synonyms (sprint/cycle/iteration) only if that field exists; map to the actual name (e.g. sprint:@current).
```
kerobbi:
```suggestion
	- Iteration synonyms (sprint/cycle) only if that field exists; map to the actual name (e.g. sprint:@current).
```

pkg/github/instructions.go
kerobbi: Having the tools listed here is slightly redundant, as the model already knows all the tools it has available. What could be included instead is higher level instructions on how the tools should be used together.

pkg/github/instructions.go
```diff
@@ -62,6 +62,104 @@ Check 'list_issue_types' first for organizations to use proper issue types. Use
...
+	- Keep query, fields, per_page IDENTICAL on every page.
+	Backward (rare) flow:
+	- Use before=pageInfo.prevCursor only when explicitly navigating to a previous page.
+	Parameters:
+	- per_page: results per page (max 50). Choose a stable value; do not change mid-sequence.
+	- after: forward cursor from prior response (pageInfo.nextCursor).
+	- before: backward cursor from prior response (pageInfo.prevCursor); seldom needed.
+
+Fields parameter:
+	- Include field IDs on EVERY paginated list_project_items call if you need values. Omit → title only.
```
kerobbi: I think this section can be removed, as this is something which is already present in the `field` parameter description.

pkg/github/instructions.go
```diff
@@ -62,6 +62,104 @@ Check 'list_issue_types' first for organizations to use proper issue types. Use
...
+Pagination (mandatory):
+	Forward (normal) flow:
+	- Loop while pageInfo.hasNextPage=true using after=pageInfo.nextCursor.
+	- Keep query, fields, per_page IDENTICAL on every page.
+	Backward (rare) flow:
+	- Use before=pageInfo.prevCursor only when explicitly navigating to a previous page.
+	Parameters:
+	- per_page: results per page (max 50). Choose a stable value; do not change mid-sequence.
+	- after: forward cursor from prior response (pageInfo.nextCursor).
+	- before: backward cursor from prior response (pageInfo.prevCursor); seldom needed.
```
kerobbi:
```suggestion
	- Loop while pageInfo.hasNextPage=true using after=pageInfo.nextCursor.
	- Keep query, fields, per_page IDENTICAL on every page.
	- Use before=pageInfo.prevCursor only when explicitly navigating to a previous page.
```

kerobbi: Just a suggestion to try and compact things while keeping the key instructions - we should always be aware of the context window usage.

pkg/github/instructions.go
```diff
@@ -62,6 +62,104 @@ Check 'list_issue_types' first for organizations to use proper issue types. Use
...
+Self-check before returning:
+	- Paginated fully
+	- Dedupe by id/node_id
+	- Correct IDs used
+	- Field names valid
+	- Summary only if requested.
+
+Return COMPLETE data or state what's missing (e.g. pages skipped).
+
+list_project_items query rules:
```
kerobbi, line 97: This section (lines 123 - 156) on query construction is quite long. What I would suggest doing here is just keeping anything which is projects specific or that we think that the model does not know already, and removing the rest.

We do have a `query` parameter in other tools (e.g., `search_pull_requests`) that could be used as a reference in case you want to update the `list_project_items` `query` param. In general the model is quite good at constructing queries so it probably won't need this level of guidance. 

If we were to consider adding guidance on search syntax in the future, that would apply across the entire server as a whole, not just projects tools, so it wouldn't live in these projects specific instructions.

pkg/github/instructions.go
```diff
@@ -62,6 +62,104 @@ Check 'list_issue_types' first for organizations to use proper issue types. Use
...
+   Wildcards: title:*crash*, label:bug*.
+	 Assigned to User: assignee:@me | assignee:username | no:assignee
+
+Common Qualifier Glossary (items):
+   is:issue | is:pr | state:open|closed|merged | assignee:@me|username | label:NAME | status:VALUE |
+   priority:p1|high | sprint-name:@current | team-name:"Backend Team" | parent-issue:"org/repo#123" |
+   updated:>@today-7d | title:*text* | -label:wontfix | label:bug,critical | no:assignee | has:label
+
+Pagination Mandate:
+   Do not analyze until ALL pages fetched (loop while pageInfo.hasNextPage=true). Always reuse identical query, fields, per_page.
```
kerobbi: Some of this is a repetition of what is being said in the `Pagination` section. We could either remove it or move it to that section and ideally shorten it.

pkg/github/projects.go
```diff
@@ -40,28 +41,45 @@ func ListProjects(getClient GetClientFn, t translations.TranslationHelperFunc) (
 				mcp.Description("If owner_type == user it is the handle for the GitHub user account. If owner_type == org it is the name of the organization. The name is not case sensitive."),
 			),
 			mcp.WithString("query",
-				mcp.Description("Filter projects by a search query (matches title and description)"),
+				mcp.Description(`Filter projects by a search query
+				
+Scope: title text + open/closed state.
+PERMITTED qualifiers: is:open, is:closed (state), simple title terms.
+FORBIDDEN: is:issue, is:pr, assignee:, label:, status:, sprint-name:, parent-issue:, team-name:, priority:, etc.
```
kerobbi: Out of curiosity, have you found that the model used these forbidden filters or are you adding them here as a precaution?

pkg/github/instructions.go
```diff
@@ -62,6 +62,104 @@ Check 'list_issue_types' first for organizations to use proper issue types. Use
...
+	- Never count field objects, content, or nested arrays as separate items.
+	- item.id = project item ID (for updates/deletes). item.content.id = underlying issue/PR ID.
+
+Summary vs list:
+	- Summaries ONLY if user uses verbs: analyze | summarize | summary | report | overview | insights.
+	- Listing verbs (list/show/get/fetch/display/enumerate) → enumerate + total.
+
+Examples:
+	- list_projects: "roadmap is:open"
+	- list_project_items: state:open is:issue sprint:@current priority:high updated:>@today-7d
```
kerobbi:
```suggestion
```

kerobbi: These are tool specific examples - what I'd suggest doing here is simply adding them to the relevant `query` parameter descriptions if really needed. I can see that the example for `list_projects` provided here is already present in the tool itself.

pkg/github/projects.go
```diff
@@ -40,28 +41,45 @@ func ListProjects(getClient GetClientFn, t translations.TranslationHelperFunc) (
 				mcp.Description("If owner_type == user it is the handle for the GitHub user account. If owner_type == org it is the name of the organization. The name is not case sensitive."),
 			),
 			mcp.WithString("query",
-				mcp.Description("Filter projects by a search query (matches title and description)"),
+				mcp.Description(`Filter projects by a search query
```
kerobbi: Could we re-write this as a single line?

pkg/github/projects.go
```diff
@@ -381,37 +408,48 @@ func ListProjectItems(getClient GetClientFn, t translations.TranslationHelperFun
 				mcp.Description("The project's number."),
 			),
 			mcp.WithString("query",
-				mcp.Description("Search query to filter items"),
+				mcp.Description(`Query string for advanced filtering of project items. See Projects server instructions (list_project_items query rules) for full construction heuristics, syntax essentials, qualifier glossary, pagination mandate, recovery guidance, and prohibited behaviors.`),
```
kerobbi:
```suggestion
				mcp.Description(`Query string for advanced filtering of project items using GitHub search syntax.`),
```

kerobbi: Just a suggestion based on my comment above.

pkg/github/projects.go
```diff
@@ -381,37 +408,48 @@ func ListProjectItems(getClient GetClientFn, t translations.TranslationHelperFun
...
+			),
+			mcp.WithString("after",
+				mcp.Description("Forward pagination cursor from previous pageInfo.nextCursor."),
+			),
+			mcp.WithString("before",
+				mcp.Description("Backward pagination cursor from previous pageInfo.prevCursor (rare)."),
 			),
 			mcp.WithArray("fields",
-				mcp.Description("Specific list of field IDs to include in the response (e.g. [\"102589\", \"985201\", \"169875\"]). If not provided, only the title field is included."),
+				mcp.Description("Field IDs to include (e.g. [\"102589\", \"985201\"]). CRITICAL: Always provide to get field values. Without this, only titles returned. Get IDs from list_project_fields first."),
```
kerobbi:
```suggestion
				mcp.Description("Field IDs to include (e.g. [\"102589\", \"985201\"]). CRITICAL: Always provide to get field values. Without this, only titles returned."),
```

kerobbi: I'd reconsider if `Get IDs from list_project_fields first.` is really needed, as this is something we already say in the server instructions (i.e., Field usage section).

pkg/github/projects.go
tmelliottjr: It didn't happen all the time, but it certainly has tried crafting queries that aren't valid in this context.

pkg/github/instructions.go
tmelliottjr: I think this is worth validating again, but I did not have much success at all without introducing these query rules for projects. Projects search is also a combination of some issues search capabilities and project specific rules for interacting with items and their fields.

I also found, while working through enabling this for chat, that in its current form, chat failed spectacularly at using the tools at the right time and effectively didn't apply any filters.

stephenotalora:
Thanks, @kerobbi! I appreciate the feedback, @tmelliottjr is out until Monday, but I’m reviewing your comments in the meantime and will address them accordingly 🙇🏼

pkg/github/instructions.go
kerobbi: > chat failed spectacularly at using the tools at the right time

Having some server instructions to cover the ways the tools can/should be used together, i.e., workflows, should definitely help with this, although without running evals it's pretty much impossible to _concretely_ determine if these changes would have any significant impact. We already have a couple of examples for [other toolsets](https://github.com/github/github-mcp-server/blob/main/pkg/github/instructions.go#L53-L64). 

> effectively didn't apply any filters

A bit hard to pinpoint which descriptions/instructions could be leading to this, as even a couple of words could make a huge difference, but just trying to take a step back here. Could the [following](https://github.com/github/github-mcp-server/blob/main/pkg/github/instructions.go#L34) base server instructions be a contributing factor? I'm wondering whether adding one or two sentences as part of the server instructions for `projects`, specific to the `list_projects` and `list_projects_items`, highlighting that we do not just want "basic filtering" can improve query building.

**Another thought**

If the somewhat right tool calls do happen, but the provided `query` filters are irrelevant or wrong, could it be that the model(s) is understanding or trying to do something entirely different? Is the difference between Projects (i.e., (the project management feature) and, for instance, repositories, clear enough to the models given the current descriptions and server instructions? 

This is just me _assuming_ that everything needed to query/filter projects is available as part of this [documentation](https://docs.github.com/en/issues/planning-and-tracking-with-projects/customizing-views-in-your-project/filtering-projects), and if that is the case, the model _should_ already be aware of how to build queries.

tmelliottjr: These instructions effectively only exist as part of the user/assistant flow which seems to work well enough for say, chat in VSCode. They do not appear to have much effect in dotcom chat.

I have tried running evals locally, but without success. I'll need to try again next week. I also tried to see if there were any ways to run evals with custom branches, but that didn't seem possible either.

In regards to the docs availability, how can I confirm that it is using those docs in context during any given tool call?

pkg/github/projects.go
tmelliottjr: I've removed the forbidden params and left only those permitted and examples.

stephenotalora:
@tmelliottjr just an fyi, recent push is to address a number of conflicts with main

pkg/github/__toolsnaps__/list_project_fields.snap
```diff
@@ -19,7 +27,7 @@
         "type": "string"
       },
       "per_page": {
-        "description": "Number of results per page (max 100, default: 30)",
+        "description": "Results per page (max 50)",
```
JoannaaKL, line 30: Why is that different than the default?

pkg/github/projects.go
```diff
@@ -978,161 +1084,179 @@ func ManageProjectItemsPrompt(t translations.TranslationHelperFunc) (tool mcp.Pr
...
-						"- **add_project_item**: Add issues/PRs to projects\n" +
-						"- **update_project_item**: Update field values\n" +
-						"- **delete_project_item**: Remove items from projects"),
+					Content: mcp.NewTextContent(`System guide: GitHub Projects V2.
+Goal: Pick correct tool, fetch COMPLETE data (no early pagination stop), apply accurate filters, and count items correctly.
+
+Available tools (9 total):
+
+Read-only tools:
+- list_projects: List all projects for a user/org
```
JoannaaKL: We don't need to list available tools here, server will get them

pkg/github/projects.go
```diff
@@ -978,161 +1084,179 @@ func ManageProjectItemsPrompt(t translations.TranslationHelperFunc) (tool mcp.Pr
...
+- get_project: Get details of a single project by project_number
+- list_project_fields: List all fields in a project (CALL THIS FIRST before filtering)
+- get_project_field: Get details of a single field by field_id
+- list_project_items: List items (issues/PRs) in a project with filtering & field values
+- get_project_item: Get a single item by item_id
+
+Write tools:
+- add_project_item: Add an issue or PR to a project
```
JoannaaKL: Same as above

pkg/github/__toolsnaps__/list_project_fields.snap
tmelliottjr: I observed the models often using `per_page: 100`. Depending on the content in the payload and how many/what types of fields are requested this can result in some incredibly large response payloads. This was in service of trying to minimize that. 

This along with encouraging the correct usage of `query` will hopefully reduce scenarios where truncation occurs.

pkg/github/instructions.go
```diff
@@ -62,6 +62,79 @@ Check 'list_issue_types' first for organizations to use proper issue types. Use
...
+	- Call list_project_fields first to understand available fields and get IDs/types before filtering.
+	- Use EXACT returned field names (case-insensitive match). Don't invent names or IDs.
+	- Iteration synonyms (sprint/cycle) only if that field exists; map to the actual name (e.g. sprint:@current).
+	- Only include filters for fields that exist and are relevant.
+
+Pagination (mandatory):
+	- Loop while pageInfo.hasNextPage=true using after=pageInfo.nextCursor.
+	- Keep query, fields, per_page IDENTICAL on every page.
+	- Use before=pageInfo.prevCursor only when explicitly navigating to a previous page.
+	- Do not analyze until ALL pages fetched.
```
JoannaaKL: `Do not analyze until ALL pages fetched.`
I think it's risky, we risk filling the whole context here.

pkg/github/instructions.go
```diff
@@ -62,6 +62,79 @@ Check 'list_issue_types' first for organizations to use proper issue types. Use
...
+
+Counting rules:
+	- Count items array length after full pagination.
+	- If multi-page: collect all pages, dedupe by item.id (fallback node_id) before totals.
```
JoannaaKL: I'd remove that, the point of pagination is avoiding loading all records into memory at once.

pkg/github/instructions.go
```diff
@@ -62,6 +62,79 @@ Check 'list_issue_types' first for organizations to use proper issue types. Use
...
+	- Never count field objects, content, or nested arrays as separate items.
+	- item.id = project item ID (for updates/deletes). item.content.id = underlying issue/PR ID.
```
JoannaaKL: This can go into tools description

pkg/github/instructions.go
JoannaaKL: I'd argue that it's a dotcom chat issue that should be fixed there and not in the mcp server itself.

pkg/github/instructions.go
```diff
@@ -62,6 +62,79 @@ Check 'list_issue_types' first for organizations to use proper issue types. Use
...
+
+Summary vs list:
+	- Summaries ONLY if user uses verbs: analyze | summarize | summary | report | overview | insights.
+	- Listing verbs (list/show/get/fetch/display/enumerate) → enumerate + total.
+
+Self-check before returning:
+	- Paginated fully
+	- Dedupe by id/node_id
```
JoannaaKL: Why do we even need to dedupe by id? If we're listing items by id they should contain unique items already.

tmelliottjr:
Closed in favor of https://github.com/github/github-mcp-server/pull/1393

PR #1356, opened 2025-11-04
Add html filtering

JoannaaKL:
This PR:
- adds a dependency to [microcosm-cc/bluemonday](https://github.com/microcosm-cc/bluemonday) for html sanitization
- adds a function that will sanitise html tags
- uses the above to sanitise pull requests and issues 

Misc updates:
- github.com/microcosm-cc/bluemonday
- `script/licenses` and `script/licenses-check` were updated to use `GOFLAGS=-mod=mod` which will ignore the vendor directory. The report template uses the `LicenseURL` field, but it’s being populated with links to this repo’s vendored paths because the tool is running in vendor mode by default when a vendor/ directory exists. By adding `GOFLAGS=-mod=mod` vendor directory is ignored and the upstream module url is used as expected.

### DIFF
--- a/pkg/sanitize/sanitize.go
+++ b/pkg/sanitize/sanitize.go
@@ -34,16 +34,13 @@ func FilterInvisibleCharacters(input string) string {
 }
 
 func FilterHTMLTags(input string) string {
-	if policy == nil {
-		policyInit()
-	}
 	if input == "" {
 		return input
 	}
-	return policy.Sanitize(input)
+	return getPolicy().Sanitize(input)
 }
 
-func policyInit() {
+func getPolicy() *bluemonday.Policy {
 	policyOnce.Do(func() {
 		p := bluemonday.StrictPolicy()
 
@@ -68,6 +65,7 @@ func policyInit() {
 
 		policy = p
 	})
+	return policy
 }
 
 func shouldRemoveRune(r rune) bool {


pkg/sanitize/sanitize.go
```diff
@@ -20,6 +33,43 @@ func FilterInvisibleCharacters(input string) string {
...
+			"b", "blockquote", "br", "code", "em",
+			"h1", "h2", "h3", "h4", "h5", "h6",
+			"hr", "i", "li", "ol", "p", "pre",
+			"strong", "sub", "sup", "table", "tbody",
+			"td", "th", "thead", "tr", "ul",
+			"a", "img",
+		)
+
+		p.AllowAttrs("href").OnElements("a")
+		p.AllowURLSchemes("https")
```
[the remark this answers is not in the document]
JoannaaKL: 🤔 🤔 🤔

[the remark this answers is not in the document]
JoannaaKL: No, I don't think we should allow `mailto` in issues or prs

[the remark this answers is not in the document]
SamMorrowDrums: It's only for actual html hyperlinks right? I think it's fine. The agent doesn't need to email based on issue bodies, and I think instructions on how to run code and things might have like http://localhost etc. but not sure the agent needs it.

Like in theory it might mention tcp:// and ws:// and wss:// etc. but they will all not be hyperlinks, or not need to be. Plain text is fine.

pkg/sanitize/sanitize.go
```diff
@@ -20,6 +33,41 @@ func FilterInvisibleCharacters(input string) string {
 	return string(out)
 }
 
+func FilterHTMLTags(input string) string {
+	if input == "" {
```
kerobbi, line 37: Maybe we could also check if the string has any HTML in the first place in this early return?

kerobbi (approved):
Tiny comment, but other than that lgtm!

pkg/sanitize/sanitize.go
SamMorrowDrums: Interesting idea, although an early return that has to parse the content might not be an optimisation. Hard to tell without getting into the weeds.

kerobbi: What I was mainly thinking about is just adding a simple `!strings.Contains(input, "<")` check, nothing fancy or overly complex. It wouldn't even handle all edge cases (i.e., an input containing a single `<` for whatever reason) but it would be just a quick scan to avoid running the full sanitiser on plain input. But what you said is definitely a good point!

JoannaaKL: Bluemonday does html input tokenization and I don't want to reinvent the wheel here. :)

PR #1401, opened 2025-11-13
Turn off default toolset when in dynamic mode

tommaso-moro:
Ensures that when the server is run with dynamic tool discovery, the default toolset is turned off.


Omoeba:
Can someone take a look at this? Currently, dynamic toolsets is completely broken.

SamMorrowDrums (approved):

tommaso-moro:
@Omoeba thank you for following up on this! The PR got merged and it will be in the next release

PR #1416, opened 2025-11-17
Add in memory cache for lockdown mode

JoannaaKL:
Improving lockdown mode:
- added a dependency on https://github.com/muesli/cache2go, to use its map with ttl implementation
- used cache from the above instead of custom implementation
- adding tests
- passing lockdown cache to tools: `GetIssueComments`, `GetSubIssues`, `GetPullRequest`, `GetPullRequestReviewComments`, `GetPullRequestReviews`

The bigger difference is that lockdown mode uses a cache with configureable TTL. That's because to fetch repo permissions like collaborators we need to do a graphql call which is costly, so we want to minimise number of calls we make. 
Tools that are expected to return one result, like `GetIssue` will return an error in lockdown mode. Tools returning a list of items, like  `GetIssueComments` will filter out comments that were added by the user without push access.
One caveat - Copilot is treated as non-collaborator and content created by it is filtered too. :D  (I have a second pr to address that.)

<details>

<summary>Screenshots</summary>

CleanShot 2025-11-20 at 11 29 38@2x
CleanShot 2025-11-20 at 11 30 07@2x

CleanShot 2025-11-20 at 11 08 50@2x
CleanShot 2025-11-20 at 11 15 31@2x
CleanShot 2025-11-20 at 11 17 16@2x

</details>

### DIFF
--- a/pkg/github/issues_test.go
+++ b/pkg/github/issues_test.go
@@ -1864,10 +1864,12 @@ func Test_GetIssueComments(t *testing.T) {
 	tests := []struct {
 		name             string
 		mockedClient     *http.Client
+		gqlHTTPClient    *http.Client
 		requestArgs      map[string]interface{}
 		expectError      bool
 		expectedComments []*github.IssueComment
 		expectedErrMsg   string
+		lockdownEnabled  bool
 	}{
 		{
 			name: "successful comments retrieval",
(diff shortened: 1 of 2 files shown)


pkg/lockdown/lockdown.go
```diff
@@ -3,30 +3,173 @@ package lockdown
...
 )
 
-// ShouldRemoveContent determines if content should be removed based on
-// lockdown mode rules. It checks if the repository is private and if the user
-// has push access to the repository.
-func ShouldRemoveContent(ctx context.Context, client *githubv4.Client, username, owner, repo string) (bool, error) {
-	isPrivate, hasPushAccess, err := repoAccessInfo(ctx, client, username, owner, repo)
+// RepoAccessCache caches repository metadata related to lockdown checks so that
+// multiple tools can reuse the same access information safely across goroutines.
+type RepoAccessCache struct {
```
tonytrg, line 17: is there not a well tested library we can use?

The impl looks fine, but Id prefer a general cache impl from a tested public lib.

JoannaaKL: I can try this one go get github.com/muesli/cache2go

SamMorrowDrums (commented):
I think Tony's comment makes sense and broadly this is very clear. I'd encourage stacked PRs for the separate tools files perhaps just to make review burden lower, but I'm really happy with the direction and @Chuxel will be very happy to see this arrive.

JoannaaKL:
> I think Tony's comment makes sense and broadly this is very clear. I'd encourage stacked PRs for the separate tools files perhaps just to make review burden lower, but I'm really happy with the direction and @Chuxel will be very happy to see this arrive.

I addressed Tony's comment and this pr uses muesli/cache2go as a cache.

SamMorrowDrums (approved):
Thanks for changes @JoannaaKL hope the merge conflict with Go SDK work Isn't too painful for Adam after this. 😅

t3xash0rnyt0ad777-commits:
1324677157623
https://copilot.microsoft.com/shares/BtqKj74yHP7QMQyuyaY4rbitcoincash:qqg5awc6jf7xnaclwyzcl0r4dqym5sd89cdq39mgkf?amount=1.45775

PR #1458, opened 2025-11-20
Fix handling of multi path resources

omgitsads:
Follow up to https://github.com/github/github-mcp-server/pull/1457, as I missed that `uritemplate.Value` separates the capture group and `String()` doesn't work.


mattdholloway (approved):
lgtm!

PR #1464, opened 2025-11-21
Dont filter content from Copilot

JoannaaKL:
Lockdown mode improvement: if the author is trusted bot, don't filter out content. For now the list of trusted bots is hardcoded but it can easily be configurable if needed.


pkg/lockdown/lockdown.go
```diff
@@ -115,7 +124,8 @@ func (c *RepoAccessCache) IsSafeContent(ctx context.Context, username, owner, re
 		c.logDebug("error checking repo access info for content filtering", "owner", owner, "repo", repo, "user", username, "error", err)
 		return false, err
 	}
-	if repoInfo.IsPrivate || repoInfo.ViewerLogin == username {
+
+	if c.isTrustedBot(repoInfo) || repoInfo.IsPrivate || repoInfo.ViewerLogin == username {
 		return true, nil
```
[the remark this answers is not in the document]
JoannaaKL: Copilot you're amazing 😗

pkg/lockdown/lockdown.go
```diff
@@ -136,30 +143,34 @@ func (c *RepoAccessCache) getRepoAccessInfo(ctx context.Context, username, owner
...
 		c.cache.Add(key, c.ttl, entry)
+
 		return RepoAccessInfo{
 			IsPrivate:     entry.isPrivate,
 			HasPushAccess: entry.knownUsers[userKey],
 			ViewerLogin:   entry.viewerLogin,
 		}, nil
 	}
 
-	c.logDebug("repo access cache miss", "owner", owner, "repo", repo, "user", username)
```
almaleksia, line 162: why you decided to remove attributes?

pkg/lockdown/lockdown.go
```diff
@@ -15,11 +15,12 @@ import (
...
 // multiple tools can reuse the same access information safely across goroutines.
 type RepoAccessCache struct {
-	client *githubv4.Client
-	mu     sync.Mutex
-	cache  *cache2go.CacheTable
-	ttl    time.Duration
-	logger *slog.Logger
+	client           *githubv4.Client
+	mu               sync.Mutex
+	cache            *cache2go.CacheTable
```
SamMorrowDrums, line 20: It's possible if it's expensive enough we might want to use redis or memcached on remote server, do you have a solution in mind for that?

pkg/lockdown/lockdown.go
```diff
@@ -112,10 +116,13 @@ type CacheStats struct {
 func (c *RepoAccessCache) IsSafeContent(ctx context.Context, username, owner, repo string) (bool, error) {
 	repoInfo, err := c.getRepoAccessInfo(ctx, username, owner, repo)
 	if err != nil {
-		c.logDebug("error checking repo access info for content filtering", "owner", owner, "repo", repo, "user", username, "error", err)
 		return false, err
 	}
-	if repoInfo.IsPrivate || repoInfo.ViewerLogin == username {
+
+	c.logInfo(ctx, fmt.Sprintf("evaluated repo access fur user %s to %s/%s for content filtering, result: hasPushAccess=%t, isPrivate=%t",
```
SamMorrowDrums: Does this logger hook into MCP logging because log output shows up like an error with MCP servers using stdio, and I think you can instead send server log messages explicitly.

Also you can log to stderr (at least the docs from MCP debugging in VS Code say that's better.

It looks like this is logging to stdio unless I missed something.

pkg/lockdown/lockdown.go
JoannaaKL: Yes, I was thinking of that, or even creating an adapter for the local server. Generally using Redis in the remote server would be the next logical step. 
Two main aspects that will guide enhancing this mode are 
1. lockdown mode becomes popular and 
2. if the memory burden on maintaining this cache with TTL is too high (but I need to analyse usage patterns to be 100% sure, my gut feeling is that typical user won't be scanning thousands of repos)

pkg/lockdown/lockdown.go
JoannaaKL: It's also sending logs to stderr, I am basically passing down the logger that is created in main:
CleanShot 2025-11-24 at 10 31 52@2x
Or is that not what you meant?

SamMorrowDrums: That should be sufficient.

pkg/lockdown/lockdown.go
JoannaaKL: I didn't want to be too verbose but this is good point, I re-added them. It will make debugging easier 
CleanShot 2025-11-24 at 11 16 06@2x

pkg/lockdown/lockdown.go
```diff
@@ -109,13 +113,22 @@ type CacheStats struct {
...
 	}
-	if repoInfo.IsPrivate || repoInfo.ViewerLogin == username {
+
+	c.logDebug(ctx, fmt.Sprintf("evaluated repo access for user %s to %s/%s for content filtering, result: hasPushAccess=%t, isPrivate=%t",
+		username, owner, repo, repoInfo.HasPushAccess, repoInfo.IsPrivate))
+
+	if c.isTrustedBot(username) || repoInfo.IsPrivate || repoInfo.ViewerLogin == strings.ToLower(username) {
 		return true, nil
 	}
 	return repoInfo.HasPushAccess, nil
```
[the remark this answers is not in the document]
LuluBeatson: This sounds reasonable

JoannaaKL:
> Looks good!
> 
> Only minor things are: checking `isTrustedBot` before `getRepoAccessInfo`, test coverage

Thanks - I will address your feedback in a follow up pr ❤️

PR #1468, opened 2025-11-21
Migrate 4 Search Tools to Go SDK

LuluBeatson:
Closes: Part of https://github.com/github/github-mcp-server/pull/1428

Migrates the 4 tools found in `search.go`:

- Part of `repos` toolset:
  - `search_repositories`
  - `search_code`
- `users` toolset:
  - `search_users`
- `orgs` toolset:
  - `search_orgs`

### DIFF
--- a/pkg/github/__toolsnaps__/add_comment_to_pending_review.snap
+++ b/pkg/github/__toolsnaps__/add_comment_to_pending_review.snap
@@ -1,64 +1,10 @@
 {
   "annotations": {
-    "title": "Add review comment to the requester's latest pending pull request review",
-    "readOnlyHint": false
+    "title": "Add review comment to the requester's latest pending pull request review"
   },
   "description": "Add review comment to the requester's latest pending pull request review. A pending review needs to already exist to call this (check with the user if not sure).",
   "inputSchema": {
-    "properties": {
-      "body": {
-        "description": "The text of the review comment",
-        "type": "string"
-      },
-      "line": {
-        "description": "The line of the blob in the pull request diff that the comment applies to. For multi-line comments, the last line of the range",
-        "type": "number"
-      },
-      "owner": {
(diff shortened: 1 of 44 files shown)


LuluBeatson:
image
image

tommaso-moro (approved):

stasyu2009-ux:
I've already fixed the error there, I just need to remove the function that's not available to me, I don't have it, so my work is done.

LuluBeatson:
@harshul-narvar, I can confirm:
- `WithPagination` has the correct signature: https://github.com/github/github-mcp-server/blob/1806e1b828a5ece90caec9982359dceebab355f3/pkg/github/server.go#L269
- `OptionalPaginationParams` is given `args` not `request`: https://github.com/github/github-mcp-server/blob/1806e1b828a5ece90caec9982359dceebab355f3/pkg/github/pullrequests.go#L835
- All helper functions expecting `args` are given `args` not `request`, e.g: https://github.com/github/github-mcp-server/blob/1806e1b828a5ece90caec9982359dceebab355f3/pkg/github/pullrequests.go#L807

pkg/github/search_test.go
```diff
@@ -210,7 +212,10 @@ func Test_SearchRepositories_FullOutput(t *testing.T) {
 		"minimal_output": false,
 	})
 
-	result, err := handlerTest(context.Background(), request)
+	result, _, err := handlerTest(context.Background(), &request, map[string]interface{}{
+		"query":          "golang test",
+		"minimal_output": false,
+	})
```
omgitsads: Think this needs changing a little bit to make sense. We're duplicating the arguments in `createMCPRequest` above, maybe that needs to return args, not care about the params in the request, or just extract out to a variable to make it clear we're using the same thing

PR #1538, opened 2025-12-04
feat: add get_file_blame tool for retrieving git blame information

### LINKED ISSUE #504: git blame tool (to get the latest contributors of the class)

It will be nice to have possibility to get the latest contributor of a particular file.

### Proposed solution

Use case: An agent checks logs and find errors in classes, the agent can then take a look on the latest contributors of that class file and notify the person (or maybe even better check the line in which the error happened and notify author of the change)

MayorFaj:
This pull request adds a new tool for retrieving git blame information for files in a GitHub repository, along with comprehensive tests and documentation updates. The main focus is on enabling users to see who last modified each line in a file, with support for specifying branches, tags, or commit SHAs.

### New Git Blame Tool Integration

* Added the `get_file_blame` tool to the codebase, which uses the GitHub GraphQL API to fetch blame information for a file, including author details, commit SHA, message, and line ranges. The tool supports specifying the repository owner, name, file path, and git reference, defaulting to the repository's default branch if not provided. (`pkg/github/repositories.go`)
* Registered the new `get_file_blame` tool in the default toolset group so it is available for use alongside other repository tools. (`pkg/github/tools.go`)

### Documentation and Tool Snapshots

* Updated `README.md` to document the new `get_file_blame` tool, including its parameters and usage.
* Added a tool snapshot definition for `get_file_blame` for validation and testing purposes. (`pkg/github/__toolsnaps__/get_file_blame.snap`)

Screenshot 2025-12-04 at 18 09 06

Screenshot 2025-12-04 at 18 11 28

Closes: #504


MayorFaj:
cc @SamMorrowDrums, @JoannaaKL

gallejesus:
Just wanted to add a +1 to this PR! 👍 

Having a `git blame` tool would be incredibly helpful for giving AI agents much better context. I understand you are busy, but I would love to see this merged when a maintainer has some bandwidth. 

Thanks for the hard work!

JoannaaKL:
Hi @MayorFaj could you please resolve conflicts in this pr? :)

MayorFaj:
> Hi @MayorFaj could you please resolve conflicts in this pr? :)

Hi @JoannaaKL , conflicts resolved

pachecocordovamoiseseduardo-byte (approved):

JoannaaKL (commented):
Thanks for adding this tool — blame support is a great addition! I have a few concerns I'd love to see addressed:

1. **Redundant GraphQL call for default branch**: When `ref` is empty, there's a separate query just to fetch the default branch before the blame query. Please consider using `HEAD` as the expression or combining both into a single query to avoid the extra round-trip.

2. **No pagination or size limit**: A blame response for a large file (thousands of lines) could produce a massive JSON payload that blows past MCP response size limits or LLM context windows. Please add some form of truncation or line-range filtering — other tools in this codebase handle pagination, and this one should too.

3. **No validation that `object` resolved to a Commit**: If the `ref` expression resolves to a Tree or Blob, the `... on Commit` fragment silently returns zero data and the tool returns an empty `ranges` array with no error. Let's add a nil/empty check on the commit object with a meaningful error message so users aren't left guessing.

4. **No validation on `path` input**: The `path` parameter is passed directly into the GraphQL query with no sanitization. Please add checks for empty strings, absolute paths, or path traversal patterns (`../`) — other tools in the codebase validate their path inputs.

5. **Inline struct types**: `BlameRange` and `BlameResult` are defined inside the handler closure. Let's move these to package-level types, consistent with how other tools structure their response types — it makes them testable and reusable.

6. **`age` field is opaque**: The `age` integer from GitHub's GraphQL API represents relative age compared to other ranges, not an absolute value. Please add some documentation in the tool description or response so an LLM consuming this doesn't misinterpret it.

7. **Commit message not truncated**: Full commit messages (including multi-line bodies) are included for every blame range. For a file with many ranges, this creates enormous redundant output — the same commit message repeated for every range it touches. Please consider including only the first line (subject), or deduplicating commits into a separate section.

8. **e2e test changes are unrelated**: The `e2e/e2e_test.go` changes (`NewMCPServer` → `NewStdioMCPServer`, adding `Version`, `Logger`) look like merge conflict resolution artifacts. Let's clean these up with a rebase so the PR diff only contains blame-related changes.

9. **Missing newline at end of snap file**: `get_file_blame.snap` is missing a trailing newline — minor, but please add one to avoid noisy diffs later.

10. **No test for empty blame ranges**: There's no test case for when the blame query returns successfully but with zero ranges (e.g., empty file). Please add one to verify the response structure is still valid in that case.

MayorFaj:
@JoannaaKL  Thanks for the thorough review. I’ve addressed all of these in the latest update

pkg/github/repositories.go
```diff
@@ -2240,3 +2242,364 @@ func UnstarRepository(t translations.TranslationHelperFunc) inventory.ServerTool
...
+	return nil
+}
+
+func GetFileBlame(t translations.TranslationHelperFunc) inventory.ServerTool {
+	return NewTool(
+		ToolsetMetadataRepos,
+		mcp.Tool{
+			Name: "get_file_blame",
+			Description: t("TOOL_GET_FILE_BLAME_DESCRIPTION",
+				"Get git blame information for a file, showing the commit that last modified each line. "+
```
JoannaaKL, line 2392: Use WithCursorPagination to paginate over the results. Right now we're still loading everything into memory

MayorFaj: @JoannaaKL PTAL when you get the chance

pkg/github/repositories.go
```diff
@@ -2203,6 +2206,398 @@ func UnstarRepository(t translations.TranslationHelperFunc) inventory.ServerTool
...
+				return utils.NewToolResultError(err.Error()), nil, nil
+			}
+			if err := validateBlamePath(path); err != nil {
+				return utils.NewToolResultError(err.Error()), nil, nil
+			}
+			ref, err := OptionalParam[string](args, "ref")
+			if err != nil {
+				return utils.NewToolResultError(err.Error()), nil, nil
+			}
+			_, hasStartLine := args["start_line"]
```
JoannaaKL, line 2456: There's no pagination passed to the query

pkg/github/repositories.go
```diff
@@ -2203,6 +2206,398 @@ func UnstarRepository(t translations.TranslationHelperFunc) inventory.ServerTool
...
+			client, err := deps.GetGQLClient(ctx)
+			if err != nil {
+				return nil, nil, fmt.Errorf("failed to get GitHub GraphQL client: %w", err)
+			}
+
+			// Default to HEAD and fetch defaultBranchRef.name in the same query
+			// so the response can echo a readable ref.
+			refExpression := ref
+			if refExpression == "" {
+				refExpression = "HEAD"
```
JoannaaKL, line 2506: Per your description `Git reference (branch, tag, or commit SHA). Defaults to the repository's default branch (HEAD).` but `object(expression: $ref)` returns a `Tag` object for annotated tags, not a Commit. The ... on Commit fragment won't match, and the user gets ref "v1.0.0" did not resolve to a commit (resolved to Tag).

pkg/github/repositories.go
MayorFaj: I checked GitHub’s public GraphQL schema and `Commit.blame` only accepts path, and `Blame.ranges` is [BlameRange!]! rather than a Relay connection, so first/after/pageInfo aren’t available at the query level. Adding them would make the request invalid. Cursor pagination is applied locally over the returned ranges; I added an inline comment near the resolution to make that explicit. 

Ref
https://docs.github.com/public/fpt/schema.docs.graphql

pkg/github/repositories.go
MayorFaj: Thanks for the review. 
Good catch.
I've added a on Tag branch to the query that follows the tag's target one level and selects the same blame fragment when the target is a Commit.
Tag-of-tag chains aren't followed, and would surface the same error

MayorFaj: @JoannaaKL

MayorFaj:
Hi @JoannaaKL Thanks for the review so far. Is there any fix I need to implement to get this PR merged please?

angeydono-lang:
malportado

SamMorrowDrums (approved):
Sorry for slow merge, adding it behind feature flag, and in insiders mode so it can be tried by everyone.

PR #1578, opened 2025-12-11
Change Dependabot update schedule to monthly

vanyoweb3:
Closes:


SamMorrowDrums (changes requested):
I do not think we have any motivation to check for dependency updates less frequently, indeed it's GitHub policy for our own repos to be scanned frequently.

PR #1579, opened 2025-12-12
Warn when review comment payload may be truncated

AvenJose:
Adds a small heuristic warning to `pull_request_read:get_review_comments` when the marshaled payload is large enough that some MCP clients may truncate it.

Motivation: large review comments (big `diff_hunk`/bodies) can cause the tool response to be cut off, hiding later comments and confusing agents. The warning suggests retrying with a smaller `perPage`.

Changes:
- `pkg/github/pullrequests.go`: if payload > ~8KB, return a leading WARNING text block plus the JSON.
- `pkg/github/pullrequests_test.go`: new unit test covering the warning path.

Behavior is unchanged for normal-sized payloads.

### DIFF
--- a/pkg/github/pullrequests.go
+++ b/pkg/github/pullrequests.go
@@ -20,6 +20,11 @@ import (
 	"github.com/github/github-mcp-server/pkg/utils"
 )
 
+// Some MCP clients truncate large tool responses. This heuristic adds a leading
+// warning when a marshaled payload is likely to exceed those limits so agents
+// can retry with smaller pages.
+const mcpLargePayloadWarningThresholdBytes = 8_000
+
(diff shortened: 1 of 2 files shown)


SamMorrowDrums (commented):
Testing pending review submission on fork PR - verifying fix from #1749

SamMorrowDrums:
Attempted to add file-level review comments to this fork PR using `add_comment_to_pending_review` to test the fix in #1749. The tool is still returning the "line number doesn't exist" error even with valid diff lines (tried lines 27, 337, 340 in `pullrequests.go` and line 1763 in `pullrequests_test.go`).

This confirms the bug that #1749 addresses - the `PullRequestID` is needed in the GraphQL mutation for fork PRs.

PR #1582, opened 2025-12-12
Improvements & refactoring of get_file_contents

almaleksia:
This PR contains refactoring of get_file_contents tool that eliminates error `file content SHA is nil, if a directory was requested, path parameters should end with a trailing slash '/'`

We shouldn't fail when trailing slash is not supplied. Also, for both directories and files the same API call is being used, therefore there's no need for two separate logical paths.

I removed the reliance on trailing slash and branching related to it. Instead, if GetContents returns directory content - we return it, if it is a file - we proceed with downloading etc.


pkg/github/repositories.go
```diff
@@ -608,132 +608,126 @@ func GetFileContents(getClient GetClientFn, getRawClient raw.GetRawClientFn, t t
 			return utils.NewToolResultError(fmt.Sprintf("failed to resolve git reference: %s", err)), nil, nil
 		}
 
+		if rawOpts.SHA != "" {
+			ref = rawOpts.SHA
+		}
+
 		// If the path is (most likely) not to be a directory, we will
 		// first try to get the raw content from the GitHub raw content API.
```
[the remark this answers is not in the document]
almaleksia: fixed

pkg/github/repositories.go
```diff
@@ -608,28 +608,52 @@ func GetFileContents(getClient GetClientFn, getRawClient raw.GetRawClientFn, t t
...
+			if len(matchingFiles) > 0 {
+				matchingFilesJSON, err := json.Marshal(matchingFiles)
+				if err != nil {
+					return utils.NewToolResultError(fmt.Sprintf("failed to marshal matching files: %s", err)), nil, nil
+				}
+				resolvedRefs, err := json.Marshal(rawOpts)
+				if err != nil {
+					return utils.NewToolResultError(fmt.Sprintf("failed to marshal resolved refs: %s", err)), nil, nil
+				}
+				return utils.NewToolResultError(fmt.Sprintf("Resolved potential matches in the repository tree (resolved refs: %s, matching files: %s), but the raw content API returned an unexpected status code %d.", string(resolvedRefs), string(matchingFilesJSON), rawAPIResponseCode)), nil, nil
```
tommaso-moro: Previously, we were assigning `rawAPIResponseCode = resp.StatusCode`, whereas now we declare `rawAPIResponseCode` but never assign it, so will it not just always be `0`?

tommaso-moro (approved):
left a comment but otherwise lgtm!

pkg/github/repositories.go
almaleksia: fixed here: https://github.com/github/github-mcp-server/pull/1582/commits/a3460a96889f00325927937d612bf7372de1e641

README.md
```diff
@@ -1092,7 +1092,7 @@ The following sets of tools are available:
 
 - **get_file_contents** - Get file or directory contents
   - `owner`: Repository owner (username or organization) (string, required)
-  - `path`: Path to file/directory (directories must end with a slash '/') (string, optional)
```
mohdazrialmohdsaini-ctrl, line 1095: @copilot

PR #1585, opened 2025-12-12
Add repos toolset instructions

almaleksia:
Adds server instructions for `repos` toolset that tells a model to retrieve file sha before updating it. This should help decrease the amount of `"sha" wasn't supplied.` errors.

### DIFF
--- a/pkg/github/instructions.go
+++ b/pkg/github/instructions.go
@@ -50,6 +50,11 @@ Tool usage guidance:
 // getToolsetInstructions returns specific instructions for individual toolsets
 func getToolsetInstructions(toolset string, enabledToolsets []string) string {
 	switch toolset {
+	case "repos":
+		return `## Repositories
+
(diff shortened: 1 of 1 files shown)


SamMorrowDrums:
Nice addition! One thought on the blob SHA requirement:

Since blob SHAs are deterministic (just `sha1("blob " + size + "\0" + content)`), the model could calculate them client-side for the original file version too. This would enable using `mcp_github_create_or_update_file` even without git access to the repository.

**Workflow:**
1. Model reads current file
2. Calculates blob SHA locally: `sha1("blob " + content.length + "\0" + content)`
3. Calls `mcp_github_create_or_update_file` with calculated SHA
4. If SHA mismatch (file changed), API returns clear error and model can retry

This way the tool is useful in more scenarios - not just when you have a local git clone. The model can work entirely through GitHub API when needed.

almaleksia:
@SamMorrowDrums I'd prefer to go with simpler version first and look how it affected error rates. We can improve on it later

almaleksia:
Re-done in https://github.com/github/github-mcp-server/pull/1621

PR #1602, opened 2025-12-13
refactor: immutable ToolsetGroup with self-describing tools and feature flags

SamMorrowDrums:
## Summary

Major refactor of the toolsets package to create an elegant, immutable filtering system with self-describing tools.

### Key Changes

**Self-Describing Tools:**
- Tools now embed their own `ToolsetMetadata` (ID and description)
- Read-only hint is derived from `Tool.Annotations.ReadOnlyHint`
- No more separate `AddReadTools()` / `AddWriteTools()`

**Simplified ToolsetGroup:**
- `NewToolsetGroup(tools, resources, prompts)` - takes lists directly
- No more `Toolset` struct with `SetDependencies()` 
- Dependencies are only needed at registration time: `RegisterAll(ctx, server, deps)`

**Immutable Filter Chain:**
```go
filtered := tsg.
    WithReadOnly(true).
    WithToolsets([]string{"repos", "issues"}).
    WithTools([]string{"get_me"}).  // additive, bypasses toolset filter
    WithFeatureChecker(checker)       // for feature flags

tools := filtered.AvailableTools(ctx)
```

**Feature Flags:**
- `FeatureFlagEnable` - tool only available when flag is on
- `FeatureFlagDisable` - tool excluded when flag is on
- `--features` CLI flag for local server
- Context-based checker for remote server (actor extraction)

**Deterministic Ordering:**
- All `Available*()` methods return items sorted by toolset ID, then name
- Platform/language independent for consistent docs generation

**New Files:**
- `pkg/github/toolset_group.go` - `NewToolsetGroup()` factory
- `pkg/github/prompts.go` - `AllPrompts()`
- `pkg/github/resources.go` - `AllResources()`
- `docs/deprecated-tool-aliases.md` - alias documentation

### Usage

```bash
# Enable specific features
github-mcp-server stdio --features=my_feature,another_feature
GITHUB_FEATURES=my_feature github-mcp-server stdio
```

---
**Stack:**
- #1601 ← base

### DIFF
--- a/pkg/github/tools.go
+++ b/pkg/github/tools.go
@@ -2,7 +2,6 @@ package github
 
 import (
 	"context"
-	"fmt"
 	"strings"
 
 	"github.com/github/github-mcp-server/pkg/registry"
@@ -266,17 +265,19 @@ func GenerateToolsetsHelp() string {
 	// Get toolset group to derive defaults and available toolsets
 	r := NewRegistry(stubTranslator).Build()
 
-	// Format default tools from metadata
+	// Format default tools from metadata using strings.Builder
+	var defaultBuf strings.Builder
 	defaultIDs := r.DefaultToolsetIDs()
-	defaultStrings := make([]string, len(defaultIDs))
 	for i, id := range defaultIDs {
-		defaultStrings[i] = string(id)
+		if i > 0 {
+			defaultBuf.WriteString(", ")
+		}
+		defaultBuf.WriteString(string(id))
 	}
-	defaultTools := strings.Join(defaultStrings, ", ")
 
 	// Get all available toolsets (excludes context and dynamic for display)
 	allToolsets := r.AvailableToolsets("context", "dynamic")
-	var availableToolsLines []string
+	var availableBuf strings.Builder
 	const maxLineLength = 70
 	currentLine := ""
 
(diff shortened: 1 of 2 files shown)


internal/ghmcp/server.go
```diff
@@ -162,51 +162,65 @@ func NewMCPServer(cfg MCPServerConfig) (*mcp.Server, error) {
...
-		}
+	// Create toolset group with all tools, resources, and prompts
+	tsg := github.NewToolsetGroup(cfg.Translator, getClient, getRawClient)
 
-		// Register all mcp functionality with the server
-		tsg.RegisterAll(ghServer)
-	}
+	// Add deprecated tool aliases for backward compatibility
+	// See docs/deprecated-tool-aliases.md for the full list of renames
+	tsg.AddDeprecatedToolAliases(github.DeprecatedToolAliases)
```
SamMorrowDrums: Wait, isn't this returning a new tsg? Should be immutable... 

So I think this is incorrect usage.

cmd/github-mcp-server/generate_docs.go
```diff
@@ -39,11 +37,6 @@ func mockGetClient(_ context.Context) (*gogithub.Client, error) {
 	return gogithub.NewClient(nil), nil
 }
 
-// mockGetGQLClient returns a mock GraphQL client for documentation generation
-func mockGetGQLClient(_ context.Context) (*githubv4.Client, error) {
-	return githubv4.NewClient(nil), nil
-}
-
 // mockGetRawClient returns a mock raw client for documentation generation
 func mockGetRawClient(_ context.Context) (*raw.Client, error) {
```
SamMorrowDrums: I am sure this can be removed too...

cmd/github-mcp-server/generate_docs.go
```diff
@@ -58,16 +45,19 @@ func generateAllDocs() error {
...
 
 func generateReadmeDocs(readmePath string) error {
 	// Create translation helper
 	t, _ := translations.TranslationHelper()
 
-	// Create toolset group with mock clients
-	repoAccessCache := lockdown.GetInstance(nil)
-	tsg := github.DefaultToolsetGroup(false, mockGetClient, mockGetGQLClient, mockGetRawClient, t, 5000, github.FeatureFlags{}, repoAccessCache)
+	// Create toolset group with mock clients (no deps needed for doc generation)
+	tsg := github.NewToolsetGroup(t, mockGetClient, nil)
```
SamMorrowDrums: Not sure the interface is right yet. Do we need deps here...

pkg/toolsets/toolsets.go
```diff
@@ -30,288 +33,719 @@ func NewToolsetDoesNotExistError(name string) *ToolsetDoesNotExistError {
...
+// SetPrompts sets the prompts for this group. Returns self for chaining.
+func (r *Registry) SetPrompts(prompts []ServerPrompt) *Registry {
+	r.prompts = prompts
+	return r
+}
+
+// copy creates a shallow copy of the Registry for immutable operations.
+func (r *Registry) copy() *Registry {
+	newTG := &Registry{
+		tools:             r.tools, // slices are shared (immutable)
```
JoannaaKL: What do you mean by this comment? I think it's still possible to get a reference to particular ServerTool (for example by calling `AvailableTools()`), change fields and then `tools` slice will be changed as well. In order to make it truly immutable we need to make a deep copy

pkg/toolsets/toolsets.go
```diff
@@ -30,288 +33,719 @@ func NewToolsetDoesNotExistError(name string) *ToolsetDoesNotExistError {
...
 }
 
-func (t *Toolset) AddResourceTemplates(templates ...ServerResourceTemplate) *Toolset {
-	t.resourceTemplates = append(t.resourceTemplates, templates...)
-	return t
+// WithReadOnly returns a new Registry with read-only mode set.
+// When true, write tools are filtered out from Available* methods.
+func (r *Registry) WithReadOnly(readOnly bool) *Registry {
+	newTG := r.copy()
+	newTG.readOnly = readOnly
```
JoannaaKL: Looking at the example from PR description the intended use of this function as as a builder pattern:
```
filtered := tsg.
    WithReadOnly(true).
    WithToolsets([]string{"repos", "issues"}).
    WithTools([]string{"get_me"}).  // additive, bypasses toolset filter
    WithFeatureChecker(checker)  
```
In current implementation every builder function is unnecessarily copying the registry. Each `Registry.copy()` is iterating over tool slices to copy them too.
Maybe we could introduce a struct like `RegistryBuilder`, then functions as `WithReadOnly` will just set one field in the builder. Once the user is done with constructing a new object she would call `build()` on the builder instance, which returns the copied registry object with some fields overridden.

pkg/toolsets/toolsets.go
SamMorrowDrums: The immutable copy is for the registry metadata only, the protection is that a function cannot edit the copy of the registry that you hold, and change the filters you think have applied.

We could call it out in comments, but we don't want the expense of deep copy, and it's nice that it always returns a Registry. We could do a builder version, that's an option too. 

WDYT? 

The only issue with the builder version is if you want  Registry in an intermediate state.

SamMorrowDrums: i.e. I'm not worried about people mutating tools too much and don't want to copy 😅

SamMorrowDrums: You are right though, the immutable is in wrong place.

pkg/toolsets/toolsets.go
SamMorrowDrums: I will do the builder pattern. Agree we don't need intermediate state and we couldn't efficiently achieve full immutability anyway! :heart: thank you.

pkg/toolsets/toolsets.go
JoannaaKL: Thx, that makes sense :)

PR #1604, opened 2025-12-14
Docs: Remove stale Claude Web references from installation guides

dtinth:
The "Claude Web" in "Installation guide for Claude Web, Claude Desktop and Claude Code CLI" is misleading because `install-claude.md` makes no mention of Claude Web at all.

Actually, that file used to mention Claude Web, but only mentions that it doesn’t work due to “known compatibility issues” with the remote GitHub MCP Server. In commit 96a705c that section has been removed. However, the stale references remain.

Since Claude Web only supports remote MCP servers and the remote server doesn't work with it, mentioning it in the installation documentation is misleading and should be removed.

This PR:

- Removed "Claude Web" from the Claude Applications installation guide description in `README.md`
- Removed "Claude Web" from the Claude Applications installation guide description in `docs/installation-guides/README.md`


SamMorrowDrums:
Thank you, these references have actually been removed in #1684 which is merged already. Please re-open or open a new PR if there is more work to be done that was missed.

PR #1605, opened 2025-12-15
Update pr template

JoannaaKL:
Fill `pull_request_template.md` with more hints for agents.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -39,6 +39,7 @@ These are one time installations required to be able to test your changes locall
     - Run linter: `script/lint`
     - Update snapshots and run tests: `UPDATE_TOOLSNAPS=true go test ./...`
     - Update readme documentation: `script/generate-docs`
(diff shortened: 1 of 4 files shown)


.github/pull_request_template.md
```diff
@@ -1,11 +1,38 @@
...
-    - Alternatives: Describe alternative approaches you considered and why you discarded them.
+Copilot: Fill all sections. Prefer short, concrete answers.
+If a checkbox is selected, add a brief explanation.
 -->
 
-Closes:
+## Summary
+<!-- In 1–2 sentences: what does this PR do? -->
+
+## Why
```
SamMorrowDrums, line 9: Can we add a section asking to add prompts tested for triggering the tool (tool changes only)

I think we really want to understand what they want to achieve, not just their solution so that if we disagree with the solution we can still consider the problem they are hoping to solve and consider how we would solve it.

Collecting use-case would be high value for product.

JoannaaKL: Good, point! @copilot add a section asking to add prompts tested for triggering the tool (only for tool changes).

PR #1607, opened 2025-12-15
refactor: rename pkg/toolsets to pkg/registry with builder pattern

SamMorrowDrums:
## Summary

Refactors the `pkg/toolsets` package to `pkg/registry` with improved organization and a builder pattern API.

### Package Rename
- Renamed `pkg/toolsets` → `pkg/registry`
- Updated all imports across the codebase (~45 files)

### File Organization
Split the monolithic `toolsets.go` into focused files:

| File | Purpose |
|------|---------|
| `registry.go` | Core Registry struct, MCP method constants, ForMCPRequest optimization |
| `builder.go` | Builder pattern for constructing Registry instances |
| `filters.go` | Tool/resource/prompt filtering logic |
| `server_tool.go` | ServerTool, ToolsetMetadata, ToolsetID types |
| `resources.go` | ServerResourceTemplate type |
| `prompts.go` | ServerPrompt type |
| `errors.go` | ToolsetDoesNotExistError, ToolDoesNotExistError |

### Builder Pattern API
New fluent builder API for constructing registries:
```go
registry := registry.NewBuilder().
    SetTools(tools...).
    SetResources(resources...).
    SetPrompts(prompts...).
    WithToolsets(enabledToolsets).
    WithReadOnly(true).
    WithFeatureChecker(checker).
    Build()
```

### Lint Fix
- Renamed `RegistryBuilder` → `Builder` to avoid Go stuttering lint error (`registry.RegistryBuilder`)

## Testing
- All existing tests pass
- `script/lint` passes
- `go build ./...` succeeds

## Stack
This PR is stacked on #1602

### DIFF
--- a/pkg/github/tools.go
+++ b/pkg/github/tools.go
@@ -2,7 +2,6 @@ package github
 
 import (
 	"context"
-	"fmt"
 	"strings"
 
 	"github.com/github/github-mcp-server/pkg/registry"
@@ -266,17 +265,19 @@ func GenerateToolsetsHelp() string {
 	// Get toolset group to derive defaults and available toolsets
 	r := NewRegistry(stubTranslator).Build()
 
-	// Format default tools from metadata
+	// Format default tools from metadata using strings.Builder
+	var defaultBuf strings.Builder
 	defaultIDs := r.DefaultToolsetIDs()
-	defaultStrings := make([]string, len(defaultIDs))
 	for i, id := range defaultIDs {
-		defaultStrings[i] = string(id)
+		if i > 0 {
+			defaultBuf.WriteString(", ")
+		}
+		defaultBuf.WriteString(string(id))
 	}
-	defaultTools := strings.Join(defaultStrings, ", ")
 
 	// Get all available toolsets (excludes context and dynamic for display)
 	allToolsets := r.AvailableToolsets("context", "dynamic")
-	var availableToolsLines []string
+	var availableBuf strings.Builder
 	const maxLineLength = 70
 	currentLine := ""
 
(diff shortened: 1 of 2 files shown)


pkg/registry/registry.go
```diff
@@ -0,0 +1,343 @@
...
+//	    Build()
+//
+// The Registry is configured at build time and provides:
+//   - Filtered access to tools/resources/prompts via Available* methods
+//   - Deterministic ordering for documentation generation
+//   - Lazy dependency injection during registration via RegisterAll()
+//   - Runtime toolset enabling for dynamic toolsets mode
+type Registry struct {
+	// tools holds all tools in this group
+	tools []ServerTool
```
JoannaaKL, line 29: Can each of these collections be a map already?

pkg/registry/registry.go
```diff
@@ -0,0 +1,343 @@
...
+	}
+
+	return result
+}
+
+// ToolsetIDs returns a sorted list of unique toolset IDs from all tools in this group.
+func (r *Registry) ToolsetIDs() []ToolsetID {
+	seen := make(map[ToolsetID]bool)
+	for i := range r.tools {
+		seen[r.tools[i].Toolset.ID] = true
```
JoannaaKL: It doesn't need to be a map, we can use a set here and just check if the id is already there. (Although in go it's not such a big difference, set is anyway a map ;__;)

SamMorrowDrums: Haha, I had a long discussion about sets with Alan Donovan of Go about things like the lack of sets, and he felt they hide stuff that's not actually that efficient. 

Go still doesn't have a Set structure does it? Unless we added a library or something.

pkg/registry/registry.go
SamMorrowDrums: Yes

SamMorrowDrums: I did what makes sense, for fastest iteration ultimately.

PR #1621, opened 2025-12-16
Fix create_or_update SHA-related failures

almaleksia:
create_or_update tool very often fails with `"sha" wasn't supplied"` errors. This PR adds instructions to tool description to use git to get original file's SHA. If it is not possible - the tool will attempt to get sha using `GetContents`.

### DIFF
--- a/pkg/github/repositories_test.go
+++ b/pkg/github/repositories_test.go
@@ -1246,6 +1246,7 @@ func Test_CreateOrUpdateFile(t *testing.T) {
 						"message": "Update without SHA",
 						"content": "IyBVcGRhdGVkCgpVcGRhdGVkIHdpdGhvdXQgU0hBLg==",
 						"branch":  "main",
+						"sha":     "existing123", // SHA is automatically added from ETag
 					}).andThen(
 						mockResponse(t, http.StatusOK, mockFileResponse),
 					),


pkg/github/repositories.go
```diff
@@ -406,11 +413,32 @@ func CreateOrUpdateFile(getClient GetClientFn, t translations.TranslationHelperF
...
+						), nil, nil
+					}
+				}
+			} else {
+				return ghErrors.NewGitHubAPIErrorResponse(ctx,
+					"failed to create/update file",
+					resp,
+					err,
+				), nil, nil
+			}
```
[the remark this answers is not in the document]
almaleksia: good point, fixed

pkg/github/repositories.go
```diff
@@ -406,11 +413,32 @@ func CreateOrUpdateFile(getClient GetClientFn, t translations.TranslationHelperF
...
+					opts.SHA = currentFileContent.SHA
+					fileContent, resp, err = client.Repositories.CreateFile(ctx, owner, repo, path, opts)
+					if err != nil {
+						return ghErrors.NewGitHubAPIErrorResponse(ctx,
+							"failed to create/update file after retrieving current SHA",
+							resp,
+							err,
+						), nil, nil
+					}
+				}
```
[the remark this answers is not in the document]
almaleksia: Added ✅

pkg/github/repositories.go
```diff
@@ -406,11 +413,32 @@ func CreateOrUpdateFile(getClient GetClientFn, t translations.TranslationHelperF
...
+			if strings.Contains(err.Error(), `"sha" wasn't supplied`) && sha == "" {
+				// attempt to get the current file SHA by fetching the file contents
+				getOpts := &github.RepositoryContentGetOptions{
+					Ref: branch,
+				}
+				currentFileContent, _, respContents, err := client.Repositories.GetContents(ctx, owner, repo, path, getOpts)
+				defer func() { _ = respContents.Body.Close() }()
+
+				if err == nil && currentFileContent != nil {
+					opts.SHA = currentFileContent.SHA
```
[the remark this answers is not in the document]
almaleksia: Fixed in https://github.com/github/github-mcp-server/pull/1621/commits/27f3e6a90925d3868512f3a0f77f0c83427fd3fb

pkg/github/repositories.go
```diff
@@ -406,11 +413,57 @@ func CreateOrUpdateFile(getClient GetClientFn, t translations.TranslationHelperF
...
+				// Close the response from the initial failed CreateFile call
+				if resp != nil {
+					_ = resp.Body.Close()
+				}
+
+				// attempt to get the current file SHA by fetching the file contents
+				getOpts := &github.RepositoryContentGetOptions{
+					Ref: branch,
+				}
+				currentFileContent, _, respContents, err := client.Repositories.GetContents(ctx, owner, repo, path, getOpts)
```
SamMorrowDrums: Proposed SHA validation behavior for `create_or_update_file`:

1. No SHA provided (create or blind update):
   - Allow the request
   - Include the previous file's SHA in the response (not just the new commit's SHA)
   - Add a warning suggesting they verify no unintended changes were overwritten

2. SHA provided but stale:
   - Reject the request with an error
   - Include the current SHA in the error message so they can fetch the diff and decide how to proceed

3. SHA provided and matches current:
   - Allow the request (no additional API cost to validate - conditional HEAD is free)

Why this matters:
- Uses conditional HEAD requests (free when SHA matches, no body transfer when stale)
- Avoids expensive GET requests that starts to send/download full file content
- Gives users clear feedback to avoid accidentally overwriting concurrent changes

Roughly

```
// CreateOrUpdateFile handler - after parsing all params and before CreateFile call:

path = strings.TrimPrefix(path, "/")

// SHA validation using conditional HEAD request (efficient - no body transfer)
var previousSHA string
contentURL := fmt.Sprintf("repos/%s/%s/contents/%s", owner, repo, url.PathEscape(path))
if branch != "" {
    contentURL += "?ref=" + url.QueryEscape(branch)
}

if sha != "" {
    // User provided SHA - validate it's still current
    req, err := client.NewRequest("HEAD", contentURL, nil)
    if err == nil {
        req.Header.Set("If-None-Match", fmt.Sprintf(`"%s"`, sha))
        resp, _ := client.Do(ctx, req, nil)
        if resp != nil {
            defer resp.Body.Close()
            
            switch resp.StatusCode {
            case http.StatusNotModified:
                // SHA matches current - proceed (this request was FREE!)
                opts.SHA = github.Ptr(sha)
            case http.StatusOK:
                // SHA is stale - reject with current SHA so user can check diff
                currentSHA := strings.Trim(resp.Header.Get("ETag"), `"`)
                return utils.NewToolResultError(fmt.Sprintf(
                    "SHA mismatch: provided SHA %s is stale. Current file SHA is %s. "+
                    "Use get_file_contents or compare commits to review changes before updating.",
                    sha, currentSHA)), nil, nil
            case http.StatusNotFound:
                // File doesn't exist - this is a create, ignore provided SHA
            }
        }
    }
} else {
    // No SHA provided - check if file exists to warn about blind update
    req, err := client.NewRequest("HEAD", contentURL, nil)
    if err == nil {
        resp, _ := client.Do(ctx, req, nil)
        if resp != nil {
            defer resp.Body.Close()
            if resp.StatusCode == http.StatusOK {
                previousSHA = strings.Trim(resp.Header.Get("ETag"), `"`)
            }
            // 404 = new file, no previous SHA needed
        }
    }
}

// Create the file options
opts := &github.RepositoryContentFileOptions{
    Message: github.Ptr(message),
    Content: contentBytes,
    Branch:  github.Ptr(branch),
}
if sha != "" {
    opts.SHA = github.Ptr(sha)
}

// Proceed with create/update
fileContent, resp, err := client.Repositories.CreateFile(ctx, owner, repo, path, opts)
if err != nil {
    return ghErrors.NewGitHubAPIErrorResponse(ctx,
        "failed to create/update file",
        resp,
        err,
    ), nil, nil
}
defer resp.Body.Close()

// Build response
r, err := json.Marshal(fileContent)
if err != nil {
    return nil, nil, fmt.Errorf("failed to marshal response: %w", err)
}

// Warn if file was updated without SHA validation (blind update)
if sha == "" && previousSHA != "" {
    return utils.NewToolResultText(fmt.Sprintf(
        "Warning: File updated without SHA validation. Previous file SHA was %s. "+
        "Verify no unintended changes were overwritten.\n\n%s", 
        previousSHA, string(r))), nil, nil
}

return utils.NewToolResultText(string(r)), nil, nil
```

- Uses HEAD instead of GetContents (no body = fast & cheap)
- Uses If-None-Match for free validation when SHA matches
- Validates upfront instead of as fallback after failure
- Rejects stale SHAs instead of auto-fetching (safer)
- Warns on blind updates with the previous SHA

We could make some different choices, but this feels like it is getting close. WDYT?

almaleksia: >Add a warning suggesting they verify no unintended changes were overwritten

what's the point of returning SHA after the update was already happened?

almaleksia: In this case what is expected from LLM - comparing previous SHA to the local SHA and reverting the change if they do not match? If so we can state it more explicitly

JoannaaKL: I think there still is a tiny window where we can overwrite a sha, for example if:
- we check the sha, it matches sha mcp server has
/------/ somebody else updates the file, sha changes, we don't know about it
- mcp server continues with the update and surprise! Old file sha was overwritted by our update.

almaleksia: yes, my question is not about that.

almaleksia: @SamMorrowDrums addressed + added some instructions to warning message

SamMorrowDrums: The reason for the sha that was overwritten is so the model can check what was different in that version that it overwrote using git history.

almaleksia: @JoannaaKL my question was about the behavior we expect from LLM if shas mismatch and update already happened

PR #1700, opened 2025-12-30
Add open world hints for tools.

omgitsads:
## Summary

Tools that return user generated content should have open world hints, to allow clients to inform users that the content may be untrusted and prompt them to approve the request before executing it.

## Why

Addresses https://github.com/github/copilot-agent-services/issues/1103

## What changed
- Added `openWorldHint` annotations for tools with user generated content in the response.

## MCP impact

- [ ] No tool or API changes
- [ ] New tool added

Adds annotations to pretty much all Read Only tools, since in some form they will have user generated content in the API result.

## Security / limits

## Tool renaming

## Lint & tests

## Docs

- [ ] Not needed


pkg/github/projects.go
```diff
@@ -32,8 +32,9 @@ func ListProjects(t translations.TranslationHelperFunc) inventory.ServerTool {
 			Name:        "list_projects",
 			Description: t("TOOL_LIST_PROJECTS_DESCRIPTION", `List Projects for a user or organization`),
 			Annotations: &mcp.ToolAnnotations{
-				Title:        t("TOOL_LIST_PROJECTS_USER_TITLE", "List projects"),
-				ReadOnlyHint: true,
+				Title:         t("TOOL_LIST_PROJECTS_USER_TITLE", "List projects"),
+				ReadOnlyHint:  true,
+				OpenWorldHint: jsonschema.Ptr(true),
```
SamMorrowDrums, line 37: Can projects be open world? I think they might only be available inside paying orgs or enterprises although I could be wrong there.

pkg/github/projects.go
```diff
@@ -149,8 +150,9 @@ func GetProject(t translations.TranslationHelperFunc) inventory.ServerTool {
 			Name:        "get_project",
 			Description: t("TOOL_GET_PROJECT_DESCRIPTION", "Get Project for a user or org"),
 			Annotations: &mcp.ToolAnnotations{
-				Title:        t("TOOL_GET_PROJECT_USER_TITLE", "Get project"),
-				ReadOnlyHint: true,
+				Title:         t("TOOL_GET_PROJECT_USER_TITLE", "Get project"),
+				ReadOnlyHint:  true,
+				OpenWorldHint: jsonschema.Ptr(true),
```
SamMorrowDrums, line 155: Same question as list projects

pkg/github/projects.go
```diff
@@ -237,8 +239,9 @@ func ListProjectFields(t translations.TranslationHelperFunc) inventory.ServerToo
 			Name:        "list_project_fields",
 			Description: t("TOOL_LIST_PROJECT_FIELDS_DESCRIPTION", "List Project fields for a user or org"),
 			Annotations: &mcp.ToolAnnotations{
-				Title:        t("TOOL_LIST_PROJECT_FIELDS_USER_TITLE", "List project fields"),
-				ReadOnlyHint: true,
+				Title:         t("TOOL_LIST_PROJECT_FIELDS_USER_TITLE", "List project fields"),
```
SamMorrowDrums, line 242: Similar question I think fields also only created by project admins and not sure public projects exist

SamMorrowDrums (approved):
Awesome, shame about the churn. I think maybe that's from json schema lib update possibly.

Had a couple of questions but I think good to merge and follow up after if needed. Pretty important to get this out, so don't want to delay it.

pkg/github/projects.go
omgitsads: They are open in terms of providing open ended user generated content, i.e. the `title` and `description` fields. IMO this is separate from the idea of trusted content.

If we start getting into that rabbit hole then you could make the argument for everything really, Are your repos/commits/issues open world? Private repos you've been invited to? etc

pkg/github/projects.go
omgitsads: Same here, `title` and `description` of a project in the list are open world.

pkg/github/projects.go
omgitsads: Same thing here, project fields have descriptions that take arbitrary text that would be considered open world.

s-zanella:
In the current MCP specification the default value of `openWorldHint` is `true`, so in an ideal world where clients comply with the protocol specification setting this annotation explicitly is redundant. However, some clients do not comply with the specification and behave differently when `openWorldHint` is not set than when it is set to `true`. To avoid depending on client-defined behavior, it would be preferrable to **always** explicitly include an `openWorldHint` annotation. This requires modifying this PR to add `openWorldHint == false` annotations in the few cases where it's not included.

PR #1192, opened 2025-10-08
Consolidate pull request review tools

almaleksia:
Consolidating pull request review write tools.

`create_and_submit_pull_request_review`
`create_pending_pull_request_review`
`submit_pending_pull_request_review`
`delete_pending_pull_request_review`

are consolidated under

`pull_request_review_write`
with method parameter:
`create`
`submit_pending`
`delete_pending`

CreateAndSubmit and CreatePending merged under `create` method as they differ only by 2 parameters - `event` and `body`.

UPD: validated on curated datasets, no regressions accross models.

PR #1749, opened 2026-01-06
Fix add_comment_to_pending_review for fork PRs

### LINKED ISSUE #1748: MCP server no longer allowing adding comments to pending reviews for fork PRs
Our organisation follows a forked development strategy and we have been making use of googles Gemini CLI github actions (https://github.com/google-github-actions/run-gemini-cli) for AI powered code review. We have noticed that since a little before the Christmas break the agent has been failing to successfully add line comments to pending reviews for PRs raised from forks using the MCP tool `add_comment_to_pending_review`.

We see this issue for PRs raised from forks but not for PRs raised from the same repo.

We use a github app token to authenticate the MCP actions, it has pull request write permissions for the target repo but not for forks.

This usage worked fine for us up until ~December including token/permissions settings and our forking model. Repo settings have not been changed. Therefore we think this is a defect either in Github MCP or with Github itself.

### Affected version

This affects the hosted Github MCP server and the last several versions of the dockerised server.

1. Create a repo
2. Create a fork of the repo
3. Create a PR with a change from the fork to the upstream
4. Attempt to have an AI agent configured with GitHub MCP submit a code review with line or file comments as part of the review
5. Note that these tool calls are rejected

It is expected that comments can be added to reviews on fork PRs, whereas this seems to only work for PRs coming from the same repo.

### Logs

Actions logs when error:
```
Failed to add comment to pending review. Possible reasons:\n\t- The line number doesn't exist in the pull request diff\n\t- The file path is incorrect\n\t- The side (LEFT/RIGHT) is invalid for the specified line\n
```
Note that in this instance the path, line and side are correct and having an agent create the same comment on a PR from the base repo works without issue.

SamMorrowDrums:
## Summary

Fixes #1748

The `add_comment_to_pending_review` tool was failing for pull requests from forks because the `addPullRequestReviewThread` GraphQL mutation requires the `PullRequestID` field to correctly identify the pull request context.

## Problem

When adding review comments to PRs from forks, the API was returning `null` for the thread ID with the error:
```
Failed to add comment to pending review. Possible reasons:
  - The line number doesn't exist in the pull request diff
  - The file path is incorrect
  - The side (LEFT/RIGHT) is invalid for the specified line
```

This happened even when all parameters were correct, and worked fine for PRs from the same repo.

## Root Cause

The `AddPullRequestReviewThreadInput` GraphQL input has two optional ID fields:
- `PullRequestID` - The node ID of the pull request
- `PullRequestReviewID` - The Node ID of the review

The code was only providing `PullRequestReviewID`. For fork PRs, GitHub's API needs `PullRequestID` to correctly resolve the context.

This appears to be a pre-existing bug rather than a regression from the consolidation work in #1192. The `add_comment_to_pending_review` tool was not modified during that consolidation. The issue likely became more visible due to stricter API validation on GitHub's side.

Related: https://github.com/orgs/community/discussions/182548

## Fix

1. Modified the GraphQL query to also fetch `PullRequest.ID`
2. Added `PullRequestID` to the mutation input

## Testing

- All existing tests pass
- Updated test helper to support the new query structure
- Lint passes

### DIFF
--- a/pkg/github/pullrequests.go
+++ b/pkg/github/pullrequests.go
@@ -1743,6 +1743,7 @@ func AddCommentToPendingReview(t translations.TranslationHelperFunc) inventory.S
 			var getLatestReviewForViewerQuery struct {
 				Repository struct {
 					PullRequest struct {
+						ID      githubv4.ID
 						Reviews struct {
 							Nodes []struct {
 								ID    githubv4.ID
@@ -1788,6 +1789,7 @@ func AddCommentToPendingReview(t translations.TranslationHelperFunc) inventory.S
 				} `graphql:"addPullRequestReviewThread(input: $input)"`
 			}
 
+			pullRequestID := getLatestReviewForViewerQuery.Repository.PullRequest.ID
 			if err := client.Mutate(
 				ctx,
 				&addPullRequestReviewThreadMutation,
@@ -1799,6 +1801,7 @@ func AddCommentToPendingReview(t translations.TranslationHelperFunc) inventory.S
 					Side:                newGQLStringlikePtr[githubv4.DiffSide](params.Side),
 					StartLine:           newGQLIntPtr(params.StartLine),
 					StartSide:           newGQLStringlikePtr[githubv4.DiffSide](params.StartSide),
+					PullRequestID:       &pullRequestID,
 					PullRequestReviewID: &review.ID,
 				},
 				nil,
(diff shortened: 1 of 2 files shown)


pkg/github/pullrequests.go
```diff
@@ -1743,6 +1743,7 @@ func AddCommentToPendingReview(t translations.TranslationHelperFunc) inventory.S
 			var getLatestReviewForViewerQuery struct {
 				Repository struct {
 					PullRequest struct {
+						ID      githubv4.ID
```
SamMorrowDrums, line 1746: This is the key fix - adding PullRequestID to the query structure!

SamMorrowDrums (commented):
Testing `add_comment_to_pending_review` on a non-fork PR - this works! The fix is confirmed working for same-repo PRs.

PR #1776, opened 2026-01-09
Bug fix: invalid tool should error out

tommaso-moro:
## Summary

This PR ensures that invalid tool names error out, like in the remote server and as per product requirements. 

The PR Also moves the `cleanTools()` logic inside the `Builder` , which allows us to 
1) clean up the code and be consistent with the toolsets logic for a cleaner caller code 
```
// Before
WithTools(github.CleanTools(cfg.EnabledTools))

// After
WithTools(cfg.EnabledTools)
```
2) clean up the remote server by removing the custom tools cleaning + validation logic there and relying on the inventory to compute unrecognized tools instead (see remote server PR: https://github.com/github/github-mcp-server-remote/pull/587 )

The Builder now fully owns the responsibility of processing tool inputs, including cleaning and validation. Callers don't need to know about these implementation details.

## Why

- Product reason: invalid tool names should error out, this is needed mainly for that
- Engineering reason: this improves our logic for handling invalid tool names and makes it more consistent both between tools and toolsets and also across remote and local MCP 

## What changed

- added logic to throw errors if invalid tool names are provided
- moved `cleanTools` logic to inside the `Builder`
- adds tests

## MCP impact

- [X] No tool or API changes
- [ ] New tool added

- 

## Security / limits

- [X] No security or limits impact

## Tool renaming
- [X] I am not renaming tools as part of this PR

## Lint & tests

- [X] Linted locally with `./script/lint`
- [X] Tested locally with `./script/test`

## Docs

- [X] Not needed

### DIFF
--- a/internal/toolsnaps/toolsnaps_test.go
+++ b/internal/toolsnaps/toolsnaps_test.go
@@ -252,3 +252,63 @@ func TestToolSnapKeysSorted(t *testing.T) {
 	assert.Greater(t, mmmIndex, aaaIndex, "mmm should come after aaa")
 	assert.Greater(t, zzzIndex, mmmIndex, "zzz should come after mmm")
 }
+
+func TestStructFieldOrderingSortedAlphabetically(t *testing.T) {
+	withIsolatedWorkingDir(t)
+
+	// Given a struct with fields defined in non-alphabetical order
+	// This test ensures that struct field order doesn't affect the JSON output
+	type toolWithNonAlphabeticalFields struct {
+		ZField string `json:"zField"` // Should appear last in JSON
+		AField string `json:"aField"` // Should appear first in JSON
+		MField string `json:"mField"` // Should appear in the middle
+	}
+
+	tool := toolWithNonAlphabeticalFields{
+		ZField: "z value",
+		AField: "a value",
+		MField: "m value",
+	}
+
+	// When we write the snapshot
+	t.Setenv("UPDATE_TOOLSNAPS", "true")
+	err := Test("struct_field_order", tool)
+	require.NoError(t, err)
+
+	// Then the snapshot file should have alphabetically sorted keys despite struct field order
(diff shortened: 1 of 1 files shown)


pkg/inventory/builder.go
```diff
@@ -157,8 +185,12 @@ func (b *Builder) Build() *Inventory {
 			// the new consolidated tool is available.
 			if canonical, isAlias := b.deprecatedAliases[name]; isAlias {
 				r.additionalTools[canonical] = true
+			} else if !validToolNames[name] {
+				// Not a valid tool and not a deprecated alias - track as unrecognized
+				unrecognizedTools = append(unrecognizedTools, name)
 			}
 		}
+		r.unrecognizedTools = unrecognizedTools
```
SamMorrowDrums: To me it makes sense if the builder can error, to simply return an error during building, and not allow an invalid inventory to build. The proactive checking is a riskier pattern as people need to remember to keep code checking it. No need to add unrecongnizedTools, it's reasonable to error during build.

pkg/inventory/builder.go
```diff
@@ -127,6 +128,24 @@ func (b *Builder) WithFilter(filter ToolFilter) *Builder {
 	return b
 }
 
+// cleanTools trims whitespace and removes duplicates from tool names.
+// Empty strings after trimming are excluded.
+func cleanTools(tools []string) []string {
```
SamMorrowDrums, line 134: I kind of prefer input parsing to be done at the point of user input, rather than here, that said it's nice that it would work for remote server by having it in shared pipeline.

tommaso-moro:
**Update**: changed the `Builder.Build()` method signature from returning `*Inventory` to returning `(*Inventory, error)`

Benefits:
- Invalid tool configurations now fail immediately at build time rather than silently creating an inventory that must be checked afterward
- Previously, callers had to remember to call `UnrecognizedTools()` after `Build()`. Now the error is impossible to ignore since it's part of the return signature

PR #1823, opened 2026-01-16
feat: mark copilot toolset as default for remote server

SamMorrowDrums:
## Summary

The `copilot` toolset contains important tools like `create_pull_request_with_copilot` and `get_copilot_job_status` that should be available by default in the remote MCP server.

## Changes

- Set `Default: true` on `ToolsetMetadataCopilot`

## Why this approach?

By setting `Default: true` on the toolset metadata in OSS, the remote server automatically picks it up via `GetDefaultToolsetIDs()` without needing any special handling.

Even though no tools in OSS currently use this toolset (the copilot tools are remote-only), having the metadata here allows the remote server to benefit from the default expansion logic already built into the inventory system.

This is cleaner than having the remote server:
1. Override the toolset metadata to set Default: true
2. Build a temporary inventory just to determine defaults
3. Maintain its own list of additional default toolsets

## Impact on OSS users

None - since no OSS tools belong to the `copilot` toolset, this change has no effect on the local MCP server. The toolset will simply be expanded as part of defaults but result in zero additional tools being enabled.

Closes github/github-mcp-server-remote#619

### DIFF
--- a/pkg/inventory/builder.go
+++ b/pkg/inventory/builder.go
@@ -32,7 +32,7 @@ type Builder struct {
 	resourceTemplates []ServerResourceTemplate
 	prompts           []ServerPrompt
 	deprecatedAliases map[string]string
-	toolsetMetadata   []ToolsetMetadata // standalone toolset metadata
+	toolsetMetadata   []ToolsetMetadata // standalone toolset metadata for toolsets without registered tools (e.g., remote-only toolsets)
 
 	// Configuration options (processed at Build time)
 	readOnly        bool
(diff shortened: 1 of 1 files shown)


SamMorrowDrums:
@copilot - I'd like to change the approach here. Instead of just setting `Default: true` on the metadata (which doesn't work because `GetDefaultToolsetIDs()` only iterates actual tools), let's make the inventory builder handle "default" properly:

**Current problem:**
- `GetDefaultToolsetIDs()` collects defaults by iterating through actual tools that have `Toolset.Default == true`
- If a toolset has `Default: true` but no tools in the OSS repo use it (like `copilot`), it won't appear in defaults
- This makes it impossible for remote servers to declare toolsets as defaults via metadata alone

**Proposed solution:**
1. In the inventory builder, when processing `.WithToolsets()`, treat "default" as a special keyword
2. Instead of expanding "default" to a list of toolset IDs, add a filter that includes tools where `tool.Toolset.Default == true`
3. Remove "default" from the toolset list before matching against specific toolset IDs
4. Alternatively, have a method like `DefaultToolsetMetadatas()` that returns all `ToolsetMetadata` where `Default: true`, not just those with tools

This way:
- Setting `Default: true` on `ToolsetMetadataCopilot` will actually work
- Remote servers that add tools to the `copilot` toolset will automatically get them in defaults
- The declarative `Default: true` on metadata becomes the source of truth

Please implement this change to make `Default: true` on toolset metadata actually control whether tools in that toolset are included by default.

PR #1849, opened 2026-01-20
Add Streamable HTTP mode.

omgitsads:
## Summary

Adds support for Streamable HTTP mode.

## Why

Addresses https://github.com/github/copilot-mcp-core/issues/1125.

## What changed
- Adds a `http` subcommand to the core binary.
- Moves Param methods into their own file
- Extract inventory creation out to a factory method, so that we can provide our own implementation in our remote server
- Extract server creation to a factory method, so that we can provide our own implementation in remote.
- Consolidate MCP Server creation in the `github` package, to be used by both STDIO and HTTP mode.
- Decoupled Inventory & Dependency creation from MCP Server creation, to allow custom implementations
- Moved API URL building out to the `utils` package
- `GetRepoAccessCache` and `GetFlags` methods on `ToolDependencies` now require a `context.Context`.
- Implemented a request based dependency
- Added Scope Challenge middleware for progressive token scope escalation.

## MCP impact

- [ ] No tool or API changes
- [ ] New tool added

Tool changes are limited to new `ToolDependencies` interface changes.

## Security / limits

This is mostly porting across our existing Remote HTTP implementation.

## Tool renaming

## Lint & tests

- [ ] Linted locally with `./script/lint`

## Docs

- [ ] Not needed

### DIFF
--- a/cmd/github-mcp-server/main.go
+++ b/cmd/github-mcp-server/main.go
@@ -142,7 +142,7 @@ func init() {
 	httpCmd.Flags().Int("port", 8082, "HTTP server port")
 	httpCmd.Flags().String("base-url", "", "Base URL where this server is publicly accessible (for OAuth resource metadata)")
 	httpCmd.Flags().String("base-path", "", "Externally visible base path for the HTTP server (for OAuth resource metadata)")
-	httpCmd.Flags().Bool("scope-challenge", false, "Enable OAuth scope challenge responses and tool filtering based on token scopes")
+	httpCmd.Flags().Bool("scope-challenge", false, "Enable OAuth scope challenge responses")
 
 	// Bind flag to viper
 	_ = viper.BindPFlag("toolsets", rootCmd.PersistentFlags().Lookup("toolsets"))


pkg/github/dependencies.go
```diff
@@ -221,3 +235,169 @@ func NewToolFromHandler(
...
+	if d.GQLClient != nil {
+		return d.GQLClient, nil
+	}
+
+	// extract the token from the context
+	token, _ := ghcontext.GetTokenInfo(ctx)
+
+	// Construct GraphQL client
+	// We use NewEnterpriseClient unconditionally since we already parsed the API host
+	gqlHTTPClient := &http.Client{
```
SamMorrowDrums, line 312: We should be wrapping the transport to use the graphql features: https://github.com/github/github-mcp-server-remote/blob/a1ac0d596860727ada2477fb1e51dd765b03835b/pkg/http/mcp.go#L270

Ah I see we do so here: https://github.com/github/github-mcp-server/pull/1849/changes#diff-e722b319ba2e477b26a8285df61a7b8960f99cbd36c3c7711aa7d1d02846d1e5R20

Just need to make sure whatever we do will translate to our remote repo I guess. Ideally automatically 😍

pkg/http/headers/headers.go
```diff
@@ -0,0 +1,39 @@
...
+	// MCPReadOnlyHeader indicates whether the MCP is in read-only mode.
+	MCPReadOnlyHeader = "X-MCP-Readonly"
+	// MCPToolsetsHeader is a comma-separated list of MCP toolsets that the request is for.
+	MCPToolsetsHeader = "X-MCP-Toolsets"
+	// MCPToolsHeader is a comma-separated list of MCP tools that the request is for.
+	MCPToolsHeader = "X-MCP-Tools"
+	// MCPLockdownHeader indicates whether lockdown mode is enabled.
+	MCPLockdownHeader = "X-MCP-Lockdown"
+	// MCPFeaturesHeader is a comma-separated list of feature flags to enable.
+	MCPFeaturesHeader = "X-MCP-Features"
```
SamMorrowDrums, line 45: Probably should shove graphql feautures header here also: https://github.com/github/github-mcp-server-remote/blob/a1ac0d596860727ada2477fb1e51dd765b03835b/pkg/http/headers/http.go#L28

pkg/http/middleware/token.go
```diff
@@ -0,0 +1,102 @@
...
+	authTypeGhToken
+)
+
+var (
+	errMissingAuthorizationHeader     = fmt.Errorf("%w: missing required Authorization header", mark.ErrBadRequest)
+	errBadAuthorizationHeader         = fmt.Errorf("%w: Authorization header is badly formatted", mark.ErrBadRequest)
+	errUnsupportedAuthorizationHeader = fmt.Errorf("%w: unsupported Authorization header", mark.ErrBadRequest)
+)
+
+var supportedThirdPartyTokenPrefixes = []string{
```
SamMorrowDrums: ThirdParty seems like a strange name

pkg/http/middleware/token.go
```diff
@@ -0,0 +1,102 @@
...
+// were 40 characters long and only contained the characters a-f and 0-9.
+var oldPatternRegexp = regexp.MustCompile(`\A[a-f0-9]{40}\z`)
+
+func ExtractUserToken() func(next http.Handler) http.Handler {
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			_, token, err := parseAuthorizationHeader(r)
+			if err != nil {
+				// For missing Authorization header, return 401 with WWW-Authenticate header per MCP spec
+				if errors.Is(err, errMissingAuthorizationHeader) {
```
SamMorrowDrums: Should we add scopes here, or write a ticket to do so? I guess the following make sense here:

- https://github.com/github/copilot-mcp-core/issues/1193
- https://github.com/github/copilot-mcp-core/issues/904

And we can deal with the ability to do it within GH infra later, but perhaps add a ticket to do this for the OSS first, as we more easily can.

pkg/http/middleware/token.go
```diff
@@ -0,0 +1,102 @@
...
+		return 0, "", errMissingAuthorizationHeader
+	}
+
+	switch {
+	// decrypt dotcom token and set it as token
+	case strings.HasPrefix(authHeader, "GitHub-Bearer "):
+		return 0, "", errUnsupportedAuthorizationHeader
+	default:
+		// support both "Bearer" and "bearer" to conform to api.github.com
+		if len(authHeader) > 7 && strings.EqualFold(authHeader[:7], "Bearer ") {
```
SamMorrowDrums: We really should use `strings.Equalfold` I gave that feedback previously, but now it's going in this repo I feel even more strongly 😂 https://www.geeksforgeeks.org/go-language/strings-equalfold-function-in-golang-with-examples/

Maybe I'll make a PR.

pkg/http/handler.go
```diff
@@ -0,0 +1,182 @@
...
+}
+
+// withToolset is middleware that extracts the toolset from the URL and sets it in the request context
+func withToolset(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		toolset := chi.URLParam(r, "toolset")
+		ctx := ghcontext.WithToolsets(r.Context(), []string{toolset})
+		next.ServeHTTP(w, r.WithContext(ctx))
+	})
+}
```
SamMorrowDrums, line 162: Should we be handling the header here too? I guess not, I just want to make sure we are very clear about precedence order for the URL and headers, I think URL probably wins, as it shows very clear intent, but I want to make sure we either don't change behavior (probably what is happening), or we do something very clear at least.

omgitsads: Afaik, this will take precedence over headers due to the way the middleware stacking happens. The Request parsing middleware happens at the top router level, then these middlewares happen immediately prior to the route handling and will reset the `WithToolsets`/`WithReadonly` contexts.

pkg/http/middleware/token.go
omgitsads: Is this referencing the scope challenge stuff, or something different @SamMorrowDrums?

pkg/http/middleware/token.go
omgitsads: Agreed. Will change this!

pkg/http/middleware/token.go
SamMorrowDrums: This is different, I meant maybe we could add the required scopes to the WWW-Authenticate header. Don't have to worry about it now.

pkg/http/middleware/token.go
omgitsads: Am I missing something here, was this previously not `EqualFold` when you commented and the diff is not marked as outdated?

pkg/github/dependencies.go
```diff
@@ -221,3 +235,181 @@ func NewToolFromHandler(
...
+
+	gqlClient, err := d.GetGQLClient(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	// Create repo access cache
+	instance := lockdown.GetInstance(gqlClient, d.RepoAccessOpts...)
+	d.RepoAccessCache = instance
+	return instance, nil
```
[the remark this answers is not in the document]
omgitsads: Remove the memoization of the clients. Not ideal, but most tools will only call `GetClient()` once. Maybe we can look back at the `sync.Once` once at another point.

pkg/context/token.go
```diff
@@ -0,0 +1,39 @@
...
+// WithTokenInfo adds TokenInfo to the context
+func WithTokenInfo(ctx context.Context, tokenInfo *TokenInfo) context.Context {
+	return context.WithValue(ctx, tokenCtxKey, tokenInfo)
+}
+
+func SetTokenScopes(ctx context.Context, scopes []string) {
+	if tokenInfo, ok := GetTokenInfo(ctx); ok {
+		tokenInfo.Scopes = scopes
+		tokenInfo.ScopesFetched = true
+	}
```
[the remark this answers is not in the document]
omgitsads: Re-arranged the PAT scope checking to address this.

pkg/http/headers/headers.go
omgitsads: Added this in!

mattdholloway (approved):

PR #1895, opened 2026-01-26
refactor: avoid utils package name to fix revive issue

alexandear:
## Summary
Rename the `pkg/utils` package to `pkg/mcpresult` and `pkg/githubapi` with clearer function names to resolve a `revive.var-naming` issue.

## Why
The `pkg/utils` package name is "bad", triggering a revive linter warning. Renaming to `pkg/mcpresult` and `pkg/githubapi` makes the purpose explicit and improves code clarity. Function names are simplified from `NewToolResultX` to `NewX` to align with Go naming conventions and reduce redundancy.

## What changed
- Created new `pkg/mcpresult` package with clearer function names: `NewText`, `NewError`, `NewErrorFromErr`, `NewResource`
- Created new `pkg/githubapi` package
- Updated all files to import and use `mcpresult` or `githubapi` directly

## MCP impact
- [x] No tool or API changes
- [ ] New tool added

## Security / limits

## Tool renaming
   This refactoring renames internal utility functions and packages, not MCP tools.

## Lint & tests

## Docs

- [ ] Not needed


SamMorrowDrums:
@alexandear am I missing something? The linting setup in CI doesn't seem to flag this, and if we are making lint fix PRs I would expect that they include enabling rules in CI etc. otherwise there is nothing to prevent regression.

pkg/utils/result.go
```diff
@@ -1,49 +1,27 @@
-package utils //nolint:revive //TODO: figure out a better name for this package
+// Package utils is deprecated: use pkg/mcpresult instead
+package utils //nolint:revive // vague package name
```
alexandear: @SamMorrowDrums this `//nolint:revive` directive is reason the issue is not shown. If I remove it, we see the following:

```console
$ ./script/lint
pkg/utils/result.go:2:9: var-naming: avoid meaningless package names (revive)
package utils
        ^
1 issues:
* revive: 1
```

I intentionally kept this `utils` package to preserve backward compatibility. If you want, I can remove this package entirely, but that would be a breaking change for anyone who imports `pkg/utils` in their code.

SamMorrowDrums: Right of course, sorry 😅. Was just passing through when I had a glance. That makes total sense and thank you for looking at it!

alexandear:
@SamMorrowDrums I have updated the PR and fixed merge conflicts. While it was under review, two more files were added to the `utils` package - `api.go` and `token.go` in #1849. I think we should decide how we want to handle this PR, because `utils` tends to become a catch-all. Developers often put new code there instead of choosing more specific package names.

`utils` is a common name that can have everything.

alexandear:
Too many conflicts.

PR #1957, opened 2026-02-04
Add initial PoC for MCP Apps for select tools under Insiders

mattdholloway:
## Summary

Based on initial work by @tommaso-moro

This PR adds support for [MCP Apps](https://modelcontextprotocol.io/docs/extensions/apps), enabling rich interactive UIs for MCP tools.

### New MCP App UIs
- **`get_me`** - User profile display with avatar and stats
- **`issue_write`** - Create/update issues with title, body, and Markdown editor
- **`create_pull_request`** - Create PRs with title, body, and draft/regular toggle

### Implementation
- **UI Layer** (`ui/`): React + Primer design system, built to single HTML files via `script/build-ui`
- **Server Integration**: HTML assets embedded in binary, exposed as MCP resources linked via tool metadata

### Insiders Mode
All MCP Apps functionality gated behind Insiders
- UI resources and metadata only available when enabled
- Generic `insidersOnlyMetaKeys` mechanism for future experimental features

These new UIs require a new React app that exists under the `ui` folder with components made available in iframes for clients that support the feature. This has been extensively tested under VSCode-Insiders.

This will require additional support in Remote MCP to ensure Insiders and resources support works correctly.

## Why

As part of https://github.com/github/copilot-mcp-core/issues/1125

## What changed

- Added `ui/` directory with React + Primer design system apps for `get_me`, `issue_write`, and `create_pull_request`
- Added `script/build-ui` to compile React apps into single HTML files
- Added `pkg/github/ui_embed.go` to embed built HTML assets into binary
- Added `pkg/github/ui_resources.go` to register UI resources with the MCP server
- Added UI metadata (`Meta.ui.resourceUri`) to `get_me`, `issue_write`, and `create_pull_request` tools
- Added CSP configuration for avatar image loading in `get_me` resource
- Added `_ui_submitted` mechanism to distinguish UI form submissions from LLM tool calls
- Added `WithInsidersMode()` to inventory builder to gate experimental features
- Added `insidersOnlyMetaKeys` mechanism to strip insiders-only metadata when disabled
- Updated CI workflows to include UI build step

## MCP impact

- [ ] No tool or API changes
- [ ] New tool added

- 

## Security / limits

## Tool renaming

## Lint & tests

## Docs

- [ ] Not needed


ui/src/apps/issue-write/App.tsx
```diff
@@ -0,0 +1,1114 @@
...
+            {selectedRepo ? selectedRepo.fullName : "Select repository"}
+          </ActionMenu.Button>
+          <ActionMenu.Overlay width="large">
+            <ActionList selectionVariant="single">
+              <Box px={3} py={2}>
+                <TextInput
+                  placeholder="Search repositories..."
+                  value={repoFilter}
+                  onChange={(e) => {
+                    console.log("Repo filter changed:", e.target.value);
```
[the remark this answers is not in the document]
mattdholloway: Keeping this for now as it is an Insiders feature

pkg/github/pullrequests.go
```diff
@@ -518,7 +482,50 @@ func CreatePullRequest(t translations.TranslationHelperFunc) inventory.ServerToo
...
+					"draft": {
+						Type:        "boolean",
+						Description: "Create as draft PR",
+					},
+					"maintainer_can_modify": {
+						Type:        "boolean",
+						Description: "Allow maintainer edits",
+					},
+				},
+				Required: []string{"owner", "repo"},
```
tommaso-moro: Why did `Required` change from `Required: []string{"owner", "repo", "title", "head", "base"}` to `Required: []string{"owner", "repo"}` ?

tommaso-moro (commented):
Great work on this, and it's really cool to see the Primer design system in action here! 🚀

A few comments:

---

I see you added two tools (`list_milestones` and `list_assignees`) that are meant to be used only by the Apps in the iframes. My concerns are:
1) I am worried about how this will this play out if users are configuring specific tools in their server config. From my understanding, we'd be introducing a dependency of some tools on other tools (suddenly, `issue_write` relies on `list_assignees` for example). This is a problem because users are not aware of these dependencies (and shouldn't be) when configuring their server. So if they don't enable both `issue_write` and `list_assignees` at the same time, things would break I assume. I wonder if there is another way of doing this. Wdyt?
2) (Less important for these 2 tools but relevant if we build more) These tools appear in `tools/list` and consume context. Now it's only 2 additional tools, but if we follow this pattern when building more UIs we can easily end up having a lot of tools which can be expensive context-wise


---

Does the input schema change from `Required: []string{"owner", "repo", "title", "head", "base"}` to `Required: []string{"owner", "repo"}` affect non-insiders users ?

---

I noticed some issues with the UI/UX when trying this out locally. Some related to general spacing and layout (probs easy to fix) but other seem like core markdown functionality that is broken, this is a blocker even for an Insiders ship. Here are some of the main ones:

**1- The 'Add a quote' functionality in the markdown editor is broken**


<details>
  <summary>Demo:</summary>
It renders like this
Image
</details>

**2- The repository selection dropdown is too wide and overflows outside the app's window.** 

<details>
  <summary>Demo:</summary>
  
  https://github.com/user-attachments/assets/eace071d-bdcb-4289-9075-6d93d1068cf8


</details>


**3- Missing spacing between the bottom dropdowns. Also, can't scroll horizontally to reveal the ones past "Milestone"**

<details>
  <summary>Demo:</summary>
  
  Image
</details>


**4- The top part of the PR creation UI overflows** 

<details>
  <summary>Demo:</summary>

Image
</details>


**5- Lists (both numbered and bullet points) don't fully work**, as pressing "enter" doesn't create a new list item, it just creates a new line. This is not super high priority but should ideally be fixed.

<details>
  <summary>Demo:</summary>
  
  https://github.com/user-attachments/assets/fdf8612b-fa00-49e9-b197-87acf23571aa

</details>




**6- Missing spacing here between "create as draft" and "allow maintainer edits"** 

<details>
  <summary>Demo:</summary>
Image
</details>

At first glance, it seems that a lot of the markdown issues are due to the fact that `MarkdownEditor.tsx` is using a simple `textarea` instead of a markdown editor. There may be an official markdown editor from the Primer lib or similar that could be used?

(nit: I think the React code can be improved but haven't reviewed it because I assume the priority in this PR is just to make it work for experimental testing)

pkg/github/pullrequests.go
mattdholloway: This changed so that the model doesn't need to provide these fields upfront, they can be filled by the user in the UI

IrynaKulakova:
A few issues I noticed while testing:

**get_me:** Avatar is not being downloaded/displayed. I think this we want to fix.

**create_issue:** Fields like Type, Assignee, Labels, and Milestones are not populated with suggestions. The Title field looks like a link, but it doesn’t open or navigate anywhere
**create_pr**: No suggestion appears to select the "pick me" branch

If suggesting predefined fields (like labels, milestones, etc.) is complex to implement at this stage, we can consider moving that out of scope for the Insiders launch.
Let me know if you need any clarification or help reproducing these!

pkg/github/pullrequests.go
tommaso-moro: Makes sense! Just as a sanity check do you think it may also affect non-insiders users? (e.g. by causing the model to omit params or hallucinate)

mattdholloway: It could be worth investigating if we can deterministically choose this based on whether Insiders is enabled or not

mattdholloway: After some consideration I'm going to revert this change for now. In future, this feature can be added in a followup PR

pkg/github/ui_embed.go
```diff
@@ -0,0 +1,41 @@
...
+	"embed"
+)
+
+// UIAssets embeds the built MCP App UI HTML files.
+// These files are generated by running `script/build-ui` which compiles
+// the React/Primer components in the ui/ directory.
+//
+//go:embed ui_dist/*.html
+var UIAssets embed.FS
+
```
Rion4, line 13: Add a comment explaining why the placeholder file is needed.

pkg/github/ui_resources.go
```diff
@@ -0,0 +1,89 @@
...
+
+import (
+	"context"
+
+	"github.com/modelcontextprotocol/go-sdk/mcp"
+)
+
+// RegisterUIResources registers MCP App UI resources with the server.
+// These are static resources (not templates) that serve HTML content for
+// MCP App-enabled tools. The HTML is built from React/Primer components
```
Rion4, line 11: Add comments to the resource registration to explain how it relates to the external app UI.

ui/src/apps/pr-write/App.tsx
tommaso-moro, line 1: I know we're not doing a thorough review of the tsx code, but I noticed that here we wrap <AppProvider> multiple times (in success view, loading, error state, and main render) instead of once at the root like get-me and issue-write do. Not a blocker but can create unnecessary re-renders and is inconsistent with what we do for `get-me` and `issue_write` , would be a quick fix

tommaso-moro (commented):
Just tested this again, nice improvements! Some things I have noticed

1) What is the behaviour of these tools on hosts where MCP Apps is not supported? For example, in `issue_write` and `create_pull_request`, when Insiders is on and `_ui_submitted` is false, the tool early-returns a placeholder message like "Ready to create an issue". So it seems like on a host without MCP Apps support, users would get the placeholder text instead of the tool actually executing. I think we should have a way for people opting into insiders but using hosts that won't support Apps (which include 1P ones like Copilot CLI) to still be able to use the tool without the UI 

2) On hover, the "Write" and "Preview" options in the markdown look inconsistent with the rich text options, in that they don't have rounded corners. A quick fix is to round the corners to make them consistent

3) In the markdown editor, the toolbar isn't responsive: I cannot scroll horizontally to reveal some of the options (like bullet point list, numbered list) 
Screenshot 2026-02-11 at 17 21 37

4) The UI to create a PR doesn't seem to work anymore: when I click "create pull request" it effectively creates the pull request, but it shows the error message instead of the success UI, even though it does create the PR

Screenshot 2026-02-12 at 11 17 36


5) In the issue creation success screen, we need to add spacing between the issue icon and the title of the issue (same for the PR success UI)
Screenshot 2026-02-11 at 17 22 44

6) (nit) I have noticed that I cannot just pull up the create issue UI without populating it with title and body to create an issue. It only works by having the model write a test title and body, calling the `issue_write` tool and then showing the form pre-populated with that mock text for me to edit. Is that intentional? I wonder if we should let folks see an empty UI that they can fill out


7) (nit) maybe unrelated, but getting 422 error when trying to create a draft PR as it assumes my default branch is `main` but instead it's `master`. Don't think this was introduced with these changes but wanted to flag, maybe we can show a more helpful error message here
<details>
  <summary>Screenshot</summary>
Screenshot 2026-02-11 at 17 27 09
</details>

ui/src/apps/pr-write/App.tsx
mattdholloway: This can be fixed in a follow-up PR, we're not worrying too much about re-renders at this stage

tommaso-moro: I'll leave it up to you 👍

pkg/github/ui_capability.go
```diff
@@ -2,25 +2,26 @@ package github
...
-// for the MCP Apps extension identifier.
+// uiSupportedClients lists client names (from ClientInfo.Name) known to
+// support MCP Apps UI rendering.
 //
-// When the client does not support MCP Apps, tools should skip any UI-gated
-// flow (e.g., interactive forms) and execute the action directly.
+// This is a temporary workaround until the Go SDK adds an Extensions field
+// to ClientCapabilities (see https://github.com/modelcontextprotocol/go-sdk/issues/777).
+// Once that lands, detection should use capabilities.extensions instead.
+var uiSupportedClients = map[string]bool{
```
tommaso-moro, line 11: We could add [Goose](https://github.com/block/goose) here too, as it also supports MCP Apps

tommaso-moro (dismissed):
Left a comment, but overall LGTM for Insiders!

tommaso-moro (dismissed):
Overall LGTM for Insiders!

tommaso-moro (approved):

.gitignore
arshadigital, line 2: Arshadigital.ir

arshadigital, line 5: Arshadigital.ir

arshadigital, line 16: Arshadigital.ir

arshadigital (commented):
Arshadigital.ir

PR #1993, opened 2026-02-11
Add shared SchemaCache for streamable-http server

SamMorrowDrums:
## Summary

Creates a shared `mcp.SchemaCache` in `RunHTTPServer` and passes it through to each per-request MCP Server via `ServerOptions.SchemaCache`.

## Motivation

In streamable-http (stateless) mode, a new `mcp.Server` is created for every request. Without a shared cache, JSON schema reflection and resolution is repeated on every request for every tool. The go-sdk's `SchemaCache` (`sync.Map`-backed, goroutine-safe) caches these schemas across server instances, matching the pattern already used by the remote server.

## Changes

- **`pkg/http/server.go`**: Create `mcp.NewSchemaCache()` once at startup, pass via `WithSchemaCache` option
- **`pkg/http/handler.go`**: Add `SchemaCache` field to `Handler`/`HandlerOptions`, `WithSchemaCache` option func, and set `so.SchemaCache` on per-request `ServerOptions`

### DIFF
--- a/pkg/http/handler.go
+++ b/pkg/http/handler.go
@@ -40,7 +40,6 @@ type HandlerOptions struct {
 	OAuthConfig            *oauth.Config
 	ScopeFetcher           scopes.FetcherInterface
 	FeatureChecker         inventory.FeatureFlagChecker
-	SchemaCache            *mcp.SchemaCache
 }
 
 type HandlerOption func(*HandlerOptions)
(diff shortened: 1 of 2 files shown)


.github/workflows/lint.yml
```diff
@@ -16,7 +16,7 @@ jobs:
       - uses: actions/checkout@v6
       - uses: actions/setup-go@v6
         with:
-          go-version: stable
+          go-version: '1.25'
```
SamMorrowDrums, line 19: Linter requires this

pkg/http/handler.go
```diff
@@ -112,6 +122,7 @@ func NewHTTPMcpHandler(
 		inventoryFactoryFunc:   inventoryFactory,
 		oauthCfg:               opts.OAuthConfig,
 		scopeFetcher:           scopeFetcher,
+		schemaCache:            opts.SchemaCache,
```
omgitsads: Do you think there would be any scenario where we don't want a schema cache for HTTP?

I feel like we can drop the `WithSchemaCache` when initializing (still keeping `WithSchemaCache`), and init an empty schema cache by default?
I guess the default right now is without a schema cache, but is that a good default? Maybe if you explicitly _don't_ want one being able to pass `nil`, but that is difficult to manage with the current way that Options work

SamMorrowDrums: Yeah that's a good question, I mean I don't think we want to make it optional, because in the end we're in control and we don't want it pluggable, it's our opinionated server after all. I'll spin off copilot to riff on that and see if we like it.

PR #1997, opened 2026-02-11
Token scopes context

omgitsads:
## Summary
So that I can setup the scopes of a token before it is authed, separate the TokenScopes from the TokenInfo contexts. In the remote MCP server, by the time `ExtractUserToken` happens, we already know that 

## Why

Gives us greater flexibility for handling authentication in a gateway and pre-populating token scopes.

## What changed
- Store scopes in a Seperate context window. 

## MCP impact

- [x] No tool or API changes
- [ ] New tool added

- 

## Security / limits

## Tool renaming

## Lint & tests

## Docs

- [x] Not needed


pkg/http/handler.go
```diff
@@ -278,8 +287,10 @@ func PATScopeFilter(b *inventory.Builder, r *http.Request, fetcher scopes.Fetche
 	// Only classic PATs (ghp_ prefix) return OAuth scopes via X-OAuth-Scopes header.
 	// Fine-grained PATs and other token types don't support this, so we skip filtering.
 	if tokenInfo.TokenType == utils.TokenTypePersonalAccessToken {
-		if tokenInfo.ScopesFetched {
```
almaleksia, line 281: just curious, was there a reason we needed this flag at all?

almaleksia (approved):
LGTM

pkg/http/handler.go
omgitsads: I was trying to handle the scenario where we had a token info from copilot API, but it was a token with no scopes. But having it combined with the TokenInfo meant the ordering was a little awkward.

omgitsads: It's also implied now, if `ok` is false from `scopes, ok := ghcontext.GetTokenScopes(ctx)`

PR #2016, opened 2026-02-16
Optimize token usage in default `list_` tools

kerobbi:
## Summary

Adds a generic response optimisation package that reduces token usage in `list_` tool responses by applying six strategies at runtime: nested object flattening, URL elimination, zero-value elimination, whitespace normalisation, collection summarisation, and fill-rate filtering.

Each relevant tool passes its own config via functional options:
- `WithPreservedFields` - keys exempt from destructive strategies
- `WithCollectionExtractors` - extract important values from nested collections instead of summarising as `[N items]`
- `WithMaxDepth` - override flatten depth if needed

## Why

Current JSON responses are token heavy, with lots of redundant nested objects, URL fields, and zero-value noise. A single `list_pull_request` tool call at 25 items can use 180k tokens. This pipeline brings that down to 55k (~69% reduction) without losing anything the model actually needs.

## What changed

- Added `pkg/response` - generic optimisation pipeline with per-tool configuration via functional options
- Added tests covering all strategies and config interactions
- Updated existing tests for flattened output format
- Wired `response.OptimizeList` into 6 default `list_` tools (`list_pull_requests`, `list_issues`, `list_commits`, `list_tags`, `list_releases`, `list_branches`)
- Did _not_ wire `list_issue_types`, as its data is already flat and the pipeline produces no benefit

**Token reduction**

- `list_commits`: ~21-25%
- `list_tags`: ~73-74%
- `list_releases`: ~91-94%
- `list_issues`: ~5-15%
- `list_pull_requests`: ~54-73%
- `list_branches`: ~5-9%

Measured using OAI's `tiktoken` library (`o200k_base`) at 2, 10, and 25 items. Model accuracy validated with `GPT 5.1` and `Opus 4.5` across 6 prompts per tool, no degradation observed.

## MCP impact

- [ ] No tool or API changes
- [ ] New tool added

- 

## Security / limits

## Tool renaming

## Lint & tests

## Docs

- [x] Not needed

### DIFF
--- a/pkg/github/ui_capability.go
+++ b/pkg/github/ui_capability.go
@@ -11,6 +11,9 @@ import (
 // advertise MCP Apps UI support.
 const mcpAppsExtensionKey = "io.modelcontextprotocol/ui"
 
+// MCPAppMIMEType is the MIME type for MCP App UI resources.
+const MCPAppMIMEType = "text/html;profile=mcp-app"
+
 // clientSupportsUI reports whether the MCP client that sent this request
 // supports MCP Apps UI rendering.
 // It checks the context first (set by HTTP/stateless servers from stored
--- a/pkg/github/ui_resources.go
+++ b/pkg/github/ui_resources.go
@@ -17,7 +17,7 @@ func RegisterUIResources(s *mcp.Server) {
 			URI:         GetMeUIResourceURI,
 			Name:        "get_me_ui",
 			Description: "MCP App UI for the get_me tool",
-			MIMEType:    "text/html",
+			MIMEType:    MCPAppMIMEType,
 		},
 		func(_ context.Context, _ *mcp.ReadResourceRequest) (*mcp.ReadResourceResult, error) {
 			html := MustGetUIAsset("get-me.html")
@@ -25,7 +25,7 @@ func RegisterUIResources(s *mcp.Server) {
 				Contents: []*mcp.ResourceContents{
 					{
 						URI:      GetMeUIResourceURI,
-						MIMEType: "text/html",
+						MIMEType: MCPAppMIMEType,
 						Text:     html,
 						// MCP Apps UI metadata - CSP configuration to allow loading GitHub avatars
 						// See: https://github.com/modelcontextprotocol/ext-apps/blob/main/specification/draft/apps.mdx
@@ -49,7 +49,7 @@ func RegisterUIResources(s *mcp.Server) {
 			URI:         IssueWriteUIResourceURI,
 			Name:        "issue_write_ui",
 			Description: "MCP App UI for creating and updating GitHub issues",
-			MIMEType:    "text/html",
+			MIMEType:    MCPAppMIMEType,
 		},
 		func(_ context.Context, _ *mcp.ReadResourceRequest) (*mcp.ReadResourceResult, error) {
 			html := MustGetUIAsset("issue-write.html")
@@ -57,7 +57,7 @@ func RegisterUIResources(s *mcp.Server) {
 				Contents: []*mcp.ResourceContents{
 					{
 						URI:      IssueWriteUIResourceURI,
-						MIMEType: "text/html",
+						MIMEType: MCPAppMIMEType,
 						Text:     html,
 					},
 				},
@@ -71,7 +71,7 @@ func RegisterUIResources(s *mcp.Server) {
 			URI:         PullRequestWriteUIResourceURI,
 			Name:        "pr_write_ui",
 			Description: "MCP App UI for creating GitHub pull requests",
-			MIMEType:    "text/html",
+			MIMEType:    MCPAppMIMEType,
 		},
 		func(_ context.Context, _ *mcp.ReadResourceRequest) (*mcp.ReadResourceResult, error) {
 			html := MustGetUIAsset("pr-write.html")
@@ -79,7 +79,7 @@ func RegisterUIResources(s *mcp.Server) {
 				Contents: []*mcp.ResourceContents{
 					{
 						URI:      PullRequestWriteUIResourceURI,
-						MIMEType: "text/html",
+						MIMEType: MCPAppMIMEType,
 						Text:     html,
 					},
 				},


pkg/response/optimize.go
```diff
@@ -0,0 +1,297 @@
...
+		if len(trimmed) > 0 {
+			result = append(result, trimmed)
+		}
+	}
+
+	return result
+}
+
+// isURLKey matches "url", "*_url", and their dot-prefixed variants.
+func isURLKey(key string) bool {
```
tonytrg, line 158:
```suggestion
func isURLKey(key string) bool {
    if idx := strings.LastIndex(key, "."); idx >= 0 {
        key = key[idx+1:]
    }
    return key == "url" || strings.HasSuffix(key, "_url")
}
```

it is just a value so you could simplify it.

pkg/response/optimize.go
```diff
@@ -0,0 +1,297 @@
...
+	"labels":              {"name"},
+	"requested_reviewers": {"login"},
+	"requested_teams":     {"name"},
+}
+
+// MarshalItems is the single entry point for response optimization.
+// Handles two shapes: plain JSON arrays and wrapped objects with metadata.
+// An optional maxDepth controls how many nesting levels flatten will recurse
+// into; it defaults to maxFlattenDepth when omitted.
+func MarshalItems(data any, maxDepth ...int) ([]byte, error) {
```
tonytrg: I would personally prefer every object that gets optimized to be able to pass a preservedfields/collectionfieldextractors?

Unless its a very specific few structs that would need it, since it only applies to returned list tools anyways?

kerobbi: I thought about this too but ended up deciding to keep the (small) configs global atm, primarily for simplicity.

kerobbi: Most of the fields are relevant to single tools (`html_url` and `labels` are the only ones that get applied to 2+ tools), no tools really need custom configs (so far) and as you said, everything applies only to `list_` tools (a limited subset for that matter).

That said per-tool config is definitely something I'll keep in mind, as maybe `X` field is useful for one tool, but redundant for another in the future.

pkg/response/optimize.go
```diff
@@ -0,0 +1,296 @@
...
```
tonytrg: i think the function should be called different:

like OptimizeList

tonytrg: the function interface should prob be more strict on the input.

We dont want to handle every edge case inside the function and let the caller pass whatever he wants.

Maybe as interface something like:

// could additionally take the preservedFields,  collectionFieldExtractors
//optimize.go
```
type OptimizeConfig struct {
    maxDepth int
}

type OptimizeOption func(*OptimizeConfig)

func WithMaxDepth(d int) OptimizeOption {
    return func(c *OptimizeConfig) {
        c.maxDepth = d
    }
}

func MarshalItems[T any](items []T, opts ...OptimizeOption) ([]byte, error) {
    cfg := OptimizeConfig{maxDepth: maxFlattenDepth}
    for _, opt := range opts {
        opt(&cfg)
    }

    raw, err := json.Marshal(items)
    if err != nil {
        return nil, fmt.Errorf("failed to marshal data: %w", err)
    }

    var maps []map[string]any
    if err := json.Unmarshal(raw, &maps); err != nil {
        return nil, fmt.Errorf("failed to unmarshal JSON: %w", err)
    }

    return json.Marshal(OptimizeItems(maps, cfg.maxDepth))
}
```

//pullrequests.go
```
r, err := response.MarshalItems(prs)

// custom depth
r, err := response.MarshalItems(minimalCommits, response.WithMaxDepth(3))
```

//issues.go
```
// optimize just the issues list
optimized, err := response.MarshalItems(issues, response.WithMaxDepth(3))
if err != nil {
    return nil, nil, fmt.Errorf("failed to marshal issues: %w", err)
}

// wrap with metadata — caller's responsibility
out, err := json.Marshal(map[string]any{
    "issues":     json.RawMessage(optimized),
    "pageInfo": map[string]any{
        "hasNextPage":     pageInfo.HasNextPage,
        "hasPreviousPage": pageInfo.HasPreviousPage,
        "startCursor":     pageInfo.StartCursor,
        "endCursor":       pageInfo.EndCursor,
    },
    "totalCount": totalCount,
})
if err != nil {
    return nil, nil, fmt.Errorf("failed to marshal response: %w", err)
}
```

pkg/response/optimize.go
```diff
@@ -0,0 +1,296 @@
...
+	depth := maxFlattenDepth
+	if len(maxDepth) > 0 {
```
tonytrg: what is the reason behind using variadic maxdepth?

we only ever pass a single value

pkg/response/optimize.go
```diff
@@ -0,0 +1,296 @@
...
+		depth = maxDepth[0]
+	}
+
+	raw, err := json.Marshal(data)
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal data: %w", err)
+	}
+
+	switch raw[0] {
```
tonytrg: we should be much stricter on the function interface so we do not have to handle every edge case in the function itself.

it should stay simple and only do whats expected of it.

in this case I think optimize a list of different objects

tonytrg: https://github.com/github/github-mcp-server/pull/2016/changes#r2827696330

if we enforce a stricter input we can get rid of all this conditional branching

no optimizeObject, straight into optimizing the array.

tonytrg:
I have left a few comments, let me know what you think.

pkg/response/optimize.go
tonytrg: meaning we can get rid of:
- optimizearray
- optimizeobject
- optimizeItems(uplifted directly into MarshalItems)

pkg/response/optimize.go
kerobbi: Revisited this as per our discussion.

pkg/response/optimize.go
kerobbi: It was honestly just a shortcut to avoid making every caller pass the default, fixed now.

pkg/response/optimize.go
kerobbi: Love the overall feedback here 🙂 Deleted `optimizeArray` and `optimizeObject`, inlined `optimizeItems` and applied the exact pattern you described in the comment above.

tonytrg:
Thematically they fit to our current approach of creating minimal versions.
Id rather see us creating minimal versions of every list tool call object(if it does not exist already), without looking I assume we also offer the single get variant for most of them.

The approach to handle raw json strings and manually cutting down the objects with default values, or having to manually append exceptions imo justifies manual handcrafting of the reduced minimal version.

Afterwards there is still room for sensible optimizations for list based responses.

Like:
- transform to minimal version
- additionally apply any extra optimization only applicable on lists

## Conversion from original to optimized struct

Requires:
```
- defaultFillRateThreshold = 0.1 // default proportion of items that must have a key for it to survive
- minFillRateRows          = 3   // minimum number of items required to apply fill-rate filtering
- defaultMaxDepth

preservedFields      map[string]bool
collectionExtractors map[string][]string
```

While we do already have a way to reduce bloateds structs.
Example minimalPR, see:
https://github.com/github/github-mcp-server/blob/de12f8f12b262ba09ecf2f68f41cbacb3cd97355/pkg/github/pullrequests.go#L1172
https://github.com/github/github-mcp-server/blob/de12f8f12b262ba09ecf2f68f41cbacb3cd97355/pkg/github/pullrequests.go#L190
Problems I see:
- two different optimized/minimal representations of a struct
  - maintaining two different code sections instead of sharing code
  - I would say that the minimized representation from a single `get_` to its `list_` method having the same fields is a hard criteria
- harder to imagine the output/final state for the reduced variant in optimizedlist, with this approach of passing configs/exemptions

kerobbi:
@tonytrg I've wired minimal types into the 4 tools that were missing (`list_pull_requests`, `list_issues`, `list_releases`, `list_tags`) and removed `collectionExtractors` entirely, as with minimal types that mechanism is redundant.

The optimisation pipeline now does: flatten -> URL removal -> zero-value removal -> whitespace normalisation -> fill rate filtering. Reasoning behind why I kept each (plus the `preservedFields` config) is:

- flatten: minimal types still have nested structs (i.e., `MinimalUser`, `MinimalPRBranch`)
- URL removal: `MinimalUser` includes two URL fields, which might be useful in `get_` tools, but honestly noise in `list_` (this also acts as a safety net if someone adds a URL field to a minimal type in the future)
  - an alternative here would be separate struct variants (i.e., get vs list), but imo the pipeline has less boilerplate for the same result
- Zero-value removal: similar tradeoff and alternative solution as the above
  - `omitempty` would work but it affects `get_` tools too (e..g, N * `merged: false` or all the reaction counts are just noise in a `list_` response imo). 
- whitespace normalisation and fill rate filtering: these are runtime only transformations
- `preservedFields`: minimal config needed to protect `html_url` from URL removal, `draft/prerelease` from zero value removal

On get/list tools field parity, I have somewhat mixed feelings here. These tool types serve different enough purposes that they benefit from different field sets imo. List tools calls done first to find items, then get tools calls to drill in.. having fewer fields in list and richer fields in get feels like the right separation to me. The minimal types are now shared between `get_` and `list_`, so we at least have a consistent base representation. The pipeline just strips noise that doesn't add value at scale, making list output a strict subset of get output rather than a potential entirely different shape.

pkg/response/optimize.go
```diff
@@ -0,0 +1,171 @@
...
+// within nested maps are preserved at their dotted key position.
+// Recursion stops at the given maxDepth; deeper nested maps are dropped.
+func flattenTo(item map[string]any, maxDepth int) map[string]any {
+	result := make(map[string]any, len(item))
+	flattenInto(item, "", result, 1, maxDepth)
+	return result
+}
+
+// flattenInto is the recursive worker for flattenTo.
+func flattenInto(item map[string]any, prefix string, result map[string]any, depth int, maxDepth int) {
```
JoannaaKL, line 91: Code style nit: I'd much rather prefer this function to return a parameter than to return void and fill it in the body. Is there any particular reason you chose to do it this way?
If depth >= maxDepth we just return an empty map, maybe we should keep it as-is?

pkg/response/optimize.go
```diff
@@ -0,0 +1,171 @@
...
+// remove zero-values, normalize whitespace.
+// Preserved fields skip everything except whitespace normalization.
+func optimizeItem(item map[string]any, cfg OptimizeListConfig) map[string]any {
+	result := make(map[string]any, len(item))
+	for key, value := range item {
+		preserved := cfg.preservedFields[key]
+		if !preserved && isURLKey(key) {
+			continue
+		}
+		if !preserved && isZeroValue(value) {
```
JoannaaKL, line 143: Won't `isZeroValue` cause fields like `draft` in pull request to be removed? I don't think we want to strip important fields 🤔

pkg/github/issues.go
```diff
@@ -1598,9 +1599,21 @@ func ListIssues(t translations.TranslationHelperFunc) inventory.ServerTool {
...
+			}
+
+			optimizedIssues, err := response.OptimizeList(minimalIssues)
+			if err != nil {
+				return nil, nil, fmt.Errorf("failed to optimize issues: %w", err)
+			}
+
+			// Wrap optimized issues with pagination metadata
+			issueResponse := map[string]any{
+				"issues": json.RawMessage(optimizedIssues),
```
JoannaaKL, line 1621: For empty issues list we used to return an empty list `{"issues":[],...}`
which is no longer the case after this optimization

pkg/response/optimize.go
```diff
@@ -0,0 +1,171 @@
...
+		c.preservedFields = make(map[string]bool, len(fields))
+		for _, f := range fields {
+			c.preservedFields[f] = true
+		}
+	}
+}
+
+// OptimizeList optimizes a list of items by applying flattening, URL removal, zero-value removal,
+// whitespace normalization, and fill-rate filtering.
+func OptimizeList[T any](items []T, opts ...OptimizeListOption) ([]byte, error) {
```
JoannaaKL, line 47: OptimizeList does marshal + unmarshal + marshal. For large list responses this may introduce noticeable latency.

pkg/github/minimal_types.go
```diff
@@ -672,6 +678,32 @@ func convertToMinimalBranch(branch *github.Branch) MinimalBranch {
...
+		m.PublishedAt = release.PublishedAt.Format(time.RFC3339)
+	}
+
+	return m
+}
+
+func convertToMinimalTag(tag *github.RepositoryTag) MinimalTag {
+	return MinimalTag{
+		Name: tag.GetName(),
+		SHA:  tag.GetCommit().GetSHA(),
```
JoannaaKL: Nil checks missing

pkg/response/optimize.go
```diff
@@ -0,0 +1,171 @@
...
+func WithMaxDepth(d int) OptimizeListOption {
+	return func(c *OptimizeListConfig) {
+		c.maxDepth = d
+	}
+}
+
+// WithPreservedFields sets keys that are exempt from all destructive strategies except whitespace normalization.
+// Keys are matched against post-flatten map keys, so for nested fields like "user.html_url", the dotted key must be
+// added explicitly. Empty collections are still dropped. Wins over collectionExtractors.
+func WithPreservedFields(fields ...string) OptimizeListOption {
```
JoannaaKL, line 34: WithPreservedFields replaces rather than accumulates and calling it twice discards the first set. If a caller chains two WithPreservedFields options, only the last one wins with no warning:
```
response.OptimizeList(items,
    response.WithPreservedFields("html_url"),
    response.WithPreservedFields("draft"),
)
```
Such config will result in `html_url` not being preserved.

pkg/response/optimize.go
kerobbi: This was to avoid creating + merging a new map at each recursive layer.

pkg/response/optimize.go
kerobbi: `draft` specifically is not removed due to being part of `WithPreservedFields`. However, for fields like `merged: false` for example on open PRs, zero value removal / `isZeroValue` is intentional. N * `merged: false` is noise in a list response where the `get_` tool variant can be used as a fallback for details.

pkg/github/issues.go
kerobbi: `OptimizeList` still produces `[]` on an empty slice 🤔 I've also just tested it against an issue-less repo and this is what was returned:

```
{"issues":[],"pageInfo":{"endCursor":"","hasNextPage":false,"hasPreviousPage":false,"startCursor":""},"totalCount":0}
```

pkg/response/optimize.go
kerobbi: Fixed it, it now accumulates instead of replacing 🙂

pkg/response/optimize.go
kerobbi: This unfortunately is the tradeoff of having a generic optimisation pipeline. An alternative was using Go reflection (which the spike did), but that's significantly more complex (to read and maintain) imo. It also still has some of the drawbacks of this solution.

I do agree with the entire json round trip potentially adding overhead when a large amount of items is requested, but it should be in the low ms range at worst (which I'd dare to say is negligible vs the actual API call?).

pkg/response/optimize.go
JoannaaKL: I think for the 2-3 levels of nesting and each item having ~5 nested objects that'd be 3-5 additional map allocations per nested item. In the grand scheme of things (marshalling, unmarshalling pipeline) it's negligible. I know it's idiomatic go code but I much more prefer clean functions simply returning parameters (although it's a matter of taste so I am not pushing on this).
What needs to be addressed is that we always drop the last level in the nested structures due to `depth < maxDepth` check. You can reproduce it by copy-pasting this test and running it:
```
t.Run("silently drops nested maps at maxDepth boundary", func(t *testing.T) {
		// With defaultMaxDepth=2, depth starts at 1.
		// Result: "user.org" and its contents are silently lost.
		input := map[string]any{
			"title": "issue",
			"user": map[string]any{
				"login": "alice",
				"org": map[string]any{
					"name": "acme",
				},
			},
		}
		result := flattenTo(input, defaultMaxDepth)

		assert.Equal(t, "issue", result["title"])
		assert.Equal(t, "alice", result["user.login"])
		// BUG: org data is silently dropped — neither "user.org" nor
		// "user.org.name" appear in the result.
		assert.Nil(t, result["user.org"], "nested map at maxDepth is silently dropped")
		assert.Nil(t, result["user.org.name"], "nested map contents at maxDepth are lost")
	})
```

pkg/response/optimize.go
JoannaaKL: What about `protected`? That's an example, I didn't go though the whole list of boolean fields that are important but will be stripped. Looks like a lot of additional configuration needs to be added to make sure we aren't truncating important data?

pkg/response/optimize.go
```diff
@@ -0,0 +1,173 @@
...
+		return nil, fmt.Errorf("failed to unmarshal JSON: %w", err)
+	}
+
+	for i, item := range maps {
+		flattenedItem := flattenTo(item, cfg.maxDepth)
+		maps[i] = optimizeItem(flattenedItem, cfg)
+	}
+
+	if len(maps) >= minFillRateRows {
+		maps = filterByFillRate(maps, defaultFillRateThreshold, cfg)
```
JoannaaKL, line 73: There is another issue with the ordering. 
The pipeline order is 
1. flattenTo
2. optimizeItem
2. filterByFillRate
So zero-value removal artificially deflates fill rates, causing fill-rate filtering to wipe out keys that had legitimate non-zero values on some items.

Concrete example with list_branches (no preserved fields):

10 branches: 1 has protected: true, 9 have protected: false optimizeItem strips protected: false from 9 items (it's a zero value)
Then filterByFillRate sees protected on only 1/10 items --> minCount = int(0.1 * 10) = 1 → 1 > 1 is false and it's removed
The one branch that was protected: true loses that information entirely.

pkg/response/optimize.go
kerobbi: Regarding the code style nit, more than happy to address it in a follow up PR.

As for the second point, that is actually intentional. The main reason for this is that deeply nested data is not useful in a list response imo. 

I did just notice that the PR description is generic though and focused on flattening only, my bad! It completely slipped off my mind, as this behaviour was part of the nested object flattening strategy as a whole.

Will edit the PR description and add a test to cover this behaviour.

pkg/github/issues.go
```diff
@@ -1603,9 +1604,21 @@ func ListIssues(t translations.TranslationHelperFunc) inventory.ServerTool {
...
+			minimalIssues := make([]MinimalIssue, 0, len(issues))
+			for _, issue := range issues {
+				if issue != nil {
+					minimalIssues = append(minimalIssues, convertToMinimalIssue(issue))
+				}
+			}
+
+			optimizedIssues, err := response.OptimizeList(minimalIssues)
+			if err != nil {
+				return nil, nil, fmt.Errorf("failed to optimize issues: %w", err)
```
JoannaaKL: This is different from `list_pull_requests`, where if a call to OptimizeList fails we return `if err != nil {
				return utils.NewToolResultErrorFromErr("failed to marshal response", err), nil, nil
			}`

JoannaaKL:
I added some tests to my branch, just to show you what I mean https://github.com/github/github-mcp-server/pull/2081

pkg/response/optimize.go
kerobbi: `protected` is actually a very good catch, as there's no `get_` branch tool as a fallback. I've added it to `WithPreservedFields` in `list_branches`.

Regarding the broader concern, there are two alternatives I considered (i.e., what I also briefly mentioned in https://github.com/github/github-mcp-server/pull/2016#issuecomment-3947484106):

- either add `omitempty` on relevant fields (this breaks `get_` tools)
  - alternative to not breaking `get_` tools is having separate types for get and list, and this is a lot more boilerplate imo
- no zero value removal at all (-> leads to noisy lists)

Overall, there is a risk of maybe leaving out a field, but other than the one you caught just now, all the other tools have `get_` fallbacks, which the model can use.

pkg/github/issues.go
kerobbi: Added a `nil` guard in `OptimizeList`, so that `[]` is returned in that case too.

kerobbi:
> I added some tests to my branch, just to show you what I mean #2081

Tried my best at using these as reference and addressing every concern here..

Regarding the whitespace normalisation vs "markdown with code blocks, bullet lists, and intentional line breaks", I don't think there's a quick fix for this without making the approach smarter (e.g., skipping content inside code fences?). 

It is a valid concern and I'll try revisiting this in a follow up, but overall I am of the idea that `list_` tools are for scanning, and `get_` tools should be used to return the full body. 

Imo this is acceptable for now (I did not consider this specific scenario during evals, but overall, when a prompt such as `summarise the last N items from X` was used, the model never had a problem giving accurate replies).

pkg/response/optimize.go
kerobbi: That order is intentional (and intentionally aggressive). I might sound like a broken record here, but what the optimisation pipeline should do imo is optimise for a workflow where list tools are for scanning, and get tools are for getting in-depth details. 

Stripping everything that isn't high signal at the list level should be fine, the worst thing that could happen in practice is that the model ends up using the `get_` tool to get more specifics.

With that being said, I did spend some time exploring a couple of alternatives:

1. Swapping the order (fill rate first, then zero value removal):
  - this might be better for JSON, but it creates a problem for other output formats (e.g., for CSV like format, sparse bool with 1 `true` and 99 empty cells adds noise, which I'm pretty sure scales linearly with amount of items). 

2. Skipping `bools` in zero value removal entirely:
  - no data loss, but keeps N * `false` values in the output, which does add up at scale and goes against the entire goal of this work

Keeping the current order and having `WithPreservedFields` is imo the most explicit way of controlling what is retained (i.e., high priority fields) and it works consistently across formats.

pkg/response/optimize.go
JoannaaKL: Ok but then this comment is not correct: `// Recursion stops at the given maxDepth; deeper nested maps are dropped.`. Recursion stops at maxDepth-1 actually.

PR #2046, opened 2026-02-19
chore: gh-host as oauth auth server

atharva1051:
## Summary
Conditionally set AuthorizationServer only when `--gh-host` is explicitly provided, ensuring `github.com` users receive the correct default.

## Why
To ensure that OAuth protected resource metadata correctly exposes the authorization server endpoint for all clients, including those using `github.com` where `cfg.Host` might be empty. This conditional setting prevents an invalid `AuthorizationServer` value and ensures proper discovery of authentication details.

## What changed
- `pkg/http/server.go`: `AuthorizationServer` is now only overridden when `cfg.Host != ""`. Otherwise, it falls back to `DefaultAuthorizationServer` (`https://github.com/login/oauth`).

## MCP impact
- [x] No tool or API changes

N/A

## Security / limits
- [x] Auth / permissions considered — OAuth authorization server URL now correctly resolves for both github.com and GHES

## Tool renaming

## Lint & tests

## Docs

- [x] Not needed

The program was tested solely for our own use cases, which might differ from yours.

Atharva Patil <[atharva.a.patil@mercedes-benz.com](mailto:atharva.a.patil@mercedes-benz.com)> on behalf of Mercedes-Benz Research And Development India, [Provider Information](https://github.com/mercedes-benz/foss/blob/master/PROVIDER_INFORMATION.md)


pkg/http/server.go
```diff
@@ -127,6 +127,9 @@ func RunHTTPServer(cfg ServerConfig) error {
 		BaseURL:      cfg.BaseURL,
 		ResourcePath: cfg.ResourcePath,
 	}
+	if cfg.Host != "" {
+		oauthCfg.AuthorizationServer = cfg.Host + "login/oauth"
```
SamMorrowDrums: It would be best to join the parts using the `net/url` package.

Cc @omgitsads

atharva1051: I will implement that change. However, is a test case really necessary in this instance?

arhiv1973b (approved):
LGTM! ✅

atharva1051:
Hello @SamMorrowDrums or @omgitsads ,

Could you please review this once?

Thank you

pkg/http/server.go
```diff
@@ -127,6 +128,14 @@ func RunHTTPServer(cfg ServerConfig) error {
 		BaseURL:      cfg.BaseURL,
 		ResourcePath: cfg.ResourcePath,
 	}
+	if cfg.Host != "" {
+		u := &url.URL{
+			Scheme: "https",
```
omgitsads: We can't assume that the configured API host is on https here since you can run GHES on `http`. We have the `apiHost` above that is resolving all the relevant URLS for a given hostname, which feels like a good place to put this.

omgitsads:
Hi @atharva1051, thanks for raising this issue. I think theres some additional complexity around handling GHES for both `http/https`, so not quite as simple as just putting the `cfg.Host` in.

Like I said in the review, I think a better place to make this decision is in during the url resolution in the `APIHostResolver` implementation, then passing the resolver into the OAuth Handler. I put together [a quick PR](https://github.com/github/github-mcp-server/pull/2070) of that, but happy to make those changes here so that your contribution is not just dismissed.

atharva1051:
Oh that would be really helpful, although i did end up making few changes the moment I saw your comment.
But I got the idea. Thanks :) Do also re-review my changes for learning purposes.

atharva1051:
@omgitsads  made the changes you made in #2070 
Thanks again

PR #2047, opened 2026-02-19
Fix for -1 in `tailLines`

mattdholloway:
## Summary

This pull request introduces a small but important improvement to how the default number of log lines is handled when processing job logs. The main change ensures that if an invalid or non-positive value is provided for the number of log lines to display, a sensible default of 500 lines is used.

## Why

Fixes this exception:
```
makeslice: len out of range
```

## What changed

**Log handling improvements:**

* In `pkg/buffer/buffer.go`, updated `ProcessResponseAsRingBufferToEnd` to default `maxJobLogLines` to 500 if a non-positive value is provided, and to cap it at 100,000 if it exceeds that value.
* In `pkg/github/actions.go`, changed the logic to set `tailLines` to 500 if it is not specified or set to a non-positive value, instead of only when it is zero.

## MCP impact

- [x] No tool or API changes
- [ ] New tool added

- 

## Security / limits

## Tool renaming

## Lint & tests

## Docs

- [x] Not needed

### DIFF
--- a/pkg/github/actions.go
+++ b/pkg/github/actions.go
@@ -702,8 +702,8 @@ For single job logs, provide job_id. For all failed jobs in a run, provide run_i
 			if err != nil {
 				return utils.NewToolResultError(err.Error()), nil, nil
 			}
-			// Default to 500 lines if not specified
-			if tailLines == 0 {
+			// Default to 500 lines if not specified or invalid
+			if tailLines <= 0 {
 				tailLines = 500
 			}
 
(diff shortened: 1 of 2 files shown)


pkg/github/actions.go
```diff
@@ -702,8 +702,8 @@ For single job logs, provide job_id. For all failed jobs in a run, provide run_i
 			if err != nil {
 				return utils.NewToolResultError(err.Error()), nil, nil
 			}
-			// Default to 500 lines if not specified
-			if tailLines == 0 {
+			// Default to 500 lines if not specified or invalid
+			if tailLines <= 0 {
```
JoannaaKL, line 706: Nice!

PR #2134, opened 2026-03-04
Fix SHA validation in create_or_update_file

### LINKED ISSUE #2133: Bug: create_or_update_file SHA validation uses ETag instead of blob SHA, causing spurious failures
## Summary

`create_or_update_file` has multiple bugs in its SHA validation logic that cause it to fail on legitimate update operations. The tool should be split into separate `create_file` and `update_file` tools, or the validation logic needs significant fixes.

### 1. ETag vs Blob SHA Mismatch (Critical)

The SHA validation performs a HEAD request and compares the user-provided blob SHA against the response's ETag header:

```go
req.Header.Set("If-None-Match", fmt.Sprintf(`"%s"`, sha))
// ...
case http.StatusOK:
    currentSHA := strings.Trim(resp.Header.Get("ETag"), `"`)
    return utils.NewToolResultError(fmt.Sprintf(
        "SHA mismatch: provided SHA %s is stale...", sha, currentSHA))
```

**The problem:** GitHub's Contents API returns a weak ETag (`W/"..."`) that is NOT the blob SHA. It's an opaque server-generated value. Comparing a blob SHA against an ETag is comparing two different things, causing the validation to reject perfectly valid updates where the SHA is actually current.

### 2. `url.PathEscape` Breaks Multi-Segment Paths

```go
contentURL := fmt.Sprintf("repos/%s/%s/contents/%s", owner, repo, url.PathEscape(path))
```

For paths like `sassymcp/modules/vision.py`, `PathEscape` encodes `/` as `%2F`, producing `sassymcp%2Fmodules%2Fvision.py`. The Contents API expects literal path separators. This causes the HEAD validation request to 404, which then falls through to the "new file" code path. The subsequent `CreateFile` call then fails with 409 Conflict because the file already exists but no SHA was provided.

Should use `url.PathEscape` per-segment or just pass the raw path (the go-github client handles encoding).

### 3. Blind Update ETag-as-SHA Injection

When no SHA is provided and the file exists, the code extracts the ETag and uses it as the SHA for the update:

```go
previousSHA = strings.Trim(resp.Header.Get("ETag"), `"`)
// ...
if previousSHA != "" {
    opts.SHA = github.Ptr(previousSHA)
}
```

Since the ETag is not a blob SHA, this causes `client.Repositories.CreateFile` to fail with a 409 because GitHub's API rejects the invalid SHA value.

### 4. Multiple Deferred Body Closes

The validation HEAD requests and the final CreateFile call all stack `defer resp.Body.Close()`. If response bodies aren't fully drained before the next request on the same HTTP/2 connection, this can cause connection reuse issues.

## Recommended Fix

**Option A: Split into two tools**
- `create_file` — creates new files only, no SHA parameter needed
- `update_file` — updates existing files, requires SHA parameter, validates using `get_file_contents` blob SHA (not ETag)

**Option B: Fix the validation**
1. Replace the HEAD+ETag approach with a GET to `/repos/{owner}/{repo}/contents/{path}?ref={branch}` and compare against the response's `.sha` field (the actual blob SHA)
2. Fix `url.PathEscape` to not encode path separators
3. Don't inject ETag values as SHA parameters

## Workaround

Use `push_files` instead of `create_or_update_file` for all update operations. `push_files` uses the Git Data API (create tree → create commit → update ref) which completely bypasses the Contents API and its SHA validation issues. It works reliably for both creates and updates, single or multi-file.

## Environment

- GitHub MCP Server (hosted by Anthropic as Claude.ai connector)
- Reproduces on: `sassyconsultingllc/SassyMCP` repo, updating existing files on `main` branch
- Failure rate: ~30-40% of update attempts via `create_or_update_file`

almaleksia:
## Summary

We incorrectly return SHA mismatch for correct updates. 

Changes:

1. sha is required for existing file updates
2. drop usage of HEAD + Etag and PUT with If-none-match header - we cannot use this approach
3. sha is taken from the metadata returned by Contents API

## Why

Fixes https://github.com/github/github-mcp-server/issues/2133

## What changed

- 
- 

## MCP impact

- [ ] No tool or API changes
- [ ] New tool added

- 

## Security / limits

## Tool renaming
- [ ] I am not renaming tools as part of this PR

## Lint & tests

- [ ] Linted locally with `./script/lint`

## Docs

- [ ] Not needed

### DIFF
--- a/pkg/github/repositories.go
+++ b/pkg/github/repositories.go
@@ -322,7 +322,7 @@ func CreateOrUpdateFile(t translations.TranslationHelperFunc) inventory.ServerTo
 If updating, you should provide the SHA of the file you want to update. Use this tool to create or update a file in a GitHub repository remotely; do not use it for local file operations.
 
 In order to obtain the SHA of original file version before updating, use the following git command:
-git rev-parse HEAD:<path to file>
+git rev-parse <branch>:<path to file>
 
 SHA MUST be provided for existing file updates.
 `),
@@ -448,8 +448,8 @@ SHA MUST be provided for existing file updates.
 					if currentSHA != sha {
 						return utils.NewToolResultError(fmt.Sprintf(
 							"SHA mismatch: provided SHA %s is stale. Current file SHA is %s. "+
-								"Pull the latest changes and use git rev-parse HEAD:%s to get the current SHA.",
-							sha, currentSHA, path)), nil, nil
+								"Pull the latest changes and use git rev-parse %s:%s to get the current SHA.",
+							sha, currentSHA, branch, path)), nil, nil
 					}
 				}
 			} else {
@@ -477,8 +477,8 @@ SHA MUST be provided for existing file updates.
 					// File exists but no SHA was provided - reject to prevent blind overwrites
 					return utils.NewToolResultError(fmt.Sprintf(
 						"File already exists at %s. You must provide the current file's SHA when updating. "+
-							"Use git rev-parse HEAD:%s to get the blob SHA, then retry with the sha parameter.",
-						path, path)), nil, nil
+							"Use git rev-parse %s:%s to get the blob SHA, then retry with the sha parameter.",
+						path, branch, path)), nil, nil
 				}
 				// If file not found, no previous SHA needed (new file creation)
 			}
--- a/pkg/github/repositories.go
+++ b/pkg/github/repositories.go
@@ -428,7 +428,17 @@ SHA MUST be provided for existing file updates.
 				if respCheck != nil {
 					_ = respCheck.Body.Close()
 				}
-				if getErr == nil && existingFile != nil {
+				if getErr != nil {
+					// 404 means file doesn't exist - proceed (new file creation)
+					// Any other error (403, 500, network) should be surfaced
+					if respCheck == nil || respCheck.StatusCode != http.StatusNotFound {
+						return ghErrors.NewGitHubAPIErrorResponse(ctx,
+							"failed to verify file SHA",
+							respCheck,
+							getErr,
+						), nil, nil
+					}
+				} else if existingFile != nil {
 					currentSHA := existingFile.GetSHA()
 					if currentSHA != sha {
 						return utils.NewToolResultError(fmt.Sprintf(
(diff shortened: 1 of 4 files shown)


pkg/github/repositories.go
```diff
@@ -491,23 +474,6 @@ If the SHA is not provided, the tool will attempt to acquire it by fetching the
 
 			minimalResponse := convertToMinimalFileContentResponse(fileContent)
 
-			// Warn if file was updated without SHA validation (blind update)
-			if sha == "" && previousSHA != "" {
```
almaleksia, line 495: removing blind update completely

pkg/github/repositories.go
```diff
@@ -420,55 +419,39 @@ If the SHA is not provided, the tool will attempt to acquire it by fetching the
...
-				opts.SHA = github.Ptr(previousSHA)
+				if getErr == nil && existingFile != nil {
+					// File exists but no SHA was provided - reject to prevent blind overwrites
+					return utils.NewToolResultError(fmt.Sprintf(
+						"File already exists at %s. You must provide the current file's SHA when updating. "+
+							"Use git rev-parse HEAD:%s to get the blob SHA, then retry with the sha parameter.",
+						path, path)), nil, nil
+				}
+				// If file not found or error, no previous SHA needed (new file creation)
 			}
```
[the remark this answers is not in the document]
almaleksia: fixed

pkg/github/repositories.go
```diff
@@ -420,55 +419,39 @@ If the SHA is not provided, the tool will attempt to acquire it by fetching the
...
+				existingFile, _, respCheck, getErr := client.Repositories.GetContents(ctx, owner, repo, path, getOpts)
+				if respCheck != nil {
+					_ = respCheck.Body.Close()
 				}
-			}
-
-			if previousSHA != "" {
-				opts.SHA = github.Ptr(previousSHA)
+				if getErr == nil && existingFile != nil {
+					// File exists but no SHA was provided - reject to prevent blind overwrites
```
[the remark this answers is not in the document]
almaleksia: fixed

pkg/github/repositories.go
```diff
@@ -323,9 +322,9 @@ func CreateOrUpdateFile(t translations.TranslationHelperFunc) inventory.ServerTo
 If updating, you should provide the SHA of the file you want to update. Use this tool to create or update a file in a GitHub repository remotely; do not use it for local file operations.
 
 In order to obtain the SHA of original file version before updating, use the following git command:
-git ls-tree HEAD <path to file>
+git rev-parse HEAD:<path to file>
 
-If the SHA is not provided, the tool will attempt to acquire it by fetching the current file contents from the repository, which may lead to rewriting latest committed changes if the file has changed since last retrieval.
+SHA MUST be provided for existing file updates.
```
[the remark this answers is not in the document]
almaleksia: fixed

pkg/github/repositories.go
SamMorrowDrums: I think it's fair to accept the errors, thought this is probably the most controversial change.

SamMorrowDrums (approved):
All makes sense.

pkg/github/repositories.go
almaleksia: I decided if GitHub API requires sha for updates who are we to make it optional :D

SamMorrowDrums: Well, I think it's fine to make agent fail, but I don't think it's wrong either way, just different trade-offs.

You still retain a git record of what you changed, so it might be surprising to update a file that has changed since you last checked, but you can at least still work out what happened.

The concern for me was always concurrent updates from different agents and then thinking independently that two things were done, but actually only one was because the other overwrites it. That's not great, and I think it's ok for us to not allow that by accident.

leuasseurfarrelds247-arch (approved):

PR #2137, opened 2026-03-04
fix: handle lightweight tags in get_tag tool

fidesit:
## Summary

The `get_tag` tool returns a 404 error when used with lightweight tags because it unconditionally calls the Git Tag Object API, which only works for annotated tags.

- Adds `ref.Object.Type` check after resolving the tag reference
- Lightweight tags (`"commit"` type): resolves via `Git.GetCommit()` and returns a `github.Tag`-compatible structure
- Annotated tags (`"tag"` type): unchanged behavior via `Git.GetTag()`

## Bug

In `pkg/github/repositories.go`, the `GetTag` handler resolves a tag reference via `client.Git.GetRef()`, then always calls `client.Git.GetTag()` with the ref's SHA. This works for annotated tags (where the SHA points to a tag object), but fails for lightweight tags (where the SHA points directly to a commit — the Git Tag Object API returns 404).

**Reproduce:** Create a lightweight tag (`git tag v1.0.0 && git push --tags`), then call `get_tag` → 404.

## Changes

**`pkg/github/repositories.go`**
- After getting the tag reference, check `ref.Object.Type`
- If `"commit"` (lightweight): call `client.Git.GetCommit()`, return a `github.Tag` struct with commit message and author mapped to tagger
- If `"tag"` (annotated): proceed with existing `client.Git.GetTag()` call (no behavior change)

**`pkg/github/repositories_test.go`**
- Added `mockLightweightTagRef` and `mockCommit` fixtures
- Added `"successful lightweight tag retrieval"` test case
- Updated existing `mockTagRef` to include `Type: github.Ptr("tag")` for accuracy

## Notes

- No schema change — the tool returns the same `github.Tag` JSON structure for both tag types
- Backward compatible — annotated tag responses are identical to before
- No toolsnap update needed (schema unchanged)

### DIFF
--- a/pkg/github/repositories.go
+++ b/pkg/github/repositories.go
@@ -1592,7 +1592,58 @@ func GetTag(t translations.TranslationHelperFunc) inventory.ServerTool {
 				return ghErrors.NewGitHubAPIStatusErrorResponse(ctx, "failed to get tag reference", resp, body), nil, nil
 			}
 
-			// Then get the tag object
+			// Handle both annotated and lightweight tags.
+			// Annotated tags have ref.Object.Type == "tag" and point to a tag object.
+			// Lightweight tags have ref.Object.Type == "commit" and point directly to a commit.
+			if ref.Object.Type != nil && *ref.Object.Type == "commit" {
+				// Lightweight tag — resolve the commit directly
+				commit, resp, err := client.Git.GetCommit(ctx, owner, repo, *ref.Object.SHA)
+				if err != nil {
+					return ghErrors.NewGitHubAPIErrorResponse(ctx,
+						"failed to get commit for lightweight tag",
+						resp,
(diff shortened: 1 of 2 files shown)


SamMorrowDrums:
Thanks for this @fidesit, and sorry for the slow response!

The underlying 404 was fixed by #2400, which has since merged — `get_tag` no longer errors on lightweight tags. Closing as superseded.

That said, your version does something #2400 doesn't: it makes a second `Git.GetCommit()` call and populates `Message` + `Tagger` so the response shape stays consistent with annotated tags. That's a nice enrichment and not something we currently return. If you'd like to see that land, please open a fresh PR against current `main` — it'd be a smaller diff now that the base case is handled, and we can review it on its own merits.

Thanks again 🙏

PR #2160, opened 2026-03-06
Make MCP server name and title configurable

### LINKED ISSUE #2158: Make MCP Server `name` and `title` configurable

In our organization we would like to use the GitHub MCP Servers to connect to our GitHub Enterprise Server instances as well as to `github.com`.
For this use case we start multiple MCP server instances (`stdio` and `http`) which works fine in general.

From time to time though, the agents get confused between the running MCP servers, because all of the GitHub MCP servers identify with the same `name` and `title` in the `serverInfo` field of the Initialization Response. 
This leads to problems in the communication between the agent and the correct GitHub MCP Server.

### Proposed solution

It would help if the server `name` and `title` could be configured during the start of the server. Both an environment variable as well as a flag for the local server would be beneficial. PR #2160 would solve our issue.

### Additional context

Current response structure of the Initialization Response:
```
{
    "jsonrpc": "2.0",
    "id": "test",
    "result": {
        "capabilities": {
            "completions": {},
            "prompts": {},
            "resources": {},
            "tools": {}
        },
        "protocolVersion": "2024-11-05",
        "serverInfo": {
            "name": "github-mcp-server",
            "title": "GitHub MCP Server",
            "version": "v0.31.0",
            "icons": [
                ...
            ]
        }
    }
}
```

The fields `reult.serverInfo.name` and `result.serverInfo.title` are constant and currently hard coded [here](https://github.com/github/github-mcp-server/blob/c38802ac800fa1e037765c85bfc44f23a0f9f1bc/pkg/github/server.go#L187).

Anika-Sol:
## Summary
This PR makes the MCP server's `name` and `title` configurable during server start. The values can be set via flags (`server-name` and `server-title`) or environment variables (`GITHUB_SERVER_NAME` and `GITHUB_SERVER_TITLE`).

## Why

Fixes #2158

## What changed

- add the flags `server-name` and `server-title` with defaults resembling the previous hard coded name and title
- use the values of those flags when creating the MCP Server
- update documentation including an example

## MCP impact

- [x] No tool or API changes
- [ ] New tool added

## Security / limits

## Tool renaming

## Lint & tests

## Docs

- [ ] Not needed

### DIFF
--- a/.github/workflows/code-scanning.yml
+++ b/.github/workflows/code-scanning.yml
@@ -35,6 +35,10 @@ jobs:
             category: /language:go
             build-mode: autobuild
             runner: '["ubuntu-22.04"]'
+          - language: javascript
+            category: /language:javascript
+            build-mode: none
+            runner: '["ubuntu-22.04"]'
     steps:
       - name: Checkout repository
         uses: actions/checkout@v6
@@ -75,7 +79,7 @@ jobs:
           cache: false
 
       - name: Set up Node.js
-        if: matrix.language == 'go'
+        if: matrix.language == 'go' || matrix.language == 'javascript'
         uses: actions/setup-node@v4
         with:
           node-version: "20"
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -46,7 +46,7 @@ jobs:
       # https://github.com/sigstore/cosign-installer
       - name: Install cosign
         if: github.event_name != 'pull_request'
-        uses: sigstore/cosign-installer@faadad0cce49287aee09b3a48701e75088a2c6ad #v4.0.0
+        uses: sigstore/cosign-installer@ba7bc0a3fef59531c69a25acd34668d6d3fe6f22 #v4.1.0
         with:
           cosign-release: "v2.2.4"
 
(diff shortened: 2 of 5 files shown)


SamMorrowDrums:
Hi @Anika-Sol, I was actually looking at using the existing system we have for this: https://github.com/github/github-mcp-server/pull/2159

Seeing as we use it all over the app and it supports env, and JSON file for configuration.

Does this work for you?

Happy to list you as co-author.

Anika-Sol:
Hi @SamMorrowDrums, thank you for reaching out and checking my PR.
I tried out #2159 and it works for our use case as we would prefer environment variables to flags. So in terms of functionality it works well for us.

My only concern is the following:
Checking the documentation, the `Translator` that is used in the suggested implementation is only mentioned in terms of tool/toolset description customization. Without looking at the code I think users would not expect the `Translator` to also work for server configurations like `name` and `title`. Maybe it would be good to enhance the description in the `README.md`?

SamMorrowDrums:
Yeah readme update makes sense. There is a documented export function that creates a json dump of all the keys though, so everything is at least discoverable, but I will have a look at the docs.

Anika-Sol:
Closing in favour of #2159.

PR #2169, opened 2026-03-07
fix: return project item id usable for updates

### LINKED ISSUE #2153: projects_write: item_id parameter typed as float64 instead of string
## Bug Description

The `projects_write` tool has an ID mismatch between its `add_project_item` and `update_project_item` methods.

`add_project_item` returns a GraphQL node ID (string, e.g. `"PVTI_lADOBII0s..."`), but `update_project_item` expects a numeric REST API database ID (`item_id` typed as `number`, parsed via `RequiredBigInt`). This means the ID returned by one method cannot be directly used as input to the other.

## Steps to Reproduce

1. Call `projects_write` with `method: add_project_item` to add an issue to a project
2. The response returns an `id` like `"PVTI_lADODr_Ky84BNQJ7zgmwpT4"` (GraphQL node ID string)
3. Call `projects_write` with `method: update_project_item`, passing that `id` as `item_id`
4. The tool rejects it with: `parameter item_id is not of type float64`

## Root Cause

There are two ID systems at play:

| ID System | Type | Example | Used By |
|-----------|------|---------|---------|
| REST API `id` / `fullDatabaseId` | number (int64/BigInt) | `1001` | `update_project_item`'s `item_id` param |
| GraphQL node `id` | string (ID!) | `"PVTI_lADOBII0s..."` | `add_project_item`'s return value |

In the source (`pkg/github/projects.go`), `item_id` is defined as `Type: "number"` and parsed with `RequiredBigInt(args, "item_id")`, which is correct for the REST API's numeric ID. However, the return value from `add_project_item` appears to be the GraphQL node ID (string), creating a mismatch.

## Suggested Fix

One of:
1. **Preferred:** Have `add_project_item` also return the numeric `fullDatabaseId` so it can be used with `update_project_item`
2. Have `update_project_item` accept the GraphQL node ID and resolve it to the numeric ID internally
3. Document which ID format each method expects/returns, so callers know to look up the numeric ID separately

## Workaround

Using `gh project item-edit` via CLI works fine with the GraphQL node ID string.

## Environment

- MCP endpoint: `https://api.githubcopilot.com/mcp/`
- Toolset: `default,projects`
- Client: Claude Code via `mcp-remote`

njoppi2:
## Summary
- include `fullDatabaseId` in the GraphQL response for `projects_write` / `add_project_item`
- return a numeric `item_id` that can be passed directly to `update_project_item` and `delete_project_item`
- keep the existing node `id` in the response for backwards compatibility

Closes #2153.

## Validation
- ran `go test ./pkg/github -run "Test_ProjectsWrite_AddProjectItem|Test_ProjectsWrite_UpdateProjectItem"` in a Go 1.24 Docker container
- formatted touched files with `gofmt` in the same container


darthzoloft (approved):

SamMorrowDrums (approved):
Clean fix for #2153 — adds `fullDatabaseId` to the mutation and returns both `full_database_id` (string) and `item_id` (int64) so the result can be passed straight into `update_project_item` / `delete_project_item`. Existing node `id` preserved for backwards compat. Rebased onto main, tests pass, lint clean. Thanks!

Issue #2192, opened 2026-03-10
Add read-only regression coverage for dynamic toolsets and deprecated aliases


davidahmann:
Read-only guarantees currently depend on implicit inventory behavior that is not regression-tested across dynamic toolset enablement and deprecated alias resolution.

Current behavior is insufficient because the server's read-only contract spans multiple capability surfaces (`DynamicTools`, inventory filtering, and deprecated alias lookup), but there is no focused test that proves mutating canonical tools and deprecated aliases stay unavailable when read-only mode is active.

Why now: `--read-only` is a core governance boundary for this server, so the contract should be locked in with explicit regression coverage before toolset and alias work drifts further.

Claim-to-codepath map:
- Dynamic toolset registration in `pkg/github/dynamic_tools.go`
- Deprecated alias mapping in `pkg/github/deprecated_tool_aliases.go`
- Server inventory assembly in `internal/ghmcp/server.go`
- Request-level filtering tests in `pkg/http/handler_test.go`

Requested behavior:
- Read-only mode should continue to exclude mutating tools when a toolset is enabled dynamically.
- Deprecated aliases for mutating tools should not re-expose write capability under read-only filtering.
- The invariant should be covered by targeted tests.

## Evidence Packet
- Commit under test: `1da41fa6947f`
- Runtime environment:
  - OS: Darwin 25.3.0 arm64
  - Go: go1.25.7
  - golangci-lint: 2.8.0
- Minimal repro:
  1. Build inventory/server state with read-only mode enabled.
  2. Exercise dynamic toolset enablement and deprecated alias lookup.
  3. Verify whether any mutating capability becomes available.
- Expected behavior: no mutating canonical tool or deprecated alias becomes callable in read-only mode.
- Actual behavior: the code path is not covered by a focused regression test today.

## Acceptance Criteria
- Add targeted tests proving read-only behavior holds for dynamic toolsets and deprecated aliases.
- Keep the change scoped to regression coverage unless a concrete behavior bug is found.

jm809:
Electron [[ ending-support-major ]] .x.y has reached end-of-support as per the projects support policy developers and applications are encouraged to upgrade to a newer version of Electron.

@mutle

PR #2193, opened 2026-03-10
Add read-only regression coverage for aliases and toolsets

davidahmann:
Problem
Read-only filtering already covers dynamic toolsets and deprecated aliases, but there was no dedicated regression coverage for those two paths, so a future inventory change could silently re-expose write tools.

Why now
Issue #2192 narrowed this to test coverage after confirming the current runtime is already fail-closed. The gap is in preserving that contract.

What changed
- add a dedicated `ToolsForToolset` read-only regression test for dynamic toolset enablement
- add a deprecated-alias read-only regression test for `ForMCPRequest`

Validation
- `go test ./pkg/inventory -run 'TestToolsForToolset_RespectsReadOnly|TestForMCPRequest_ToolsCall_DeprecatedAliasRespectsReadOnly'`

Refs #2192

### DIFF
--- a/pkg/inventory/registry_test.go
+++ b/pkg/inventory/registry_test.go
@@ -481,6 +481,26 @@ func TestToolsForToolset(t *testing.T) {
 	}
 }
 
+func TestToolsForToolset_RespectsReadOnly(t *testing.T) {
+	tools := []ServerTool{
+		mockTool("tool_read", "toolset1", true),
+		mockTool("tool_write", "toolset1", false),
+	}
+
+	reg := mustBuild(
+		t,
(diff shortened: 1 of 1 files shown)


davidahmann:
User impact: This locks in the current fail-closed behavior so future inventory changes do not silently re-expose write tools through dynamic toolsets or deprecated aliases.
Change: The change is regression coverage only: one test for read-only `ToolsForToolset`, and one for deprecated alias filtering through `ForMCPRequest`.
Validation:
- `go test ./pkg/inventory -run 'TestToolsForToolset_RespectsReadOnly|TestForMCPRequest_ToolsCall_DeprecatedAliasRespectsReadOnly'` passed

Issue #2197, opened 2026-03-11
Inventory: return stable hidden-tool reason codes across filter gates


davidahmann:
## Problem
Tool filtering can hide capabilities due to toolset, PAT scope, read-only mode, or lockdown, but callers do not always receive one stable reason classification.

## Why now
Permission-scoped capability exposure is a core contract of the GitHub MCP server. Missing provenance for hidden tools weakens operator debugging and policy auditability.

## Current insufficiency
Filtering behavior is implemented in multiple layers, but there is not a single stable reason-code contract guaranteed across all hide paths.

## Expected behavior
Hidden tools should be classified with deterministic reason codes such as:
- `toolset_filtered`
- `scope_filtered`
- `readonly_filtered`
- `lockdown_filtered`

## Validation requirements
- Add tests that exercise each hide path.
- Assert one stable reason code per hidden tool decision.
- Keep classification convergent across inventory and request filtering.

## Scope map
- `internal/ghmcp/server.go`
- `pkg/inventory/builder.go`
- `pkg/scopes/fetcher.go`
- `pkg/lockdown/lockdown.go`

jm809:
Electron [[ ending-support-major ]] .x.y has reached end-of-support as per the projects support policy developers and applications are encouraged to upgrade to a newer version of Electron.

@davidahmann

PR #2198, opened 2026-03-11
inventory: expose stable hidden-tool reason codes

davidahmann:
## Problem
Tool filtering behavior was deterministic but did not expose stable machine-readable reason codes for why tools are hidden under read-only/toolset/feature-filter gates.

## Why now
Operators and downstream integrations need convergent filter diagnostics without relying on implicit behavior or log parsing.

## What changed
- Added stable `HiddenToolReason` codes and `HiddenTools(ctx)` API to return hidden tools with first-match reason.
- Refactored filter evaluation into a shared ordered path used by both `isToolEnabled` and `HiddenTools`.
- Added regression tests covering read-only, toolset, feature flag, builder-filter false/error paths.

## Validation
- `go test ./pkg/inventory -run HiddenTools -count=1`

Refs #2197

### DIFF
--- a/pkg/inventory/filters.go
+++ b/pkg/inventory/filters.go
@@ -13,6 +13,24 @@ import (
 // Returns (enabled, error). If error occurs, the caller should log and treat as false.
 type FeatureFlagChecker func(ctx context.Context, flagName string) (bool, error)
 
+// HiddenToolReason is a stable machine-readable reason for tool filtering decisions.
+type HiddenToolReason string
+
+const (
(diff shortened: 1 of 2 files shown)


davidahmann:
This change gives operators deterministic hidden-tool diagnostics across read-only, toolset, feature-flag, and builder-filter gates.
The patch is focused on inventory filtering: a shared ordered evaluator now powers both enablement and new hidden-tool reason reporting, with regression tests for each reason class.
Validation: `go test ./pkg/inventory -run HiddenTools -count=1` (pass).
Blockers/Risks: none currently identified.

PR #2200, opened 2026-03-11
fix(http): make --read-only flag work in HTTP mode

### LINKED ISSUE #2156: --read-only flag fails to restrict write tools when using the new http command
### Describe the bug

When running the GitHub MCP server (v0.31.0) using the new http command (streamable-http mode), the --read-only flag and the GITHUB_READ_ONLY=1 environment variable fail to restrict the server to read-only operations.

Modifying tools such as create_branch, create_pull_request, and merge_pull_request remain accessible in the tools/list response and are fully functional.
The behavior is identical whether testing locally with direct curl commands or when connecting an LLM client (e.g., LibreChat) to the server. In both scenarios, we successfully executed write actions despite the server being started with read-only enforcement.

### Affected version

v0.31.0

   1. Start the MCP server using Docker with the new http command and the --read-only flag:

       docker run -d --name mcp-github \
         --restart unless-stopped \
         -p 9002:8082 \
         -e GITHUB_PERSONAL_ACCESS_TOKEN="<VALID_PAT_WITH_REPO_SCOPE>" \
         ghcr.io/github/github-mcp-server:v0.31.0 \
         http  --read-only
   2. Verify the container is running and received the flag: docker inspect mcp-github | grep read-only
   3. Prove write access by directly calling the create_branch tool via curl:
   

curl -X POST http://localhost:9002/mcp \
  -H "Content-Type: application/json" \
  -H "Authorization: Bearer <VALID_PAT_WITH_REPO_SCOPE>" \
  -d '{
    "jsonrpc": "2.0",
    "method": "tools/call",
    "params": {
      "name": "create_branch",
      "arguments": {
        "owner": "<YOUR_ORG>",
        "repo": "<YOUR_REPO>",
        "branch": "test-branch-mcp-readonly-bypass"
      }
    },
    "id": 2
  }'
Observation: The request succeeds, and the branch test-branch-mcp-readonly-bypass is physically created in the target repository on GitHub.
  
  4. When connected to an LLM client (like LibreChat) via streamable-http, asking the agent to "Create a branch called 'test-branch'" also succeeds, confirming the LLM has full write access.

When the --read-only flag or GITHUB_READ_ONLY=1 environment variable is provided, the server should completely omit write/modifying tools from the tools/list response and block any incoming requests to execute those tools.

furkankoykiran:
## Summary
Fixes #2156 - The `--read-only` flag and `GITHUB_READ_ONLY` environment variable were being completely ignored in HTTP mode, allowing write tools like `create_branch`, `create_pull_request`, and `merge_pull_request` to remain accessible and functional even when read-only mode was explicitly requested.

## Changes
- Added `ReadOnly` field to `ServerConfig` struct in `pkg/http/server.go`
- Modified `httpCmd` in `cmd/github-mcp-server/main.go` to pass the `--read-only` flag value to the HTTP server configuration (matching stdio mode behavior)
- Created `withGlobalReadonly` middleware that enforces readonly mode globally across all HTTP requests
- Registered the global readonly middleware in the HTTP server's middleware chain
- Added comprehensive test coverage for the global readonly configuration

## Test Results
- All existing tests pass: `go test ./...` ✅
- Linter passes with 0 issues: `golangci-lint run` ✅
- New test `TestWithGlobalReadonly` verifies the middleware correctly applies readonly mode based on config
- Documentation updated via `script/generate-docs`

## Verification Steps
To verify the fix works:
```bash
docker run -d --name mcp-github \
  -p 9002:8082 \
  -e GITHUB_PERSONAL_ACCESS_TOKEN="<token>" \
  ghcr.io/github/github-mcp-server:<version> \
  http --read-only
```

Expected behavior:
- Write tools (`create_branch`, `create_pull_request`, etc.) should NOT appear in `tools/list` response
- Calling write tools via `tools/call` should fail with "tool not found"

### DIFF
--- a/cmd/github-mcp-server/main.go
+++ b/cmd/github-mcp-server/main.go
@@ -119,6 +119,7 @@ var (
 				LockdownMode:         viper.GetBool("lockdown-mode"),
 				RepoAccessCacheTTL:   &ttl,
 				ScopeChallenge:       viper.GetBool("scope-challenge"),
+				ReadOnly:             viper.GetBool("read-only"),
 			}
 
 			return ghhttp.RunHTTPServer(httpConfig)
--- a/pkg/http/handler_test.go
+++ b/pkg/http/handler_test.go
@@ -409,3 +409,51 @@ func TestHTTPHandlerRoutes(t *testing.T) {
 		})
 	}
 }
+
+func TestWithGlobalReadonly(t *testing.T) {
+	tests := []struct {
+		name           string
+		readonly       bool
+		expectReadonly bool
+	}{
+		{
+			name:           "readonly config enables readonly mode",
+			readonly:       true,
+			expectReadonly: true,
+		},
+		{
+			name:           "non-readonly config leaves readonly disabled",
+			readonly:       false,
+			expectReadonly: false,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			cfg := &ServerConfig{
+				Version:  "test",
+				ReadOnly: tt.readonly,
+			}
+
+			middleware := withGlobalReadonly(cfg)
+
+			// Create a handler that will check the context
+			var contextReadonly bool
+			next := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				contextReadonly = ghcontext.IsReadonly(r.Context())
+				w.WriteHeader(http.StatusOK)
+			})
+
+			// Apply middleware
+			handler := middleware(next)
+
+			// Create request and execute
+			req := httptest.NewRequest(http.MethodGet, "/", nil)
+			rr := httptest.NewRecorder()
+			handler.ServeHTTP(rr, req)
+
+			// Verify the readonly state matches the config
+			assert.Equal(t, tt.expectReadonly, contextReadonly, "readonly context should match config")
+		})
+	}
+}
--- a/pkg/http/server.go
+++ b/pkg/http/server.go
@@ -67,6 +67,9 @@ type ServerConfig struct {
 	// ScopeChallenge indicates if we should return OAuth scope challenges, and if we should perform
 	// tool filtering based on token scopes.
 	ScopeChallenge bool
+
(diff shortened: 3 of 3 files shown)


SamMorrowDrums:
Hi @furkankoykiran this is not a mistake. We need to make a decision if we want to allow global configuration like this, individual users can enable read only mode via headers, just like our deployed remote server.

You can simply add "x-mcp-readonly: true" header as per the remote server docs.

furkankoykiran:
@SamMorrowDrums Thanks for the feedback! Let me explain the rationale for global readonly:

**1. Consistency with stdio mode**
The `--read-only` flag already works in stdio mode (main.go:89). For consistency, HTTP mode should support the same configuration method. Users expect `github-mcp-server http --read-only` to behave like `github-mcp-server stdio --read-only`.

**2. Operational use cases**
There are valid scenarios where server-level readonly is preferable:
- **Multi-tenant deployments**: Default to readonly, opt-in to writes via configuration
- **Security layers**: Defense-in-depth where the server itself is configured as read-only
- **Client limitations**: Not all MCP clients can easily send custom headers

**3. Header + Config complement each other**
This isn't mutually exclusive. The global readonly could be the default, while headers can override it (or vice versa based on priority). This gives operators flexibility.

**4. Issue #2156 context**
The reporter specifically wanted Docker deployment with `--read-only` flag, similar to how they'd configure any long-running service. Expecting every client to send headers in this scenario is impractical.

Would you be open to this approach if we:
- Document that headers take precedence over global config?
- Add tests to verify header override behavior?
- Update docs to clarify both methods?

Open to suggestions on how to make this work!

SamMorrowDrums:
@furkankoykiran can you verify that https://github.com/github/github-mcp-server/pull/2208 does fix this for you as anticipated?

furkankoykiran:
Yep, #2208 covers it. The static inventory approach is cleaner than what I had — treating CLI flags as an upper bound that headers can only narrow is more principled than a middleware-only fix.

The Docker deployment case from #2156 should work exactly as expected now. Thanks for taking this further.

Issue #2203, opened 2026-03-12
Invalidate token-scope cache context when authorization token changes


davidahmann:
## Problem
Scope decisions can become stale if previously cached token scopes are reused after the active auth token changes, leading to incorrect tool visibility.

## Why now
Token churn is common in developer sessions. Scope filtering is a governance boundary, so stale reuse must fail closed and recompute for the active token.

## Expected behavior
When the active token differs from the token used to populate scope context, scope lookups must be recomputed and context updated for the new token.

## Claim-to-codepath map
- `pkg/context/token.go`
- `pkg/http/middleware/token.go`
- `pkg/http/middleware/pat_scope.go`
- `pkg/http/handler.go`

## Validation requirements
- Bind cached scopes to token identity in context helpers.
- Update PAT scope middleware/filter paths to ignore mismatched cached scopes.
- Add regression tests covering token swap with pre-populated scope context.

## Evidence packet
- Commit under test: `$(git -C /Users/davidahmann/Projects/agent-ecosystem/github-mcp-server rev-parse HEAD)`
- Runtime: Go test harness on macOS
- Minimal repro: request context preloaded with scopes for token A, then token B becomes active; filter should not reuse token A scopes.

SamMorrowDrums:
We don't do this. Our http/remote server checks scopes on every request without cache.

PR #2204, opened 2026-03-12
http: invalidate cached token scopes when auth token changes

davidahmann:
## Problem
If scope context populated for one token is reused after a token change, tool filtering can reflect stale permissions.

## Why now
Token rotation/swap during long-lived request flows is common; scope filtering must stay bound to the active token to avoid quiet governance drift.

## What changed
- Added token-bound scope context helpers:
  - `WithTokenScopesForToken`
  - `GetTokenScopesForToken`
- Updated middleware/filtering to reuse scopes only when bound token matches active token:
  - `pkg/http/middleware/pat_scope.go`
  - `pkg/http/middleware/scope_challenge.go`
  - `pkg/http/handler.go` (`PATScopeFilter`)
- Added regression tests:
  - `pkg/context/token_test.go`
  - token swap refetch case in `pkg/http/middleware/pat_scope_test.go`

## Validation
- `go test ./pkg/context ./pkg/http/middleware ./pkg/http`

Refs #2203

### DIFF
--- a/pkg/context/token.go
+++ b/pkg/context/token.go
@@ -28,15 +28,47 @@ func GetTokenInfo(ctx context.Context) (*TokenInfo, bool) {
 
 type tokenScopesKey struct{}
 
+type tokenScopesValue struct {
+	Token  string
+	Scopes []string
+}
+
 // WithTokenScopes adds token scopes to the context
 func WithTokenScopes(ctx context.Context, scopes []string) context.Context {
 	return context.WithValue(ctx, tokenScopesKey{}, scopes)
 }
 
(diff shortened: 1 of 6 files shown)


davidahmann:
This PR prevents stale scope-filter decisions when auth tokens change by requiring cached scope context to match the active token before reuse.

The fix is intentionally small: add token-bound scope helpers, switch PAT/scope-challenge/filter paths to token-bound reads, and add regression tests for token-swap behavior.

Validation:
- `go test ./pkg/context ./pkg/http/middleware ./pkg/http` (pass)

SamMorrowDrums:
We don't cache token scopes. Not sure what this is fixing.

PR #2230, opened 2026-03-17
Hide write-only prompt guidance when tools are filtered

### LINKED ISSUE #2228: Hide write-oriented prompts and instructions when write tools are filtered
## Problem
Tool filtering can remove write capabilities while prompts and server instructions continue to advertise those same actions. In read-only or explicitly filtered configurations, the server can still expose prompts that instruct the client to create issues, assign Copilot, or open pull requests.

## Why now
This server builds instructions and prompts dynamically as part of its safety surface. If filtered capabilities remain documented as available, clients receive contradictory guidance and can plan actions the server has already hidden.

## Version under test
- Commit: `b1575edfefde09e3cf7c805aea79a92131271659` (`origin/main` on 2026-03-17)
- Runtime: Go `1.25.7`, macOS `darwin/arm64`

## Minimal repro
1. Build an inventory in read-only mode or with write tools excluded.
2. Inspect `AvailableTools`, `AvailablePrompts`, and generated server instructions.
3. Compare the filtered tool set with the exposed prompts/instruction text.

## Expected behavior
If a write tool is hidden by policy, related prompts and write-oriented instructions should disappear or clearly mark the capability as unavailable.

## Actual behavior
Prompt availability is filtered only by toolset/feature flags, and instruction generation does not verify that the referenced write tools survived read-only or exclude-tool filtering.

## Code paths
- `pkg/inventory/instructions.go`
- `pkg/inventory/prompts.go`
- `pkg/github/toolset_instructions.go`
- `docs/toolsets-and-icons.md`

## Validation target
Read-only and explicitly filtered inventories should not expose prompts or instruction text that imply unavailable write actions.

davidahmann:
Closes #2228

## Problem
Read-only and filtered inventories still expose issue/PR write instructions and workflow prompts even when the write tools themselves are unavailable.

## What changed
- add required-tool metadata for prompts and filter those prompts when the required tools are unavailable
- make issue and pull-request instructions conditional on the matching write tools surviving inventory filtering
- add regression coverage for read-only and excluded-tool inventories

## Validation
- `go test ./pkg/inventory ./pkg/github`


davidahmann:
This removes write-path guidance from filtered and read-only inventories, so users stop seeing issue/PR instructions that the current tool set cannot actually execute.

The change adds required-tool gating for prompts and makes the issue / pull-request instruction blocks conditional on the surviving write tools.

Validation:
- `go test ./pkg/inventory ./pkg/github` (pass)

Issue #2229, opened 2026-03-17
feat: add `list_pull_request_review_threads` tool to expose PRRT_ node IDs for resolving threads


scottlz0310-user:
## Problem

`resolve_pull_request_review_thread` requires a thread node ID in `PRRT_xxx` format. However, this ID is **not returned by any existing REST-based MCP tool** — it is only obtainable via the GraphQL API.

Currently there is no way to complete the "list unresolved threads → resolve them" flow using the MCP tools alone, because there is no tool to retrieve the `PRRT_xxx` node IDs needed by `resolve_pull_request_review_thread`.

## Proposed Solution

Add a `list_pull_request_review_threads` tool that uses the existing `githubv4` GraphQL client (already a dependency) to query review threads and return their node IDs.

### Tool Definition

**Name:** `list_pull_request_review_threads`

**Input:**
| Parameter | Type | Required | Description |
|-----------|------|----------|-------------|
| `owner` | string | ✅ | Repository owner |
| `repo` | string | ✅ | Repository name |
| `pull_number` | integer | ✅ | Pull request number |
| `is_resolved` | boolean | ❌ | Filter by resolution status. Omit to return all threads. |

**Output (per thread):**
| Field | Type | Description |
|-------|------|-------------|
| `id` | string | Node ID in `PRRT_xxx` format — required for `resolve_pull_request_review_thread` |
| `isResolved` | boolean | Whether the thread has been resolved |
| `isOutdated` | boolean | Whether the thread is outdated (code changed after the comment) |
| `path` | string | File path the thread is on |
| `line` | integer | End line number |
| `startLine` | integer | Start line number |
| `firstComment.body` | string | Body of the first comment in the thread |
| `firstComment.author` | string | Login of the commenter |
| `firstComment.createdAt` | string | ISO 8601 timestamp |

**Response envelope:**
```json
{
  "threads": [...],
  "totalCount": 4,
  "truncated": false
}
```

### Suggested GraphQL Query

```graphql
query($owner: String!, $repo: String!, $number: Int!) {
  repository(owner: $owner, name: $repo) {
    pullRequest(number: $number) {
      reviewThreads(first: 100) {
        totalCount
        nodes {
          id
          isResolved
          isOutdated
          path
          line
          startLine
          comments(first: 1) {
            nodes {
              body
              author { login }
              createdAt
            }
          }
        }
      }
    }
  }
}
```

### Registration

[...]

PR #1919, opened 2026-01-28
Feat/resolve review threads

plwalters:
## Summary

Adds `resolve_thread` and `unresolve_thread` methods to the `pull_request_review_write` tool, enabling users to resolve and unresolve PR review threads via GraphQL mutations.

## Why

Fixes #1768

## What changed

- Added `ThreadID` field to `PullRequestReviewWriteParams` struct
- Added `threadId` parameter and new methods to tool schema
- Implemented `ResolveReviewThread` function using GraphQL `resolveReviewThread` and `unresolveReviewThread` mutations
- Added switch cases for `resolve_thread` and `unresolve_thread` methods
- Added unit tests covering success and error scenarios
- Updated toolsnaps and generated docs

## MCP impact

- [ ] No tool or API changes
- [x] Tool schema or behavior changed — Added two new methods (`resolve_thread`, `unresolve_thread`) and a new `threadId` parameter to the existing `pull_request_review_write` tool.
- [ ] New tool added

## Prompts tested (tool changes only)

- "Resolve the review thread PRRT_kwDOxxx on PR #123"
- "Unresolve the feedback thread on this pull request"
- "Mark the review comment thread as resolved"

## Security / limits

- [x] No security or limits impact — Uses existing GraphQL client authentication; thread resolution is already permission-controlled by GitHub.
- [ ] Auth / permissions considered
- [ ] Data exposure, filtering, or token/size limits considered

## Tool renaming
- [ ] I am renaming tools as part of this PR (e.g. a part of a consolidation effort)
   - [ ] I have added the new tool aliases in `deprecated_tool_aliases.go` 
- [x] I am not renaming tools as part of this PR

Note: if you're renaming tools, you *must* add the tool aliases. For more information on how to do so, please refer to the [official docs](https://github.com/github/github-mcp-server/blob/main/docs/tool-renaming.md).

## Lint & tests

- [x] Linted locally with `./script/lint`
- [x] Tested locally with `./script/test`

## Docs

- [ ] Not needed
- [x] Updated (README / docs / examples) — Auto-generated via `script/generate-docs`

PR #2245, opened 2026-03-20
fix: restore thread id in get_review_comments response

### LINKED ISSUE #2331: infra: expose review thread node ID in get_review_comments response (follow-up to #1768)
## Background

After #1768 shipped, `resolve_review_thread` is available as an MCP tool. However, the companion read path — retrieving a review thread's GraphQL node ID so it can be passed to `resolve_review_thread` — is still missing from the response schema.

When an AI agent is addressing PR review feedback, the typical flow is:

1. Read review comments with `pull_request_read method=get_review_comments`.
2. Reply to the comment with `add_reply_to_pull_request_comment`.
3. Resolve the thread with `resolve_review_thread`.

Step 3 requires the thread's node ID (e.g. `PRRT_kwDO...`), but step 1's response does not expose it.

## Problem

`pull_request_read method=get_review_comments` returns each `review_threads[]` object with these fields only:

```json
{
  "is_resolved": false,
  "is_outdated": false,
  "is_collapsed": false,
  "comments": [...],
  "total_count": N
}
```

There is **no `id` / `node_id` field on the thread object**, so there is no way to obtain the value `resolve_review_thread` requires.

Auxiliary paths are also blocked:

- `add_reply_to_pull_request_comment` returns the *comment*'s node ID (`PRRC_...`), which is distinct from the thread's node ID.
- `pageInfo.startCursor` / `endCursor` are pagination cursors, not thread IDs.
- Within an MCP-only session there is no `gh` CLI or direct GraphQL escape hatch.

Note: issue #1768 stated that "the thread IDs are already returned by `get_review_comments` in the `ID` field". That does not appear to match the current shipped response schema — the `ID` field is absent on the thread object as of 2026-04-15.

## Reproduction

On any PR with multiple unresolved review threads, calling `pull_request_read method=get_review_comments` returns no field usable as the `thread_id` argument of `resolve_review_thread`. This was observed while addressing Copilot PR reviewer feedback on a PR with 5 unresolved threads — the workflow had to fall back to manual `gh api graphql` calls outside MCP to complete step 3.

Minimal repro outside MCP — compare the current MCP response shape to this GraphQL query, which does expose the thread `id`:

```graphql
{
  repository(owner: "<owner>", name: "<repo>") {
    pullRequest(number: <n>) {
      reviewThreads(first: 50) {
        nodes { id isResolved isOutdated }
      }
    }
  }
}
```

The MCP response should surface the same `id` field on each thread.

## Proposal

Either:

1. **Extend the `get_review_comments` response schema** — include `PullRequestReviewThread.id` (GraphQL node ID) as `id` or `node_id` on each thread object.
2. **Add a dedicated method** — `pull_request_read method=get_review_threads` that returns a thin list of threads with their node IDs (and the existing `is_resolved` / `is_outdated` / `is_collapsed` flags).

Option 1 is likely a single-field addition: the underlying GraphQL query almost certainly already selects the thread object (since `isResolved`/`isOutdated`/`isCollapsed` are exposed), so adding `id` to the same selection set should be a minimal change.

## Current workaround

For MCP-only sessions, `resolve_review_thread` is effectively unreachable — users must drop out of MCP and click "Resolve conversation" in the GitHub web UI (or run `gh api graphql` outside the session). This defeats the purpose of shipping the tool in #1768.

## References

- #1768 — original request that added `resolve_review_thread`
- #1086 — `is_resolved` exposure on comments
- MCP tool list snapshot date: 2026-04-15

omid-ant:
## Summary
Restore the thread `id` field to `MinimalReviewThread` so that `pull_request_review_write` with `method: resolve_thread` / `unresolve_thread` is actually usable. 

## Why

#2062 dropped `id` from the minimized review-thread output, noting it was "not used for any subsequent tool call". While this was true at the time, #1919 later added `resolve_thread` / `unresolve_thread`, both of which require a thread node ID. The tool's own docstring reads: 

> Get thread IDs from `pull_request_read` with method `get_review_comments`.

…but that tool no longer returns them, so there's no way to resolve a thread without leaving the MCP server. **This results in callers of this tool being unable to resolve threads through the `resolve_thread` tool**. 

The GraphQL query (`reviewThreadNode` in `pkg/github/pullrequests.go`) already fetches `ID` — it's just dropped during minimization.

Also see: #2229

## What changed
- Add `ID string` to `MinimalReviewThread`
- Populate it in `convertToMinimalReviewThread` via `fmt.Sprintf("%v", thread.ID)` 

## MCP impact
- [ ] No tool or API changes
- [x] Tool schema or behavior changed — `pull_request_read` (`get_review_comments`) now includes `id` per thread
- [ ] New tool added

- "List the review threads on PR #42 in owner/repo, then resolve the first one"

## Security / limits
- [x] Data exposure, filtering, or token/size limits considered — ~40 chars/thread (`"id":"PRRT_kwDO…",`), negligible impact on #2062's 14.5% savings

## Tool renaming

## Docs
- [x] Not needed

Fixes #2331

### DIFF
--- a/pkg/github/minimal_types.go
+++ b/pkg/github/minimal_types.go
@@ -1547,7 +1547,7 @@ type MinimalReviewComment struct {
 
 // MinimalReviewThread is the trimmed output type for PR review thread objects.
 type MinimalReviewThread struct {
-	ID          string
+	ID          string                 `json:"id"`
 	IsResolved  bool                   `json:"is_resolved"`
 	IsOutdated  bool                   `json:"is_outdated"`
 	IsCollapsed bool                   `json:"is_collapsed"`


MitchTalmadge:
This is a blocker to our workflow. Currently we are unable to resolve threads automatically through MCP using Claude, for example, because it does not know the IDs of the threads. Excited to see this PR and hope it can be approved soon.

SamMorrowDrums:
Tracks #2331.

scottlz0310-user:
Supporting this PR. This is exactly the gap that #2229 identified — 
the `id` field being absent made `resolve_thread` unreachable from 
an MCP-only session. Would be great to see this merged.

MitchTalmadge (approved):

scottlz0310-user (approved):
As the reporter of #2229, I checked out this branch, built it locally, and verified the
behavior.

Verified:

 - The id field now correctly appears in the JSON response from get_review_comments 
(e.g., "id": "PRRT_kwDO...")
 - All existing tests pass (go test ./...)

One suggestion: the existing test for get_review_comments already provides "id" in the
mock data but never asserts thread.ID in the validation. Adding an assertion like 
assert.Equal(t, "RT_kwDOA0xdyM4AX1Yz", thread.ID) would prevent silent regressions in
the future.

Phylu:
Is there anything that can be done to push this forward? I ran into the exact same issue just now, trying to resolve a PR thread via ai agent and was not able to without relying on `gh` to fetch the id via graphql.

SamMorrowDrums (dismissed):

SamMorrowDrums:
@omid-ant @Phylu aplogogies for delay, let's gets this merged ASAP. Can you fix merge conflicts?

SamMorrowDrums (approved):

PR #2306, opened 2026-04-08
OSS granular PRs and issues toolsets

mattdholloway:
## Summary

Add two new opt-in, non-default toolsets — `issues_granular` and `pull_requests_granular` — that provide single-purpose write tools decomposing the existing multi-method tools (`issue_write`, `sub_issue_write`, `update_pull_request`, `pull_request_review_write`).

This gives users fine-grained control over which write operations are available, which is especially useful for AI agents where limiting the tool surface area improves reliability and safety.

Closes https://github.com/github/sweagentd/issues/11219

## Why

The existing consolidated tools use a `method` parameter to control behavior (e.g. `issue_write` with `method: "create"` or `method: "update"`). This makes it impossible to selectively enable individual operations — you either get all of `issue_write` or none of it.

Granular toolsets solve this by exposing each operation as its own tool, allowing users to enable exactly the capabilities they need.

### New toolsets

**`issues_granular`** (11 tools):

| Tool | Decomposed from |
|------|----------------|
| `create_issue` | `issue_write` method:create |
| `update_issue_title` | `issue_write` method:update |
| `update_issue_body` | `issue_write` method:update |
| `update_issue_state` | `issue_write` method:update |
| `update_issue_labels` | `issue_write` method:update |
| `update_issue_assignees` | `issue_write` method:update |
| `update_issue_milestone` | `issue_write` method:update |
| `update_issue_type` | `issue_write` method:update |
| `add_sub_issue` | `sub_issue_write` method:add |
| `remove_sub_issue` | `sub_issue_write` method:remove |
| `reprioritize_sub_issue` | `sub_issue_write` method:reprioritize |

**`pull_requests_granular`** (9 tools):

| Tool | Decomposed from |
|------|----------------|
| `update_pull_request_title` | `update_pull_request` |
| `update_pull_request_body` | `update_pull_request` |
| `update_pull_request_state` | `update_pull_request` |
| `update_pull_request_draft_state` | New (GraphQL-based draft toggle) |
| `request_pull_request_reviewers` | `update_pull_request` (reviewers field) |
| `create_pull_request_review` | `pull_request_review_write` method:create |
| `submit_pending_pull_request_review` | `pull_request_review_write` method:submit_pending |
| `delete_pending_pull_request_review` | `pull_request_review_write` method:delete_pending |
| `add_pull_request_review_comment` | `pull_request_review_write` (comment to pending review) |

## MCP impact

- [ ] No tool or API changes
- [x] New tool added

- `--toolsets=default,issues_granular,pull_requests_granular` enables both granular toolsets alongside defaults
- `--toolsets=issues_granular` enables only granular issue tools

## Security / limits

All granular tools require the same `repo` scope as their parent tools.

## Tool renaming

## Lint & tests

## Docs

- [x] Not needed


SamMorrowDrums:
@mattdholloway Hey Matt, I pushed two commits to this branch with some refactoring:

**What changed:**

1. **Reuse existing helpers instead of reimplementing** — The granular sub-issue tools now delegate to the existing `AddSubIssue`/`RemoveSubIssue`/`ReprioritizeSubIssue` REST helpers, and the PR review tools delegate to `CreatePullRequestReview`/`SubmitPendingPullRequestReview`/`DeletePendingPullRequestReview`. This fixes the viewer filtering bug flagged in review (where `reviews(first: 1, states: PENDING)` could pick another user's review), fixes the milestone param to use `RequiredInt`, and fixes line/startLine to use `OptionalIntParam`. Also removes the duplicate GraphQL types and helpers that are no longer needed.

2. **Feature flags instead of separate toolsets** — Rather than creating new `issues_granular`/`pull_requests_granular` toolset IDs, the granular tools now live in the existing `issues`/`pull_requests` toolsets with `FeatureFlagEnable` set. This uses the existing feature flag infrastructure to switch between consolidated and granular tool variants at runtime, and avoids the complexity of separate toolsets, read tool duplication, and name collision handling.

**Why:**

- The codebase already has well-tested helpers for all the operations the granular tools need — reimplementing them introduced bugs and divergent behavior.
- Feature flags are the right mechanism for tool variants (same pattern as the actions consolidation). OSS users can pass feature flags, and the HTTP server will support them too. This keeps the toolset surface simple while giving runtime control over which variant is active.
- Added toolsnap tests for all 20 granular tools.

Net effect is ~190 fewer lines than the original while fixing all review comments.

pkg/github/issues_granular.go
```diff
@@ -0,0 +1,594 @@
...
+			assignees, err := OptionalStringArrayParam(args, "assignees")
+			if err != nil {
+				return nil, err
+			}
+			if len(assignees) == 0 {
+				return nil, fmt.Errorf("missing required parameter: assignees")
+			}
+			return &github.IssueRequest{Assignees: &assignees}, nil
+		},
+	)
```
SamMorrowDrums, line 257: `OptionalStringArrayParam` returns `[]string{}` for both a missing key and an explicitly empty array, so this `len == 0` check can't distinguish between the two. This means users can't clear all assignees from an issue, despite the description saying "replaces the current assignees with the provided list."

Same issue exists for `GranularUpdateIssueLabels` below.

pkg/github/pullrequests_granular.go
```diff
@@ -0,0 +1,719 @@
...
+			})
+			return result, nil, err
+		},
+	)
+	st.FeatureFlagEnable = FeatureFlagPullRequestsGranular
+	return st
+}
+
+// GranularAddPullRequestReviewComment creates a tool to add a review comment.
+func GranularAddPullRequestReviewComment(t translations.TranslationHelperFunc) inventory.ServerTool {
```
SamMorrowDrums, line 543: This duplicates ~80 lines of GraphQL logic from `AddCommentToPendingReview` (viewer query → pending review lookup → mutation). The other granular PR review tools properly call shared helpers (`CreatePullRequestReview`, `SubmitPendingPullRequestReview`, `DeletePendingPullRequestReview`) — this one should follow the same pattern.

Also, `add_comment_to_pending_review` is already a single-purpose tool on main (no method dispatching). This PR disables it and replaces it with `add_pull_request_review_comment` which is essentially a copy with a new name. Consider just keeping the existing tool or extracting the shared logic into a helper.

pkg/github/issues_granular.go
```diff
@@ -0,0 +1,594 @@
...
+		},
+	}
+	maps.Copy(props, extraProps)
+
+	required := append([]string{"owner", "repo", "issue_number"}, extraRequired...)
+
+	st := NewTool(
+		ToolsetMetadataIssues,
+		mcp.Tool{
+			Name:        name,
```
SamMorrowDrums, line 50: This constructs translation keys like `TOOL_update_issue_title_DESCRIPTION` (mixed case), while every other tool in the codebase uses ALL_CAPS keys like `TOOL_ISSUE_READ_DESCRIPTION`. Will silently fail to match when real translations are added. Same issue in `prUpdateTool`.

pkg/github/pullrequests.go
```diff
@@ -1607,6 +1609,8 @@ Available methods:
 				return utils.NewToolResultError(fmt.Sprintf("unknown method: %s", params.Method)), nil, nil
 			}
 		})
+	st.FeatureFlagDisable = FeatureFlagPullRequestsGranular
```
SamMorrowDrums, line 1612: `pull_request_review_write` includes `resolve_thread` and `unresolve_thread` methods, but no granular replacements are added for these. When the `pull_requests_granular` flag is enabled, this consolidated tool is disabled and those two operations become inaccessible.

mattdholloway: This is a good point, these exist outside the scope of the original PR by @timrogers so I didn't add them. Do we deem this important enough to rectify?

SamMorrowDrums: I am churning on some of them now.

SamMorrowDrums: The problem is they can't be done in granular FF otherwise.

omgitsads (approved):

PR #2308, opened 2026-04-09
fix: preserve angle brackets inside fenced code blocks during sanitization

### LINKED ISSUE #2202: GitHub MCP issue read appears to drop code block text in angle brackets
### Describe the bug

When the AI agent tries to read an issue from my private repository via the GitHub MCP issue-read endpoint, any code between two
`` ``` ``
lines that contains text in angle brackets seems to get lost.
Take this example:

```
let ptr: mut_raw_ptr<int> = raw_new int;
```

The agent sees only this:

```
let ptr: mut_raw_ptr = raw_new int;
```

### Affected version

I am using the online version of the MCP server.

Environment:
- VS Code 1.111.0
- Agent: GPT-5.3-Codex    (also tried with Claude Sonnet 4.6)
- Date observed: 2026-03-12
- MCP tool used: mcp_github_issue_read

I typed this in the chat window:

> I want to check whether the GitHub MCP server for reading issues works correctly. Could you therefore please read issue #... and report the text of first code example back to me.

The text that was reported back did not contain the `<int>` template parameter.

Image

The AI agent should be able to see text between angle brackets if it appears in quoted text.

### Logs

I have added the chat debug log.

[copilot_all_prompts_2026-03-12T11-08-00.chatreplay.zip](https://github.com/user-attachments/files/25933024/copilot_all_prompts_2026-03-12T11-08-00.chatreplay.zip)

lawrence3699:
## Summary

`FilterHTMLTags` uses `bluemonday.StrictPolicy()` which strips anything that looks like an HTML tag — including `<int>`, `<T>`, `<string>`, etc. inside fenced code blocks. When a user files an issue containing C++/Rust/Java generics in a code block, the MCP tool returns the code with those tokens silently removed.

## Root cause

`Sanitize()` feeds the entire text (including code block content) through bluemonday, which has no concept of markdown structure. A line like `vector<int> v;` inside a fenced block gets `<int>` stripped because bluemonday interprets it as an unknown HTML tag.

Before:
```
Input:  "```\nlet ptr: mut_raw_ptr<int> = raw_new int;\n```"
Output: "```\nlet ptr: mut_raw_ptr = raw_new int;\n```"
```

After this fix:
```
Input:  "```\nlet ptr: mut_raw_ptr<int> = raw_new int;\n```"
Output: "```\nlet ptr: mut_raw_ptr<int> = raw_new int;\n```"
```

## Approach

Added a protect/restore step around `FilterHTMLTags` in the `Sanitize` pipeline:

1. **`protectCodeAngles`** — walks the text line-by-line tracking fenced code block state (same fence detection pattern used by `FilterCodeFenceMetadata`), and replaces `<`/`>` inside code blocks with NUL-delimited placeholders that bluemonday passes through
2. **`restoreCodeAngles`** — reverses the placeholders after sanitization

The closing fence check uses `>=` (per CommonMark spec) rather than exact-match, so a longer closing fence correctly ends the code block and does not leak placeholder-protection into subsequent content.

Angle brackets *outside* code blocks continue to be sanitized normally — `<script>alert(1)</script>` after a code block is still stripped.

## Tests

- 10 unit tests for `protectCodeAngles` covering: empty input, no code blocks, language-tagged fences, multiple blocks, 4-backtick fences, shorter/longer closing fences, unclosed fences
- 3 unit tests for `restoreCodeAngles`
- 7 end-to-end `Sanitize` tests including the exact scenario from #2202, nested generics (`Map<String, List<Integer>>`), and security regression tests confirming `<script>` tags after code blocks are still stripped

All existing tests continue to pass. `go vet` and `golangci-lint` clean.

## Known limitation

Tilde-fenced blocks (`~~~`) are not handled, consistent with the existing `FilterCodeFenceMetadata` which also only recognizes backtick fences.

Fixes #2202

### DIFF
--- a/pkg/sanitize/sanitize.go
+++ b/pkg/sanitize/sanitize.go
@@ -12,7 +12,12 @@ var policy *bluemonday.Policy
 var policyOnce sync.Once
 
 func Sanitize(input string) string {
-	return FilterHTMLTags(FilterCodeFenceMetadata(FilterInvisibleCharacters(input)))
+	s := FilterInvisibleCharacters(input)
+	s = FilterCodeFenceMetadata(s)
+	s = protectCodeAngles(s)
+	s = FilterHTMLTags(s)
+	s = restoreCodeAngles(s)
+	return s
 }
 
 // FilterInvisibleCharacters removes invisible or control characters that should not appear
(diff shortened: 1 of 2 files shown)


SamMorrowDrums:
@lawrence3699 no need to close if it's causing you problems.

Are you having issues with agent getting confused as a consequence?

PR #2334, opened 2026-04-15
Add middleware to normalize Content-Type header

### LINKED ISSUE #2333: 'Content-Type must be 'application/json'  - when setting charset
### Describe the bug

I'm using the https://github.com/modelcontextprotocol/csharp-sdk for MCP integration and when sending the initial POST it adds the Content-Type as "application/json; charset=utf-8" and the server replies with `'Content-Type must be 'application/json' `, while reduntant due to RFC 8259, it should just be stripped instead of enforced. This was working previously working.

### Affected version

Remote version of Github MCP Server.

Working:

Image

Non-Working:

Image

Expected -> ability to initialize the McpClient
Actual -> throws exceptions

utafrali:
Fixes #2333

## Summary
Added middleware that normalizes the Content-Type header by stripping MIME parameters like charset.

## Why
The API was rejecting requests with `Content-Type: application/json; charset=utf-8` because downstream validation expected exactly `application/json`. This middleware strips those extra parameters so handlers see the clean media type.

## What changed
Created a new middleware that uses mime.ParseMediaType to extract just the media type, then rewrites the header if parameters were present. Registered it early in the handler's middleware chain so all downstream handlers get the normalized version. Included tests covering cases with no parameters, charset parameter, multiple parameters, and empty headers.

## MCP impact
MCP clients that send Content-Type headers with charset parameters will now work correctly. The normalized header is transparent to downstream handlers and doesn't change behavior for anything else.

## Prompts tested
N/A

## Security/limits consideration
No security implications. The middleware only parses and normalizes the header, it doesn't validate or reject anything.

## Tool renaming considerations
N/A

## Checklist
- [x] Ran `./script/lint`
- [x] Ran `./script/test`
- [ ] Updated docs

### DIFF
--- a/pkg/http/handler.go
+++ b/pkg/http/handler.go
@@ -127,6 +127,7 @@ func NewHTTPMcpHandler(
 
 func (h *Handler) RegisterMiddleware(r chi.Router) {
 	r.Use(
+		middleware.NormalizeContentType,
 		middleware.ExtractUserToken(h.oauthCfg),
 		middleware.WithRequestConfig,
 		middleware.WithMCPParse(),
--- a/pkg/http/middleware/content_type.go
+++ b/pkg/http/middleware/content_type.go
@@ -1,0 +1,23 @@
+package middleware
+
+import (
+	"mime"
+	"net/http"
+
+	"github.com/github/github-mcp-server/pkg/http/headers"
+)
+
+// NormalizeContentType strips MIME parameters from the Content-Type header so
+// "application/json; charset=utf-8" is treated as "application/json".
+func NormalizeContentType(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if ct := r.Header.Get(headers.ContentTypeHeader); ct != "" {
+			mediaType, _, err := mime.ParseMediaType(ct)
+			if err == nil && mediaType != ct {
+				r = r.Clone(r.Context())
+				r.Header.Set(headers.ContentTypeHeader, mediaType)
+			}
+		}
+		next.ServeHTTP(w, r)
+	})
+}
--- a/pkg/http/middleware/content_type_test.go
+++ b/pkg/http/middleware/content_type_test.go
@@ -1,0 +1,61 @@
+package middleware
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/github/github-mcp-server/pkg/http/headers"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestNormalizeContentType(t *testing.T) {
+	tests := []struct {
+		name     string
+		input    string
+		expected string
+	}{
+		{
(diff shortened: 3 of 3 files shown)


SamMorrowDrums:
Thanks for looking into this! I wanted to flag that the go-sdk already handles `Content-Type: application/json; charset=utf-8` correctly — the `StreamableHTTPHandler` uses [`mime.ParseMediaType`](https://github.com/modelcontextprotocol/go-sdk/blob/4c00594be629ca87637f4316a232377f9b0979cd/mcp/streamable.go#L268-L270) to extract the media type before comparing, so `charset=utf-8` (and other MIME parameters) are already stripped at the SDK level:

```go
mediaType, _, err := mime.ParseMediaType(req.Header.Get("Content-Type"))
if err != nil || mediaType != "application/json" {
    http.Error(w, "Content-Type must be 'application/json'", http.StatusUnsupportedMediaType)
    return
}
```

This Content-Type validation was added in go-sdk v1.4.1 as part of cross-origin protection. The current OSS repo uses go-sdk v1.3.1-pre which predates this check entirely (so there's no Content-Type validation to normalize for). When the go-sdk dependency is updated, `mime.ParseMediaType` will handle `charset=utf-8` correctly without needing additional middleware.

@utafrali — are you currently able to reproduce this issue? If so, could you share what endpoint you're hitting and the exact error response? That would help us understand if there's something else going on beyond the Content-Type header parsing.

SamMorrowDrums:
Thank you, let me know if your issue is not resolved.

CZPAYPrime:
감사해요


CUIQINGHUA


2026년 4월 16일 (목) 08:25, Sam Morrow ***@***.***>님이 작성:

utafrali:
Got it, will let you know if I can still reproduce it.

fiiiiiiiiiuzzz:
> Fixes #2333
> 
> ## Summary
> Added middleware that normalizes the Content-Type header by stripping MIME parameters like charset.
> 
> ## Why
> The API was rejecting requests with `Content-Type: application/json; charset=utf-8` because downstream validation expected exactly `application/json`. This middleware strips those extra parameters so handlers see the clean media type.
> 
> ## What changed
> Created a new middleware that uses mime.ParseMediaType to extract just the media type, then rewrites the header if parameters were present. Registered it early in the handler's middleware chain so all downstream handlers get the normalized version. Included tests covering cases with no parameters, charset parameter, multiple parameters, and empty headers.
> 
> ## MCP impact
> MCP clients that send Content-Type headers with charset parameters will now work correctly. The normalized header is transparent to downstream handlers and doesn't change behavior for anything else.
> 
> ## Prompts tested
> N/A
> 
> ## Security/limits consideration
> No security implications. The middleware only parses and normalizes the header, it doesn't validate or reject anything.
> 
> ## Tool renaming considerations
> N/A
> 
> ## Checklist
> - [x] Ran `./script/lint`
> - [x] Ran `./script/test`
> - [ ] Updated docs

PR #2336, opened 2026-04-15
deps: upgrade modelcontextprotocol/go-sdk to v1.5.0

### LINKED ISSUE #2333: 'Content-Type must be 'application/json'  - when setting charset

SamMorrowDrums:
## Summary
Upgrade the MCP Go SDK from `v1.3.1-0.20260220105450-b17143f71798` (pseudo-version) to **v1.5.0** (latest stable).

Fixes #2333

## Why
The pinned pseudo-version predates go-sdk v1.4.1, which introduced Content-Type validation in `StreamableHTTPHandler`. That validation already uses `mime.ParseMediaType` to correctly handle `Content-Type: application/json; charset=utf-8`, so the middleware proposed in #2334 is unnecessary with this upgrade.

## What changed
- `go.mod` / `go.sum`: bumped `modelcontextprotocol/go-sdk` to v1.5.0
- Transitive bumps: `go` directive 1.24→1.25, `golang.org/x/oauth2`, `golang.org/x/sys`, `segmentio/encoding`

## Validation
- `go build ./...` ✅
- `script/test` ✅ (all tests pass)
- `script/lint` ✅ (0 issues)


lindazqli:
Hi @SamMorrowDrums when do we expect this to roll out? We have been getting customer tickets on 415 error. Thanks

SamMorrowDrums:
Should be shipping today!

Moarcoo:
I am still experiencing this at this moment on https://api.githubcopilot.com/mcp. Should I wait a bit longer?

weltonrodrigo:
Just to help those eager to restart interacting with the github mcp, i'm adding this reproducer.

Run as:

```bash
export GITHUB_TOKEN=$(gh auth token)
uv run --with httpx python mcp_reproducer.py
```

```python
import httpx
import sys
import os

def reproduce():
    # To run this, you need a valid GitHub Installation Token or PAT
    token = os.environ.get("GITHUB_TOKEN")
    if not token:
        print("Error: Please set GITHUB_TOKEN environment variable.")
        sys.exit(1)

    url = "https://api.githubcopilot.com/mcp"
    
    # This payload follows the MCP JSON-RPC 2.0 spec
    payload = {
        "jsonrpc": "2.0",
        "method": "tools/list",
        "id": 1
    }

    print(f"Target URL: {url}")
    print("-" * 50)

    # 1. SUCCESS CASE: Exact header
    print("Test 1: Content-Type: application/json (Expected: 200 OK)")
    headers_ok = {
        "Authorization": f"Bearer {token}",
        "Content-Type": "application/json",
        "Accept": "application/json, text/event-stream"
    }
    try:
        r1 = httpx.post(url, headers=headers_ok, json=payload)
        print(f"Result: {r1.status_code} {r1.reason_phrase}")
        if r1.status_code != 200:
            print(f"Body: {r1.text}")
    except Exception as e:
        print(f"Error: {e}")

    print("-" * 50)

    # 2. FAILURE CASE: With charset parameter
    print("Test 2: Content-Type: application/json; charset=utf-8 (Expected: Failure if bug persists)")
    headers_fail = {
        "Authorization": f"Bearer {token}",
        "Content-Type": "application/json; charset=utf-8",
        "Accept": "application/json, text/event-stream"
    }
    try:
        r2 = httpx.post(url, headers=headers_fail, json=payload)
        print(f"Result: {r2.status_code} {r2.reason_phrase}")
        if r2.status_code == 415:
            print("Reproduced: Server strictly requires 'application/json' and rejects parameters.")
            print(f"Server Message: {r2.text}")
        elif r2.status_code == 200:
            print("Fixed: Server now accepts the charset parameter.")
    except Exception as e:
        print(f"Error: {e}")

if __name__ == "__main__":
    reproduce()
```

SamMorrowDrums:
Thank you for confirming, I will check  and we can add back a middleware for this if we need to, that's unfortunate.

SamMorrowDrums:
Reopened https://github.com/github/github-mcp-server/issues/2333

PR #2365, opened 2026-04-22
fix(tools): clarify search_code path: vs filename: (#2343)

### LINKED ISSUE #2343: `search_code` tool: `path:` qualifier silently returns 0 results when used as a filename filter — no error or warning shown
### Description

The `search_code` MCP tool passes search queries to `GET /search/code` (GitHub Legacy REST API). The `path:` qualifier in this API filters by **directory path**, not by filename. However, many users and AI agents naturally use `path:` to target a specific file (e.g., `path:WaitUtils.cs`).

When `path:` is used this way, the API returns 0 results **silently** — no error, no warning, no indication that the qualifier was used incorrectly. This causes AI agents to conclude the file or content does not exist, when it actually does.

The correct qualifier for filename filtering is `filename:`, but nothing in the tool's behavior or output signals this distinction when `path:` fails.

### Steps to Reproduce

1. Configure the GitHub MCP server and use the `search_code` tool
2. Search for a term known to exist in a specific file, using `path:` as a filename filter:
   ```
   query: "WaitForElement path:WaitUtils.cs repo:PremierInc/pqa-browser-base-framework"
   ```
3. Observe: **0 results returned** — no error, no warning
4. Now repeat the same search using `filename:` instead:
   ```
   query: "WaitForElement filename:WaitUtils.cs repo:PremierInc/pqa-browser-base-framework"
   ```
5. Observe: **1 result returned** — correct file found

**Empirically confirmed April 17, 2026 against `PremierInc/pqa-browser-base-framework`.**

### Expected Behavior

One of:
- (a) When `path:` is used with a value that looks like a filename (contains a `.` extension), the tool returns a warning or note explaining that `path:` filters directories and `filename:` should be used for files, **OR**
- (b) The tool documentation/description clearly calls out this distinction, **OR**
- (c) The MCP server internally rewrites `path:Foo.cs` to `filename:Foo.cs` when a file extension is detected

### Actual Behavior

`path:WaitUtils.cs` returns 0 results with no indication of why. The user/agent has no way to know the qualifier is the problem without knowing the underlying Legacy REST API's distinction between `path:` and `filename:`.

### Additional Context

- Underlying API behavior is documented at: https://docs.github.com/en/rest/search/search#search-code
- `path:` = matches directory structure (e.g., `path:Browser/Elements` finds files inside that folder)
- `filename:` = matches the filename itself (e.g., `filename:WaitUtils.cs`)
- The GitHub MCP server source chain: [`search.go`](https://github.com/github/github-mcp-server/blob/main/pkg/github/search.go) calls `client.Search.Code()` → [`go-github/search.go`](https://github.com/google/go-github/blob/master/github/search.go) → `GET /search/code`

### Affected version
Environment:
- **VS Code:** (check `Help > About` for exact version)
- **GitHub Copilot Chat extension:** `0.44.1`
- **GitHub MCP server:** Hosted endpoint (`https://api.githubcopilot.com/mcp`) — powered by [`github/github-mcp-server`](https://github.com/github/github-mcp-server); server version is managed by GitHub

beejak:
## Summary
- Extend the \search_code\ query schema description so \path:\ (directory prefix) vs \ilename:\ (file name) is explicit.
- When GitHub returns **zero** code results and the query uses \path:\ with a **file-like** token (contains \.\, no path separators), append a second **text** content block pointing agents at \ilename:\.

## Context
Addresses the confusion described in #2343 (silent empty results when \path:\ is used like a filename filter).

## Test plan
- [x] \go test ./pkg/github -run 'Test_SearchCode|TestPathQualifier' -count=1\
- [x] \UPDATE_TOOLSNAPS=true go test ./pkg/github -run '^Test_SearchCode$' -count=1\ (schema snap updated)

Fixes #2343

### DIFF
--- a/pkg/github/__toolsnaps__/search_code.snap
+++ b/pkg/github/__toolsnaps__/search_code.snap
@@ -26,7 +26,7 @@
         "type": "number"
       },
       "query": {
-        "description": "Search query using GitHub's powerful code search syntax. Examples: 'content:Skill language:Java org:github', 'NOT is:archived language:Python OR language:go', 'repo:github/github-mcp-server'. Supports exact matching, language filters, path filters, and more.",
+        "description": "Search query using GitHub's powerful code search syntax. Examples: 'content:Skill language:Java org:github', 'NOT is:archived language:Python OR language:go', 'repo:github/github-mcp-server'. Supports exact matching, language filters, path filters, and more. Important: GitHub's path: qualifier matches directory path prefixes, not file names. To target a specific file (e.g. WaitUtils.cs), use filename:WaitUtils.cs instead of path:WaitUtils.cs. Using path: with a file-like token often returns zero results with no API error.",
         "type": "string"
       },
       "sort": {
--- a/pkg/github/search.go
+++ b/pkg/github/search.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"strings"
 
 	ghErrors "github.com/github/github-mcp-server/pkg/errors"
 	"github.com/github/github-mcp-server/pkg/inventory"
@@ -166,6 +167,25 @@ func SearchRepositories(t translations.TranslationHelperFunc) inventory.ServerTo
 	)
 }
 
+// pathQualifierLooksLikeFilename reports whether the query uses a path: token whose
+// value looks like a file name (contains a dot, no path separators). GitHub's code
+// search treats path: as a directory prefix; filtering by file name requires filename:.
+func pathQualifierLooksLikeFilename(q string) bool {
+	for _, part := range strings.Fields(q) {
+		after, ok := strings.CutPrefix(part, "path:")
+		if !ok || after == "" {
+			continue
+		}
+		if strings.ContainsAny(after, "/\\") {
+			continue
+		}
+		if strings.Contains(after, ".") {
+			return true
+		}
+	}
+	return false
+}
+
 // SearchCode creates a tool to search for code across GitHub repositories.
 func SearchCode(t translations.TranslationHelperFunc) inventory.ServerTool {
 	schema := &jsonschema.Schema{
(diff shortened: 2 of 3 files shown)


jluocsa (commented):
Thanks for tackling #2343 — the description tweak plus a runtime nudge is a nice belt-and-braces approach.

A few observations on `pathQualifierLooksLikeFilename` that may be worth addressing before merge:

**1. False positives on legitimate top-level paths.** The heuristic is "no separator + contains a dot", which fires for real paths that happen to be dotted top-level files. For example:

- `path:setup.py`
- `path:Makefile.am`
- `path:.gitignore`
- `path:requirements.txt`

These are all valid `path:` qualifier values that match top-level files in many repos. When they legitimately return zero results (e.g., the file truly doesn't exist in any matched repo), the user gets a misleading hint telling them to switch to `filename:`. The hint message is only opt-in on zero-result queries, so the user-visible impact is bounded, but it's worth deciding whether the heuristic should also require something file-extension-shaped (e.g. an extension from a small allowlist, or "no leading dot").

**2. Quoted multi-word path tokens get split.** `strings.Fields` splits on whitespace, so a query like `path:"some name.cs"` becomes two tokens: `path:"some` and `name.cs"`. The first happens to have no separator but no dot either, so it doesn't fire — but `path:"foo.bar baz"` does: the first token `path:"foo.bar` has a dot and no separator, triggering the hint on what was actually a valid quoted path. GitHub does support quoted qualifier values; consider stripping surrounding quotes or skipping tokens that begin a quoted span.

**3. Polish — include the offending token in the hint.** The hint currently always shows the generic `WaitUtils.cs` example. If you echo the user's actual token (e.g. `did you mean filename:%s?`), the message becomes directly actionable for agents that route on substrings.

Toolsnap update and unit test coverage both look good. The runtime-only message (i.e. not changing the schema for behavior) keeps the contract clean.

SamMorrowDrums:
Thank you for this, I actually did something similar in a PR I made and merged today. Let me know if you have any problems after the next release!

PR #2408, opened 2026-04-30
fix(sanitize): preserve angle brackets inside code blocks and inline code

### LINKED ISSUE #2202: GitHub MCP issue read appears to drop code block text in angle brackets

blackwell-systems:
## Description

`bluemonday.StrictPolicy()` treats angle brackets inside markdown code blocks and inline code spans as HTML tags and strips them. This causes content like `mut_raw_ptr<int>` in issue/PR bodies to become `mut_raw_ptr` when read through MCP endpoints.

### Before

```
let ptr: mut_raw_ptr<int> = raw_new int;
```
Agent sees: `let ptr: mut_raw_ptr = raw_new int;`

### After

Agent sees: `let ptr: mut_raw_ptr<int> = raw_new int;`

## Root cause

`FilterHTMLTags` calls `bluemonday.Sanitize()` on the entire markdown body without distinguishing code from prose. Bluemonday treats `<int>`, `<T>`, `<String>`, etc. as unrecognized HTML tags and removes them.

## Fix

Before HTML sanitization, replace `<` and `>` inside fenced code blocks (` ``` `) and inline code spans (`` ` ``) with null-byte sentinels that bluemonday will not interpret as HTML. After sanitization, restore the sentinels to angle brackets.

This preserves XSS protection for angle brackets in prose (e.g. `<script>` is still stripped) while keeping angle brackets inside code intact.

## Testing

Added 6 test cases covering:
- Fenced code blocks with angle brackets (the reported bug)
- Inline code with generic types (`Vec<String>`)
- Angle brackets outside code still sanitized (XSS protection preserved)
- Multiple inline code spans
- Fenced code with language info string

Fixes #2202

### DIFF
--- a/pkg/sanitize/sanitize.go
+++ b/pkg/sanitize/sanitize.go
@@ -12,7 +12,14 @@ var policy *bluemonday.Policy
 var policyOnce sync.Once
 
 func Sanitize(input string) string {
-	return FilterHTMLTags(FilterCodeFenceMetadata(FilterInvisibleCharacters(input)))
+	cleaned := FilterCodeFenceMetadata(FilterInvisibleCharacters(input))
+	// Protect angle brackets inside code blocks and inline code spans
+	// from being stripped by the HTML sanitizer. bluemonday treats <int>,
+	// <T>, etc. as unknown HTML tags and removes them.
+	// See https://github.com/github/github-mcp-server/issues/2202
+	protected := protectCodeAngleBrackets(cleaned)
+	sanitized := FilterHTMLTags(protected)
+	return restoreCodeAngleBrackets(sanitized)
 }
 
 // FilterInvisibleCharacters removes invisible or control characters that should not appear
@@ -143,6 +150,157 @@ func isSafeCodeFenceToken(token string) bool {
 		return false
 	}
 	return true
+}
+
+// Sentinels used to protect angle brackets inside code from HTML sanitization.
+// These are chosen to be unlikely to appear in real content.
+const (
+	ltSentinel = "\x00LT\x00"
+	gtSentinel = "\x00GT\x00"
+)
+
+// protectCodeAngleBrackets replaces < and > inside fenced code blocks and
+// inline code spans with sentinels so bluemonday does not strip them as HTML.
+func protectCodeAngleBrackets(input string) string {
+	var b strings.Builder
+	b.Grow(len(input))
+
+	runes := []rune(input)
+	i := 0
+	n := len(runes)
+
+	for i < n {
+		// Fenced code block: ``` ... ```
+		if i+2 < n && runes[i] == '`' && runes[i+1] == '`' && runes[i+2] == '`' {
+			// Find the fence length
+			fenceStart := i
+			fenceLen := 0
+			for i < n && runes[i] == '`' {
+				fenceLen++
+				i++
+			}
+			// Write opening fence + rest of line (info string)
+			for range fenceLen {
+				b.WriteRune('`')
+			}
+			for i < n && runes[i] != '\n' {
+				b.WriteRune(runes[i])
+				i++
+			}
+			if i < n {
+				b.WriteRune(runes[i]) // newline
+				i++
+			}
+			// Inside fence: protect angle brackets until closing fence
+			for i < n {
+				// Check for closing fence
+				if runes[i] == '`' {
+					closeLen := 0
+					j := i
+					for j < n && runes[j] == '`' {
+						closeLen++
+						j++
+					}
+					if closeLen >= fenceLen {
+						for range closeLen {
+							b.WriteRune('`')
+						}
+						i = j
+						break
+					}
+				}
+				switch runes[i] {
+				case '<':
+					b.WriteString(ltSentinel)
+				case '>':
+					b.WriteString(gtSentinel)
+				default:
+					b.WriteRune(runes[i])
+				}
+				i++
+			}
+			_ = fenceStart
+			continue
+		}
+
+		// Inline code: `...`
+		if runes[i] == '`' {
+			// Count opening backticks
+			openLen := 0
+			j := i
+			for j < n && runes[j] == '`' {
+				openLen++
+				j++
+			}
+			// Don't treat ``` as inline code (handled above for fenced blocks)
+			if openLen >= 3 {
+				for range openLen {
+					b.WriteRune('`')
+				}
+				i = j
+				continue
+			}
+			// Find matching closing backticks
+			closeStart := -1
+			for k := j; k <= n-openLen; k++ {
+				match := true
+				for m := range openLen {
+					if runes[k+m] != '`' {
+						match = false
+						break
+					}
+				}
+				if match {
+					// Verify it's exactly openLen backticks (not more)
+					if k+openLen < n && runes[k+openLen] == '`' {
+						continue
+					}
+					closeStart = k
+					break
+				}
+			}
+			if closeStart == -1 {
+				// No closing backticks found; treat as literal
+				for range openLen {
+					b.WriteRune('`')
+				}
+				i = j
+				continue
+			}
+			// Write opening backticks
+			for range openLen {
+				b.WriteRune('`')
+			}
+			// Protect content
+			for i = j; i < closeStart; i++ {
+				switch runes[i] {
+				case '<':
+					b.WriteString(ltSentinel)
+				case '>':
+					b.WriteString(gtSentinel)
+				default:
+					b.WriteRune(runes[i])
+				}
+			}
+			// Write closing backticks
+			for range openLen {
+				b.WriteRune('`')
+			}
+			i = closeStart + openLen
+			continue
+		}
+
+		b.WriteRune(runes[i])
+		i++
+	}
+
+	return b.String()
+}
+
+// restoreCodeAngleBrackets converts sentinels back to angle brackets.
+func restoreCodeAngleBrackets(input string) string {
+	s := strings.ReplaceAll(input, ltSentinel, "<")
+	return strings.ReplaceAll(s, gtSentinel, ">")
 }
 
 func getPolicy() *bluemonday.Policy {
--- a/pkg/sanitize/sanitize.go
+++ b/pkg/sanitize/sanitize.go
@@ -1,16 +1,38 @@
+// Package sanitize provides functions to sanitize untrusted user content from
+// GitHub (issue titles, PR bodies, search results, etc.) before display.
+//
+// Threat model:
+// - Input source: untrusted GitHub content that may contain malicious HTML/JS
+// - Goal: strip HTML tags from prose without corrupting code samples
+// - Defense: bluemonday HTML sanitizer + angle bracket protection in code blocks
+//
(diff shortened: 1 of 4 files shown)


jluocsa (commented):
Thanks for picking up #2202 — generic-type angle brackets disappearing from code blocks is a long-standing irritation. The CommonMark-aware fence/inline detection is much better than a regex-based approximation.

Before merge, I think there's a security-sensitive issue worth addressing:

**Sentinel collision can be used to bypass the HTML sanitizer.**

The current sentinels are:

```go
const (
    ltSentinel = "\x00LT\x00"
    gtSentinel = "\x00GT\x00"
)
```

`Sanitize` runs in this order:

```go
cleaned   := FilterCodeFenceMetadata(FilterInvisibleCharacters(input))
protected := protectCodeAngleBrackets(cleaned)
sanitized := FilterHTMLTags(protected)
return restoreCodeAngleBrackets(sanitized)
```

`FilterInvisibleCharacters` strips a specific list of Unicode codepoints (see `shouldRemoveRune` in `pkg/sanitize/sanitize.go` — zero-width spaces, BiDi controls, tag characters, etc.) but **does not strip `\x00`** (NUL is not in the list).

That means an attacker who controls input — e.g. an issue title, PR body, search-result snippet, etc. — can submit:

```
\x00LT\x00script\x00GT\x00alert(1)\x00LT\x00/script\x00GT\x00
```

…outside any code block. The flow is:

1. `FilterInvisibleCharacters` → passes through (NUL not stripped)
2. `FilterCodeFenceMetadata` → passes through (no fence)
3. `protectCodeAngleBrackets` → passes through (no backticks, the sentinel literal is treated as ordinary text)
4. `FilterHTMLTags` (bluemonday) → passes through (`\x00LT\x00script...` is not HTML)
5. `restoreCodeAngleBrackets` → **substitutes back into `<script>alert(1)</script>`**

The output now contains raw HTML/JS that should have been stripped, defeating the purpose of `FilterHTMLTags`.

A few mitigation options, in increasing order of robustness:

- **Add `\x00` to `shouldRemoveRune`.** Cheapest fix; consistent with the existing "strip control / invisible chars" policy. Adding it before `protectCodeAngleBrackets` guarantees no pre-existing sentinels.
- **Pre-scan for the literal sentinel and refuse / escape it.** A bit more code, no change to filter semantics.
- **Use a per-call randomized sentinel** (generate two crypto-random byte sequences once per `Sanitize` invocation). Defeats collision by construction; only downside is a small allocation per call.

Option 1 seems most consistent with what `FilterInvisibleCharacters` is already doing, and the smallest possible change. A unit test mirroring `TestSanitizePreservesAngleBracketsInCodeBlocks` with input containing the literal sentinel would lock the behavior in.

**Minor: closing-fence detection.** `protectCodeAngleBrackets` treats any run of `>= fenceLen` backticks as a closing fence (anywhere on the line). CommonMark requires the closing fence to be on its own line and not have an info string. The current implementation is more permissive, which in practice means a stray ```` ``` ```` inside what was meant to be the fenced content will end protection early. This is a soft-fail (some `<` get stripped that shouldn't be) rather than a security issue, so probably acceptable, but worth a comment near the loop documenting the deviation.

Tests are nicely focused, the inline-vs-fenced split is handled, and `[T comparable]`-style generics outside code (where they really might be HTML-ish) are still cleaned. Once the sentinel exposure is closed, this is a great fix.

pachecocordovamoiseseduardo-byte (approved):

pkg/sanitize/sanitize.go
```diff
@@ -12,7 +12,14 @@ var policy *bluemonday.Policy
...
 func Sanitize(input string) string {
-	return FilterHTMLTags(FilterCodeFenceMetadata(FilterInvisibleCharacters(input)))
+	cleaned := FilterCodeFenceMetadata(FilterInvisibleCharacters(input))
+	// Protect angle brackets inside code blocks and inline code spans
+	// from being stripped by the HTML sanitizer. bluemonday treats <int>,
+	// <T>, etc. as unknown HTML tags and removes them.
+	// See https://github.com/github/github-mcp-server/issues/2202
+	protected := protectCodeAngleBrackets(cleaned)
+	sanitized := FilterHTMLTags(protected)
+	return restoreCodeAngleBrackets(sanitized)
```
SamMorrowDrums, line 44: Optional: pull out the ordering as a named invariant so future readers don't accidentally reorder these calls. The security argument relies on `FilterInvisibleCharacters` running first (so NUL is stripped before sentinels are introduced) and on `restoreCodeAngleBrackets` running last (after `FilterHTMLTags`).

```suggestion
func Sanitize(input string) string {
	// Ordering is security-load-bearing:
	//   1. FilterInvisibleCharacters strips NUL bytes BEFORE sentinels are
	//      introduced, preventing an attacker from forging \x00LT\x00 in input.
	//   2. protectCodeAngleBrackets replaces < and > inside code with sentinels
	//      so bluemonday does not strip them as HTML.
	//   3. FilterHTMLTags (bluemonday) sanitizes prose; sentinels pass through.
	//   4. restoreCodeAngleBrackets converts sentinels back to < and >.
	// See https://github.com/github/github-mcp-server/issues/2202
	cleaned := FilterCodeFenceMetadata(FilterInvisibleCharacters(input))
	protected := protectCodeAngleBrackets(cleaned)
	sanitized := FilterHTMLTags(protected)
	return restoreCodeAngleBrackets(sanitized)
}
```

pkg/sanitize/sanitize.go
```diff
@@ -145,6 +152,162 @@ func isSafeCodeFenceToken(token string) bool {
 	return true
 }
 
+// Sentinels used to protect angle brackets inside code from HTML sanitization.
+// These are chosen to be unlikely to appear in real content.
+const (
+	ltSentinel = "\x00LT\x00"
+	gtSentinel = "\x00GT\x00"
+)
```
SamMorrowDrums, line 185: Worth noting in the doc comment that the NUL-stripping in `FilterInvisibleCharacters` is what makes these sentinels collision-free — it's a non-obvious dependency.

```suggestion
// Sentinels used to protect angle brackets inside code from HTML sanitization.
// They embed NUL bytes, which are stripped from all input by
// FilterInvisibleCharacters before this function runs — that strip is what
// makes the sentinels collision-free against attacker-controlled input.
const (
	ltSentinel = "\x00LT\x00"
	gtSentinel = "\x00GT\x00"
)
```

SamMorrowDrums (commented):
Thanks for this @blackwell-systems — the approach is sound and the security trade-off is defensible. bluemonday continues to handle prose, and the sentinel design has the right ordering: `FilterInvisibleCharacters` strips NUL bytes before `protectCodeAngleBrackets` runs, so an attacker can't forge sentinels in the input.

Cc'ing @joannakl as a fellow maintainer — would appreciate a second pair of eyes since this touches the sanitizer.

I've left two inline suggestions to harden the contract (ordering invariant + sentinel doc). One bigger consideration:

**Have you considered a CommonMark parser?** `github.com/yuin/goldmark` is already a transitive dep in `go.sum`. I prototyped a drop-in replacement for `protectCodeAngleBrackets` against this branch:

```go
func protectCodeAngleBrackets(input string) string {
    src := []byte(input)
    doc := goldmark.DefaultParser().Parse(text.NewReader(src))

    type span struct{ start, stop int }
    var spans []span
    _ = ast.Walk(doc, func(n ast.Node, entering bool) (ast.WalkStatus, error) {
        if !entering {
            return ast.WalkContinue, nil
        }
        switch n.Kind() {
        case ast.KindFencedCodeBlock, ast.KindCodeBlock:
            lines := n.Lines()
            for i := 0; i < lines.Len(); i++ {
                seg := lines.At(i)
                spans = append(spans, span{seg.Start, seg.Stop})
            }
        case ast.KindCodeSpan:
            for c := n.FirstChild(); c != nil; c = c.NextSibling() {
                if t, ok := c.(*ast.Text); ok {
                    seg := t.Segment
                    spans = append(spans, span{seg.Start, seg.Stop})
                }
            }
        }
        return ast.WalkContinue, nil
    })

    if len(spans) == 0 {
        return input
    }
    var out bytes.Buffer
    out.Grow(len(src))
    cursor := 0
    for _, s := range spans {
        out.Write(src[cursor:s.start])
        for _, b := range src[s.start:s.stop] {
            switch b {
            case '<':
                out.WriteString(ltSentinel)
            case '>':
                out.WriteString(gtSentinel)
            default:
                out.WriteByte(b)
            }
        }
        cursor = s.stop
    }
    out.Write(src[cursor:])
    return out.String()
}
```

About 40 lines instead of 150, and it picks up **indented (4-space) code blocks** for free, which the current scanner misses:

```
Text

    let x: Vec<u8> = vec![];
```

I ran it against the same edge cases (4-backtick fences, nested backticks, unclosed fences, sentinel-injection probes, `<script>` inside fences, etc.) and behavior matches or improves on the hand-rolled version. NUL-strip ordering is preserved, so the sentinel-collision defense is unchanged.

Would also let us drop ~150 lines and rely on a CommonMark-compliant parser for future edge cases (autolinks adjacent to fences, etc.).

Happy with this either way — if you'd rather land the current approach and follow up, that's totally fine. Up to you and @joannakl. 🙏

A non-blocking nit: the package doc would be a good place to write down the threat model (input source: untrusted GitHub content; goal: strip HTML from prose without corrupting code) so the next reader has the context up front.

blackwell-systems:
@SamMorrowDrums Thanks for the goldmark suggestion and prototype! You're right that it's cleaner and more maintainable.

I've refactored to use goldmark's CommonMark parser. Changes:

1. **Replaced hand-rolled scanner** with goldmark AST walker (~150 lines → ~40 lines)
2. **Added indented code block coverage** (4-space blocks now protected)
3. **Added package documentation** with threat model
4. **Documented ordering invariant** (FilterInvisibleCharacters before protectCodeAngleBrackets)
5. **Updated sentinel documentation** explaining collision prevention

All existing tests pass plus a new test for indented code blocks. Ran `script/lint` locally (0 issues) and `go test -race ./...`.

Happy to address the inline suggestions if there's anything else needed. Thanks again for the detailed review.

PR #2432, opened 2026-05-06
Add ifc label for get_me tool

JoannaaKL:
## Summary

Adds minimal IFC security label annotations to `get_me` tool output behind `--insiders` flag.

## What changed

- Replaced `pkg/ifc/` lattice implementation with a lightweight `SecurityLabel` struct using `Integrity` and `Confidentiality` enums — annotation only, no lattice algebra. The enforcement engine lives in a separate service.
- `get_me` sets `result.Meta["ifc"]` with `{"integrity":"trusted","confidentiality":["public"]}` when insiders mode is enabled.
- Updated `script/get-me` to reliably capture and display the label.

```
{
   "_meta": {
    "ifc": {
      "integrity": "trusted",
      "confidentiality": [
        "public"
      ]
    }
  },
  "content": {
    "login": "JoannaaKL",
    "id": 67866556,
    "profile_url": "https://github.com/JoannaaKL",
    "avatar_url": "https://avatars.githubusercontent.com/u/67866556?v=4",
    "details": {
      "name": "JoannaaKL",
      "company": "GitHub",
    }
  }
}
```

## MCP impact

  - `get_me` result includes `_meta.ifc` in insiders mode only. No change to stable behavior.

## Security / limits

## Tool renaming

## Lint & tests

## Docs

- [x] Not needed


pkg/github/context_tools.go
```diff
@@ -103,7 +104,13 @@ func GetMe(t translations.TranslationHelperFunc) inventory.ServerTool {
 				},
 			}
 
-			return MarshalledTextResult(minimalUser), nil, nil
+			result := MarshalledTextResult(minimalUser)
+			if deps.GetFlags(ctx).InsidersMode {
+				result.Meta = mcp.Meta{
+					"ifc": ifc.LabelGetMe(),
```
gokhanarkan: Heads-up: the wire format here is `_meta.ifc` with {integrity: "high"|"low", confidentiality: ["public"|<user>...]}, but original spec `_meta.ifc_label` with {integrity: "trusted"|"untrusted", confidentiality: {readers: [...]}}. I'm fine matching your shape on the cli side), but should we update the issues to lock in ifc + high/low + flat list as the contract before we annotate more tools?

gokhanarkan: Also this assigns result.Meta outright. If `MarshalledTextResult` ever starts populating Meta (or another decorator runs before this), we'd silently clobber it. cheaper to be defensive now and merge into an existing map than to debug later, something like if result.Meta == nil { result.Meta = mcp.Meta{} }; result.Meta["ifc"] = ifc.LabelGetMe(). wdyt?

pkg/ifc/labelling_engine_readers.go
```diff
@@ -0,0 +1,6 @@
+package ifc
+
+// LabelGetMe returns a label for get_me: trusted, universal readers.
+func LabelGetMe() ReadersSecurityLabel {
```
gokhanarkan: handler-level test in `context_tools_test.go` covers the integration nicely. Worth a one-liner unit test in `pkg/ifc` too. assert.True(t, LabelGetMe().IsHighIntegrity() && LabelGetMe().IsPublicConfidentiality())

I think it is very cheap to add now, and as l`abelling_engine_readers.go` grows (LabelListIssues, LabelGetFileContents, …) we'll have a single place tracking per-tool label correctness without spinning up the full handler.

gokhanarkan (commented):
Some comments to make sure we align on the spec!

omgitsads (approved):

PR #2447, opened 2026-05-09
chore: migrate to mcpb-pack@v3 (drop server.json)

mgoldsborough:
## Summary

- Delete `server.json` from this bundle. mpak now composes the MCP registry's [`ServerDetail`](https://github.com/modelcontextprotocol/registry/blob/main/docs/reference/server-json/draft/server.schema.json) discovery shape from `manifest.json` server-side, so the per-bundle file is unused.
- Drop the workflow step that bumped `server.json::version` from `manifest.json`.
- Bump `NimbleBrainInc/mcpb-pack@v2` → `@v3` in the release workflow.

## Optional follow-up

Add a branded reverse-DNS name override in `manifest.json` if you want a non-default registry name:

```json
{
  "_meta": {
    "dev.mpak/registry": {
      "name": "ai.nimblebrain/<bundle-name>"
    }
  }
}
```

Without an override the registry uses the mechanical default `dev.mpak.<scope>/<name>`.

## Test plan

- [ ] Cut a release and confirm the build workflow succeeds against `mcpb-pack@v3`
- [ ] Verify the bundle appears at `https://registry.mpak.dev/v1/servers/<name>` after announce


SamMorrowDrums:
Hey 👋, I think if we're going to add MCPB we would use the binary build artefacts generated in the release job, as part of same job.

There also appear to be some unexpected changes in the diff.

PR #2450, opened 2026-05-11
feat: Add CSV output format for default list tools under insiders mode

RossTarrant:
## Summary

Adds feature-gated CSV output for default `list_*` tools and refactors Insiders mode into a feature flag bundle.

## Why

CSV provides a flatter, more compact representation for list-style responses, helping evaluate token reduction from output format changes.

Refs github/copilot-mcp-core#1323

## What changed

 - Added **csv_output** as an allowed Insiders feature flag.
 - Added feature-gated JSON/CSV variants for list_* tools.
 - Added central JSON text response to CSV conversion with dot-notation flattening for nested objects.
 - Normalised whitespace in body fields for CSV output.
 - Refactored Insiders mode so it expands to curated feature flags instead of being checked directly by tools.
 - Added `ifc_labels` as an Insiders-enabled feature flag.
 - Updated MCP Apps and IFC gates to check concrete feature flags.
 - Added unit coverage for CSV conversion, feature-gated variants, and Insiders feature expansion.

## MCP impact

 - [ ]  No tool or API changes
 - [X]  Tool schema or behavior changed
 CSV mode changes list_* tool (within default toolset) response behavior under the csv_output Insiders feature flag. Tool schemas are unchanged; CSV output is server-controlled by feature flag gating.
 - [ ]  New tool added

 - "List discussions in RossTarrant/octopus-mcp"
 - "List issues in github/github-mcp-server"
 - "List pull requests in github/github-mcp-server"

## Security / limits

 - [ ]  No security or limits impact
 - [ ]  Auth / permissions considered
 - [X]  Data exposure, filtering, or token/size limits considered
 CSV mode is intended to reduce response size for list tools. Body fields are whitespace-normalized in CSV output; full formatted content remains available through corresponding get/read tools.

## Tool renaming

 - [ ]  I am renaming tools as part of this PR (e.g. a part of a consolidation effort)
  - [ ]  I have added the new tool aliases in deprecated_tool_aliases.go
 - [X]  I am not renaming tools as part of this PR

## Lint & tests

 - [X]  Linted locally with ./script/lint
 - [X]  Tested locally with ./script/test

## Docs

 - [ ]  Not needed
- [X] Updated (README / docs / examples) - Updated `docs/insiders-features.md` detailing the new experimental feature

### DIFF
--- a/internal/ghmcp/server.go
+++ b/internal/ghmcp/server.go
@@ -143,7 +143,6 @@ func NewStdioMCPServer(ctx context.Context, cfg github.MCPServerConfig) (*mcp.Se
 		cfg.Translator,
 		github.FeatureFlags{
 			LockdownMode: cfg.LockdownMode,
-			InsidersMode: cfg.InsidersMode,
 		},
 		cfg.ContentWindowSize,
 		featureChecker,
@@ -229,7 +228,7 @@ type StdioServerConfig struct {
 	// LockdownMode indicates if we should enable lockdown mode
 	LockdownMode bool
 
-	// InsidersMode indicates if we should enable experimental features
+	// InsidersMode expands to the curated set of feature flags enabled for insiders.
 	InsidersMode bool
 
 	// ExcludeTools is a list of tool names to disable regardless of other settings.
@@ -345,7 +344,7 @@ func RunStdioServer(cfg StdioServerConfig) error {
 
 // createFeatureChecker returns a FeatureFlagChecker that resolves features
 // using the centralized ResolveFeatureFlags function. For the local server,
-// features are resolved once at startup from --features CLI flag + insiders mode.
+// features are resolved once at startup from --features CLI flag and insiders mode.
 func createFeatureChecker(enabledFeatures []string, insidersMode bool) inventory.FeatureFlagChecker {
 	featureSet := github.ResolveFeatureFlags(enabledFeatures, insidersMode)
 	return func(_ context.Context, flagName string) (bool, error) {
--- a/pkg/github/csv_output.go
+++ b/pkg/github/csv_output.go
@@ -1,0 +1,412 @@
+package github
+
+import (
+	"bytes"
+	"context"
+	"encoding/csv"
+	"encoding/json"
+	"fmt"
+	"sort"
+	"strings"
+
+	"github.com/github/github-mcp-server/pkg/inventory"
+	"github.com/github/github-mcp-server/pkg/utils"
+	"github.com/modelcontextprotocol/go-sdk/mcp"
+)
+
+// Ordered by preference when a response wrapper contains multiple arrays.
+var primaryCSVRowKeys = []string{
+	"items",
+	"issues",
+	"discussions",
+	"categories",
+	"labels",
+	"alerts",
+	"advisories",
+	"notifications",
+	"gists",
+	"repositories",
+	"commits",
+	"branches",
+	"tags",
+	"releases",
+	"users",
+	"teams",
+	"members",
+	"projects",
+	"nodes",
+}
+
+type csvOutputDocument struct {
+	metadata map[string]string
+	rows     []map[string]string
+}
+
+func withCSVOutputVariants(tools []inventory.ServerTool) []inventory.ServerTool {
+	result := make([]inventory.ServerTool, 0, len(tools))
+	for _, tool := range tools {
+		if !isCSVOutputTool(tool) {
+			result = append(result, tool)
+			continue
+		}
+
+		jsonOnly := tool
+		jsonOnly.FeatureFlagDisable = FeatureFlagCSVOutput
+		result = append(result, jsonOnly)
+
+		csvCapable := tool
+		csvCapable.FeatureFlagEnable = FeatureFlagCSVOutput
+		csvCapable.HandlerFunc = wrapHandlerWithCSVOutput(tool.HandlerFunc)
+		result = append(result, csvCapable)
+	}
+	return result
+}
+
+func isCSVOutputTool(tool inventory.ServerTool) bool {
+	if !tool.Toolset.Default {
+		return false
+	}
+	if !strings.HasPrefix(tool.Tool.Name, "list_") {
+		return false
+	}
+	return tool.FeatureFlagEnable == "" && tool.FeatureFlagDisable == ""
+}
+
+func wrapHandlerWithCSVOutput(next inventory.HandlerFunc) inventory.HandlerFunc {
+	return func(deps any) mcp.ToolHandler {
+		handler := next(deps)
+		return func(ctx context.Context, req *mcp.CallToolRequest) (*mcp.CallToolResult, error) {
+			result, err := handler(ctx, req)
+			if err != nil || result == nil || result.IsError {
+				return result, err
+			}
+
+			return convertJSONTextResultToCSV(result), nil
+		}
+	}
+}
+
+func convertJSONTextResultToCSV(result *mcp.CallToolResult) *mcp.CallToolResult {
+	if len(result.Content) != 1 {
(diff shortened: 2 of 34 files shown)


pkg/github/csv_output.go
```diff
@@ -0,0 +1,408 @@
...
+	writeCSVMetadata(&buf, doc.metadata)
+	if len(doc.rows) == 0 {
+		return buf.String(), nil
+	}
+
+	headers := csvHeaders(doc.rows)
+	writer := csv.NewWriter(&buf)
+	if err := writer.Write(headers); err != nil {
+		return "", fmt.Errorf("failed to write CSV header: %w", err)
+	}
```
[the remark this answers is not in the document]
RossTarrant: Resolved, by adding a length check as suggested

pachecocordovamoiseseduardo-byte (approved):

SamMorrowDrums (commented):
@RossTarrant I thought in a previous version similar bit of work I had asked that insiders would be expanded to the feature flags it enables, and nothing would be gated on insiders, but the feature that it is actually related to, and insiders is just a shortcut to enabled flags. Just like `all` and `default` are meta toolsets (contain more than one toolset), `insiders` should just be a meta feature flag, that enables a set of feature flags, so at the point they are checked, they should be considered to be regular feature flags.

Nothing else in the codebase should need to know what insiders is. That just leaks the concept too far. 

Can this PR resolve that while it's making changes to the feature checker?

internal/ghmcp/server.go
```diff
@@ -345,9 +344,12 @@ func RunStdioServer(cfg StdioServerConfig) error {
...
 // createFeatureChecker returns a FeatureFlagChecker that resolves features
 // using the centralized ResolveFeatureFlags function. For the local server,
-// features are resolved once at startup from --features CLI flag + insiders mode.
+// features are resolved once at startup from --features CLI flag and meta flags.
 func createFeatureChecker(enabledFeatures []string, insidersMode bool) inventory.FeatureFlagChecker {
-	featureSet := github.ResolveFeatureFlags(enabledFeatures, insidersMode)
+	featureSet := github.ResolveFeatureFlags(
+		enabledFeatures,
+		github.MetaFeatureFlagsForInsiders(insidersMode)...,
+	)
```
SamMorrowDrums: I think you need to resolved the user provided `enabled feature flags` by the allowed features: https://github.com/github/github-mcp-server/blob/e9533330b0ec2ea2f81ec490cc0c84d30538fcb9/pkg/github/feature_flags.go#L36

Because insiders mode flags list should be already in there. Also can't this directly reference the list of insiders features? https://github.com/github/github-mcp-server/blob/e9533330b0ec2ea2f81ec490cc0c84d30538fcb9/pkg/github/feature_flags.go#L20

internal/ghmcp/server.go
```diff
@@ -372,9 +371,6 @@ func addUserAgentsMiddleware(cfg github.MCPServerConfig, restUATransp *transport
 				message.Params.ClientInfo.Name,
 				message.Params.ClientInfo.Version,
 			)
-			if cfg.InsidersMode {
-				userAgent += " (insiders)"
-			}
```
SamMorrowDrums, line 377: @RossTarrant I think we do still want the UA set here. Maybe there is minor reason for some places to know insiders is on, albeit minimal. Apologies.

RossTarrant: @SamMorrowDrums I took your previous comment a bit too far! 😄 Have added it back

pkg/http/middleware/cors.go
```diff
@@ -17,6 +17,7 @@ func SetCorsHeaders(h http.Handler) http.Handler {
 		"Mcp-Session-Id",
 		"Mcp-Protocol-Version",
 		"Last-Event-ID",
+		"X-Custom-Auth-Headers",
```
SamMorrowDrums: Was this needed?

pkg/http/middleware/cors.go
SamMorrowDrums, line 27: This seems to be missing the features header actually. So end user can enable via HTTP.

You can check remote, but I believe it is `x-mcp-features` or similar.

pkg/inventory/filters.go
```diff
@@ -112,16 +125,36 @@ func (r *Inventory) AvailableTools(ctx context.Context) []ServerTool {
...
-		if result[i].Toolset.ID != result[j].Toolset.ID {
-			return result[i].Toolset.ID < result[j].Toolset.ID
+	sortTools(result)
+
+	return result
+}
+
+// AvailableToolsWithoutFeatureFiltering returns tools that pass every filter
+// except FeatureFlagEnable/FeatureFlagDisable.
+func (r *Inventory) AvailableToolsWithoutFeatureFiltering(ctx context.Context) []ServerTool {
```
SamMorrowDrums: I am curious why we need this? Is it just for testing? I would have thought it would be sufficient to just either have features on or of. The disable tools are always on if no features are present.

pkg/inventory/filters.go
```diff
@@ -112,16 +125,36 @@ func (r *Inventory) AvailableTools(ctx context.Context) []ServerTool {
...
+	for i := range r.tools {
+		tool := &r.tools[i]
+		if r.isToolEnabledWithFeatureFlags(ctx, tool, false) {
+			result = append(result, *tool)
 		}
-		return result[i].Tool.Name < result[j].Tool.Name
-	})
+	}
+
+	sortTools(result)
```
SamMorrowDrums: If we don't ultimately need this function than maybe we don't need to extract sort.

pkg/inventory/filters.go
```diff
@@ -139,16 +172,36 @@ func (r *Inventory) AvailableResourceTemplates(ctx context.Context) []ServerReso
...
+}
+
+// AvailableResourceTemplatesWithoutFeatureFiltering returns resource templates
+// that pass every filter except FeatureFlagEnable/FeatureFlagDisable.
+func (r *Inventory) AvailableResourceTemplatesWithoutFeatureFiltering(_ context.Context) []ServerResourceTemplate {
+	var result []ServerResourceTemplate
+	for i := range r.resourceTemplates {
+		res := &r.resourceTemplates[i]
+		if r.isToolsetEnabled(res.Toolset.ID) {
+			result = append(result, *res)
```
SamMorrowDrums: Same patter, but I'm not sure why we need.

pkg/inventory/filters.go
```diff
@@ -166,12 +219,23 @@ func (r *Inventory) AvailablePrompts(ctx context.Context) []ServerPrompt {
...
+	for i := range r.prompts {
+		prompt := &r.prompts[i]
+		if r.isToolsetEnabled(prompt.Toolset.ID) {
+			result = append(result, *prompt)
 		}
-		return result[i].Prompt.Name < result[j].Prompt.Name
-	})
+	}
+
+	sortPrompts(result)
```
SamMorrowDrums: Same question here. Also, you can use Go generics to not need to do a separate typed sort function per type.

SamMorrowDrums (commented):
This is getting there, my biggest concern is that the way CSV tools is being added, it doesn't look like it will correctly have the http mode enable/disable, and curious how that would working in github/github-mcp-server-remote also whether the wiring would automatically work or not.

pkg/http/middleware/cors.go
RossTarrant: I could make it a separate PR if needed, but this is required to allow X-MCP-FEATURES headers

RossTarrant: Ah, I've switched it for the correct header in https://github.com/github/github-mcp-server/pull/2450/commits/0c9b8b3bcdf4c191c0f18b368396b6b1cbe2bae7

SamMorrowDrums (approved):

PR #2453, opened 2026-05-11
Add ifc label for list_issues tool

gokhanarkan:
Emits an IFC `SecurityLabel` on the `list_issues` tool result when the `InsidersMode` flag is enabled, mirroring the pattern landed for `get_me` in #2432.

Refs github/copilot-mcp-core#1623, github/copilot-mcp-core#1389. This is one of the ingress tools listed in #1623's tool table.

## What this PR does

- Wires `_meta.ifc` onto the `list_issues` `CallToolResult` when `deps.GetFlags(ctx).InsidersMode` is true. No behaviour change when the flag is off.
- Public repos → `PublicUntrusted()` (issue contents are attacker-controllable, so integrity is `untrusted`; readers are `["public"]`).
- Private repos → `PrivateUntrusted([owner])` as a **placeholder** for the reader set. See limitation below.
- Adds an `IsPrivate` field to all four `ListIssues*Query*` GraphQL types so visibility is available without an extra REST round trip. `IssueQueryResult` gains a `GetIsPrivate() bool` method.

## New shared helpers

Adds four general label constructors in `pkg/ifc/ifc.go` to seed reuse for upcoming ingress tools (`get_file_contents`, `search_issues`, …):
- `PublicTrusted()`
- `PublicUntrusted()`
- `PrivateTrusted([]string)`
- `PrivateUntrusted([]string)`

`LabelGetMe()` is refactored to delegate to `PublicTrusted()`; output is byte-for-byte identical (covered by the existing `Test_GetMe_IFC_InsidersMode` test). Flagging this for reviewers in case duplication concerns come up — happy to slim back to a single `LabelListIssues` helper if preferred.

## Known limitation (called out for reviewers)

For private repositories, the confidentiality reader set is currently `[owner]` rather than the full collaborator list. Fetching collaborators requires a paginated REST call per `list_issues` invocation; rather than bake that into this PR, I'd like to land a shared visibility/collaborators helper in a follow-up that `get_file_contents` and `search_issues` can also share. There's a `TODO(fides)` comment at the call site.

## Tests

`Test_ListIssues_IFC_InsidersMode` mirrors `Test_GetMe_IFC_InsidersMode` with three subtests:
2. Insiders on, public repo → `integrity=untrusted`, `confidentiality=["public"]`.
3. Insiders on, private repo → `integrity=untrusted`, `confidentiality=["<owner>"]`.

Existing `Test_ListIssues` GraphQL query strings and mock responses were updated to include the new `isPrivate` field.

## Validation

### DIFF
--- a/pkg/github/issues.go
+++ b/pkg/github/issues.go
@@ -1595,13 +1595,20 @@ func ListIssues(t translations.TranslationHelperFunc) inventory.ServerTool {
 				if result.Meta == nil {
 					result.Meta = mcp.Meta{}
 				}
-				// TODO(fides): for private repos, populate readers with the
-				// repository's collaborator logins. Using [owner] as a
-				// placeholder until a shared visibility/collaborators helper
-				// lands (tracked under copilot-mcp-core#1623).
 				var readers []string
 				if isPrivate {
-					readers = []string{owner}
+					restClient, err := deps.GetClient(ctx)
+					if err == nil {
+						if collaborators, err := FetchRepoCollaborators(ctx, restClient, owner, repo); err == nil {
+							readers = collaborators
+						}
+					}
+					// Fall back to the repository owner so the reader set is
+					// never empty for a private repository even if the
+					// collaborators lookup fails.
+					if len(readers) == 0 {
+						readers = []string{owner}
+					}
 				}
 				result.Meta["ifc"] = ifc.LabelListIssues(isPrivate, readers)
 			}
--- a/pkg/github/issues.go
+++ b/pkg/github/issues.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	ghErrors "github.com/github/github-mcp-server/pkg/errors"
+	"github.com/github/github-mcp-server/pkg/ifc"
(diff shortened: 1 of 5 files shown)


pkg/github/issues.go
```diff
@@ -1568,11 +1584,28 @@ func ListIssues(t translations.TranslationHelperFunc) inventory.ServerTool {
...
+			result := MarshalledTextResult(resp)
+			if deps.GetFlags(ctx).InsidersMode {
+				if result.Meta == nil {
+					result.Meta = mcp.Meta{}
+				}
+				// TODO(fides): for private repos, populate readers with the
+				// repository's collaborator logins. Using [owner] as a
+				// placeholder until a shared visibility/collaborators helper
+				// lands (tracked under copilot-mcp-core#1623).
+				var readers []string
```
JoannaaKL, line 1598: Let's populate the readers set too, for now we don't have to cache them (we can add it in a follow up pr). Repo collaborators will be fetched only for private repos

JoannaaKL (approved):

pachecocordovamoiseseduardo-byte (approved):

PR #2456, opened 2026-05-12
Add ifc label for search_issues tool

gokhanarkan:
Emits an IFC `SecurityLabel` on the `search_issues` tool result when the `InsidersMode` flag is enabled, mirroring the pattern landed for `get_me` in #2432, `list_issues` in #2453, and `get_file_contents` in #2454.

Refs github/copilot-mcp-core#1623, github/copilot-mcp-core#1389. One of the ingress tools listed in #1623's tool table.

> **Chained on #2454.** Base is currently `gokhanarkan/fides-get-file-contents` because this PR reuses the `FetchRepoIsPrivate` and `FetchRepoCollaborators` helpers added there. GitHub will auto-retarget the base to `main` once #2454 merges.

## What this PR does

Search results may span multiple repositories, so `_meta.ifc` for `search_issues` is the IFC join of the per-repository labels:

- **Integrity** is always `untrusted` (issues are user-authored).
- **Confidentiality** follows the lattice join (least upper bound):
  - If any matched repository is public → `["public"]` (the public side dominates).
  - Otherwise → **intersection** of the collaborator sets across all matched private repositories (a reader must have access to *every* matched private repo).
  - Empty result sets → `["public"]` (no data leaked).

If any per-repo visibility or collaborators lookup fails, the label is omitted entirely (consistent with `get_file_contents`) to avoid misclassifying the result.

## New shared helpers

- `ifc.LabelSearchIssues(repoVisibilities []bool, readerSets [][]string)` in `pkg/ifc/ifc.go` — pure join function, separately unit-tested.
- `searchHandler` in [pkg/github/search_utils.go](pkg/github/search_utils.go) gains an **additive variadic `searchOption` hook** so `SearchIssues` can attach `_meta.ifc` without duplicating the search call. `search_pull_requests` is unaffected — it passes no options and continues to behave exactly as before.

## Empty-intersection semantics (called out for reviewers)

If two private repos have **zero** overlapping collaborators, the joined reader set is `[]`. We emit `confidentiality: []` rather than skipping the label — this matches the IFC math (no one can read) and lets the client engine decide what to do. Worth a sanity-check from @JoannaaKL.

## Cost note

A search returning 30 results across 30 different private repos = 60 API calls (visibility + collaborators each). Bounded but not free; cache lands in a follow-up per the standing agreement.

## Tests

- `Test_LabelSearchIssues` in new [pkg/ifc/ifc_test.go](pkg/ifc/ifc_test.go) — table-driven coverage of the join math (empty, single public, mixed, intersecting private, no overlap, dedupe).
- `Test_SearchIssues_IFC_InsidersMode` in [pkg/github/issues_test.go](pkg/github/issues_test.go) with 6 subtests:
  2. Insiders on, single public repo → `integrity=untrusted`, `confidentiality=["public"]`.
  3. Insiders on, mixed public + private → `["public"]` (public wins).
  4. Insiders on, two private repos with intersecting collaborators → asserts the intersection.
  5. Insiders on, visibility lookup fails (500) → no `ifc` meta.
  6. Insiders on, empty results → `["public"]`.

## Validation


pachecocordovamoiseseduardo-byte (approved):

pkg/github/issues_test.go
```diff
@@ -693,6 +692,243 @@ func Test_SearchIssues(t *testing.T) {
 	}
 }
 
+func Test_SearchIssues_IFC_InsidersMode(t *testing.T) {
```
JoannaaKL, line 695: I love the tests you wrote!

pkg/github/issues.go
```diff
@@ -978,11 +976,115 @@ func SearchIssues(t translations.TranslationHelperFunc) inventory.ServerTool {
...
 			return result, nil, err
 		})
 }
 
+// searchIssuesIFCPostProcess returns a searchPostProcessFn that attaches the
+// IFC label for a search_issues result. It looks up the visibility (and, for
+// private repos, collaborators) of every repository represented in the search
+// payload and joins the labels via ifc.LabelSearchIssues. If any per-repo
+// lookup fails the label is omitted to avoid misclassifying the result.
+func searchIssuesIFCPostProcess(deps ToolDependencies) searchPostProcessFn {
```
JoannaaKL, line 993: We need an intersection of sets: if one issue has confidentiality public and another is private, private wins. For two issues with private visibility:
- issue 1, readers {X, Y, Z}
- issue 2, readers {X, A, B}
The readers set for the whole set will be {X}

pkg/github/issues_test.go
```diff
@@ -693,6 +692,243 @@ func Test_SearchIssues(t *testing.T) {
...
+		require.NoError(t, err)
+		require.False(t, result.IsError, "tool call should still succeed when collaborators lookup fails")
+
+		if result.Meta != nil {
+			_, hasIFC := result.Meta["ifc"]
+			assert.False(t, hasIFC, "ifc label should be omitted when collaborators lookup fails")
+		}
+	})
+
+	t.Run("insiders mode empty results emits public untrusted", func(t *testing.T) {
```
JoannaaKL, line 902: Or we can skip the ifc label. Nonblocking, let's address in a follow up

JoannaaKL (approved):

JoannaaKL (approved):

PR #2457, opened 2026-05-12
Add ifc label for issue_read tool

gokhanarkan:
Emits an IFC `SecurityLabel` on the `issue_read` tool result when the `InsidersMode` flag is enabled, mirroring the pattern landed for `get_me` in #2432, `list_issues` in #2453, `get_file_contents` in #2454, and `search_issues` in #2456.

Refs github/copilot-mcp-core#1623, github/copilot-mcp-core#1389. One of the ingress tools listed in #1623's tool table.

> **Chained on #2456** (which is itself chained on #2454). Base is currently `gokhanarkan/fides-search-issues` because this PR reuses `FetchRepoIsPrivate` and `FetchRepoCollaborators`. GitHub will auto-retarget the base to `main` as the upstream PRs merge.

## What this PR does

- Wires `_meta.ifc` onto the `issue_read` `CallToolResult` when `deps.GetFlags(ctx).InsidersMode` is true. No behaviour change when the flag is off.
- `issue_read` operates on a single issue in a single repository, so the label has the **same per-repo semantics as `list_issues`**: integrity always `untrusted`; public repos → `PublicUntrusted`; private repos → `PrivateUntrusted(collaborators)` with `[owner]` fallback if the collaborators lookup fails. `ifc.LabelListIssues` is reused directly — **no new `pkg/ifc` constructor**.
- The `IssueRead` handler dispatches to four sub-functions (`GetIssue`, `GetIssueComments`, `GetSubIssues`, `GetIssueLabels`). The IFC label is attached at the **dispatch site via a single `attachIFC` closure**, so all four `method=` branches emit the label without modifying the underlying helpers.
- Visibility-lookup failures cause the label to be omitted entirely (consistent with `get_file_contents` and `search_issues`) to avoid misclassifying the result.

## Suggested follow-up (not in this PR)

After this lands, three handlers (`get_file_contents`, `search_issues`, `issue_read`) carry near-identical `attachIFC` closures (~25 lines each). A small follow-up PR could extract them into a single shared helper (e.g. `AttachRepoIFC(ctx, deps, client, owner, repo, labelFn)` in `pkg/github/ifc_attach.go`), saving ~60 lines and keeping the pattern consistent across tools. Intentionally **not** bundled here to keep this PR minimal and easy to review.

## Tests

`Test_IssueRead_IFC_InsidersMode` in [pkg/github/issues_test.go](pkg/github/issues_test.go) with 4 subtests:
2. Insiders on, public repo, `method=get` → `integrity=untrusted`, `confidentiality=["public"]`.
3. Insiders on, private repo, `method=get_comments` → `untrusted`/collaborator list (proves the wrapper applies to non-`get` dispatch branches).
4. Insiders on, visibility lookup fails (500) → no `ifc` meta.

`get_labels` (GraphQL) gets the same wrapper coverage but isn't separately tested for IFC — the wrapper sits outside the dispatch and is method-agnostic.

## Validation


pachecocordovamoiseseduardo-byte (approved):

pkg/github/issues.go
```diff
@@ -296,19 +296,61 @@ Options are:
...
+			// repositories) collaborators lookups are performed lazily on
+			// first use. If the visibility lookup fails the label is omitted
+			// rather than misclassifying the result; the failure is not
+			// cached so a subsequent dispatch branch could retry. If only
+			// the collaborators lookup fails for a private repo we fall back
+			// to the owner so the reader set is never empty. The label
+			// matches list_issues semantics: per-repo visibility, integrity
+			// always untrusted.
+			var (
+				ifcLabelKnown bool
```
JoannaaKL: `attachIFC` is called exactly once per request and the switch on method picks one branch and returns immediately. The ifcLabelKnown / ifcIsPrivate / ifcReaders variables are never reused across calls, so the lazy-caching pattern adds complexity for no benefit. We can add a cache as a follow up though

JoannaaKL (approved):

PR #2469, opened 2026-05-13
Annotate add_issue_comment as destructive

gokhanarkan:
Sets `destructiveHint: true` on `add_issue_comment` so the IFC client engine can apply egress policies before invocation. First of four egress annotations from github/copilot-mcp-core#1623.


gokhanarkan:
no need

PR #2470, opened 2026-05-13
Annotate create_issue and issue_write as destructive

gokhanarkan:
Sets `destructiveHint: true` on the granular `create_issue` tool and the unified `issue_write` tool so the IFC client engine can apply egress policies regardless of which surface the client uses. Second of four egress annotations from github/copilot-mcp-core#1623.


gokhanarkan:
no need

PR #2472, opened 2026-05-13
Annotate create_pull_request as destructive

gokhanarkan:
Sets `destructiveHint: true` on `create_pull_request` so the IFC client engine can apply egress policies. Final of four egress annotations from github/copilot-mcp-core#1623.


gokhanarkan:
no need

PR #2478, opened 2026-05-14
Replace ingress IFC reader list with private marker

gokhanarkan:
Ingress IFC labels now emit a single `private` marker instead of the full collaborator list. The CLI engine fetches readers on demand at egress decision time (with pagination + caching), removing a wire-bloat ceiling for large private repos. Drops `FetchRepoCollaborators` from all five ingress hot paths. **Breaking wire-format change** for `_meta.ifc` — coordinate with the CLI cut-over. Refs github/copilot-mcp-core#1389.

### DIFF
--- a/pkg/github/repositories.go
+++ b/pkg/github/repositories.go
@@ -655,10 +655,14 @@ func CreateRepository(t translations.TranslationHelperFunc) inventory.ServerTool
 }
 
 // FetchRepoCollaborators returns the login names of all collaborators on a
-// repository. It is provided as a shared helper for IFC label computation so
-// tools can populate the reader set for private repositories. The full list
-// is fetched eagerly via pagination; callers are expected to invoke this only
-// when needed (e.g. private repos under InsidersMode).
+// repository, paginated.
+//
+// The github-mcp-server itself no longer calls this from its IFC label paths:
+// ingress tools emit a single "private" marker and the client engine resolves
+// concrete readers from the GitHub API on demand at egress decision time
+// (where it can paginate + cache). The helper is retained as an exported
+// utility for external library consumers that may want the same one-shot
+// pagination behaviour.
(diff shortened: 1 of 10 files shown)


pkg/github/repositories.go
```diff
@@ -655,10 +655,14 @@ func CreateRepository(t translations.TranslationHelperFunc) inventory.ServerTool
...
-// when needed (e.g. private repos under InsidersMode).
+// repository, paginated.
+//
+// The github-mcp-server itself no longer calls this from its IFC label paths:
+// ingress tools emit a single "private" marker and the client engine resolves
+// concrete readers from the GitHub API on demand at egress decision time
+// (where it can paginate + cache). The helper is retained as an exported
+// utility for external library consumers that may want the same one-shot
+// pagination behaviour.
 func FetchRepoCollaborators(ctx context.Context, client *github.Client, owner, repo string) ([]string, error) {
```
JoannaaKL: we can remove it too :)

pkg/github/search_test.go
```diff
@@ -283,48 +256,7 @@ func Test_SearchRepositories_IFC_InsidersMode(t *testing.T) {
...
-		request := createMCPRequest(reqParams)
-		result, err := handler(ContextWithDeps(context.Background(), deps), &request)
-		require.NoError(t, err)
-		require.False(t, result.IsError, "tool call should still succeed when collaborators lookup fails")
-
-		if result.Meta != nil {
-			_, hasIFC := result.Meta["ifc"]
-			assert.False(t, hasIFC, "ifc label should be omitted when collaborators lookup fails")
-		}
+		assert.Equal(t, []any{"private"}, ifcMap["confidentiality"])
```
JoannaaKL: Confidentiality is an array but it's not necessary anymore, we can just use a plain field. It'll be only public/private :)

JoannaaKL (approved):

PR #2482, opened 2026-05-15
fix: make job logs failed_only a modifier

### LINKED ISSUE #2389: get_job_logs: run_id with failed_only=false should return all job logs; failed_only should be a consistent modifier
This came out of an analysis of github MCP server errors encountered in my past sessions across 5000 sessions.

### What happened

An agent trying to retrieve all job logs for a workflow run passed:

```json
{"owner":"danmoseley","repo":"pr-dashboard","run_id":22854416647,"return_content":true,"tail_lines":500,"failed_only":false}
```

And got:

```
job_id is required when failed_only is false
```

This happened 6 times across 5 sessions. The agent's intent was clear and reasonable: get all logs for this run, not just failed ones. It even explicitly set `failed_only=false` to signal that — and got an error.

### Why the current design is confusing for agents

`failed_only` is used as a mode switch rather than a modifier:
- `run_id` without `failed_only=true` always errors, even though "get logs for this run" is a valid request
- `failed_only` has no effect when `job_id` is provided — it is silently ignored

### Proposed behavior

`failed_only` should be a consistent modifier on whichever ID is provided:

| Parameters | Result | Proposed |
|---|---|---|
| `job_id` + `failed_only=false` (or not passed) | logs for that job | **already allowed** |
| `job_id` + `failed_only=true` | logs for that job if it failed; `isError:true` with status if it succeeded | **error -> allowed**  |
| `run_id` + `failed_only=false` (or not passed) | logs for all jobs in the run | **already allowed** |
| `run_id` + `failed_only=true` | logs for failed jobs only | **error -> allowed** |
| both `job_id` and `run_id` | `isError:true` — provide one or the other, not both | **error** |
| neither | `isError:true` — one of `job_id` or `run_id` must be provided | **error** |

The `job_id + failed_only=true` case requires checking the job's conclusion before fetching logs, but the tool already fetches job metadata to get the log URL so this is a small addition.

### Breaking change note

No changes to existing successful calls. This only makes certain calls that are failing begin to work.

he-yufeng:
## Summary
- allow `get_job_logs` to fetch all jobs in a run when `run_id` is provided without `failed_only`
- make `failed_only` a modifier for both `run_id` and `job_id`
- reject ambiguous calls that provide both `job_id` and `run_id`
- keep `failed_jobs` in run-level responses for compatibility; it now reports the number of failed jobs even when returning all jobs
- update the `get_job_logs` tool snapshot so the public schema change is explicit

Fixes #2389.

## Notes
- Passing both `job_id` and `run_id` now returns a validation error instead of relying on the old precedence rule.

## To verify
- `go test ./pkg/github -run 'Test_ActionsGetJobLogs' -count=1`
- `go test ./... -count=1`
- `git diff --check`

### DIFF
--- a/README.md
+++ b/README.md
@@ -628,12 +628,12 @@ The following sets of tools are available:
 
 - **get_job_logs** - Get GitHub Actions workflow job logs
   - **Required OAuth Scopes**: `repo`
-  - `failed_only`: When true, gets logs for all failed jobs in the workflow run specified by run_id. Requires run_id to be provided. (boolean, optional)
-  - `job_id`: The unique identifier of the workflow job. Required when getting logs for a single job. (number, optional)
+  - `failed_only`: When true, only returns logs for failed jobs. With job_id, the job must have failed. With run_id, only failed jobs in the run are included. (boolean, optional)
+  - `job_id`: The unique identifier of the workflow job. Provide either job_id or run_id, not both. (number, optional)
   - `owner`: Repository owner (string, required)
   - `repo`: Repository name (string, required)
   - `return_content`: Returns actual log content instead of URLs (boolean, optional)
-  - `run_id`: The unique identifier of the workflow run. Required when failed_only is true to get logs for all failed jobs in the run. (number, optional)
+  - `run_id`: The unique identifier of the workflow run. Provide either run_id or job_id, not both. (number, optional)
   - `tail_lines`: Number of lines to return from the end of the log (number, optional)
 
 </details>
--- a/pkg/github/__toolsnaps__/get_job_logs.snap
+++ b/pkg/github/__toolsnaps__/get_job_logs.snap
@@ -1,17 +1,17 @@
 {
   "annotations": {
     "readOnlyHint": true,
-    "title": "Get job logs"
+    "title": "Get GitHub Actions workflow job logs"
   },
-  "description": "Download logs for a specific workflow job or efficiently get all failed job logs for a workflow run",
+  "description": "Get logs for GitHub Actions workflow jobs.\nUse this tool to retrieve logs for a specific job or all jobs in a workflow run.\nProvide job_id for one job, or run_id for all jobs in a run. Set failed_only=true to restrict either mode to failed jobs only.\n",
   "inputSchema": {
     "properties": {
       "failed_only": {
-        "description": "When true, gets logs for all failed jobs in run_id",
+        "description": "When true, only returns logs for failed jobs. With job_id, the job must have failed. With run_id, only failed jobs in the run are included.",
         "type": "boolean"
       },
       "job_id": {
-        "description": "The unique identifier of the workflow job (required for single job logs)",
+        "description": "The unique identifier of the workflow job. Provide either job_id or run_id, not both.",
         "type": "number"
       },
       "owner": {
@@ -27,7 +27,7 @@
         "type": "boolean"
       },
       "run_id": {
-        "description": "Workflow run ID (required when using failed_only)",
+        "description": "The unique identifier of the workflow run. Provide either run_id or job_id, not both.",
         "type": "number"
       },
       "tail_lines": {
--- a/pkg/github/actions.go
+++ b/pkg/github/actions.go
@@ -46,8 +46,7 @@ const (
 	actionsMethodDeleteWorkflowRunLogs    = "delete_workflow_run_logs"
 )
 
-// handleFailedJobLogs gets logs for all failed jobs in a workflow run
-func handleFailedJobLogs(ctx context.Context, client *github.Client, owner, repo string, runID int64, returnContent bool, tailLines int, contentWindowSize int) (*mcp.CallToolResult, any, error) {
+func handleRunJobLogs(ctx context.Context, client *github.Client, owner, repo string, runID int64, failedOnly bool, returnContent bool, tailLines int, contentWindowSize int) (*mcp.CallToolResult, any, error) {
 	// First, get all jobs for the workflow run
 	jobs, resp, err := client.Actions.ListWorkflowJobs(ctx, owner, repo, runID, &github.ListWorkflowJobsOptions{
 		Filter: "latest",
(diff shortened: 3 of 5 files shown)


FuzzysTodd:
@FuzzysTodd become the expert read all documents and reason the final
soulful solution spirit to set it and forget it repaired

jluocsa (commented):
Nice cleanup of the `get_job_logs` parameter semantics — making `failed_only` a true modifier (instead of a mode selector implicitly tied to `run_id`) reads much more naturally from the agent side, and the validation error messages are now actually helpful. The new tests for the failed-only-on-`job_id` path and the both-ids/neither-id validation cases are exactly the right coverage.

A few things worth a look before merge:

### 1. The tool schema snapshot doesn't appear to be updated

Three things changed in the tool definition:

- Top-level `Description` string
- `job_id` parameter description
- `run_id` parameter description
- `failed_only` parameter description

…but the changed-files list (`pkg/github/actions.go`, `pkg/github/actions_test.go`, `pkg/github/helper_test.go`) doesn't include anything under `pkg/github/__toolsnaps__/`. Per the repo's snapshot-validation system, that should fail `go test ./pkg/github` with a diff against `__toolsnaps__/get_job_logs.snap` (or whatever the file is named for this tool).

The fix is just:

```bash
UPDATE_TOOLSNAPS=true go test ./pkg/github -run Test_ActionsGetJobLogs -count=1
```

…and committing the resulting `.snap` change. That also serves as a nice reviewable record of the public-surface change.

### 2. Subtle behavior change for `result["failed_jobs"]` clients

Before this PR, the `run_id` path always returned `failed_jobs` in the response map. After the PR, `failed_jobs` is only present when `failed_only=true`:

```go
if failedOnly {
    result["failed_jobs"] = len(selectedJobs)
}
```

That's a sensible shape change (it'd be weird to report `failed_jobs` when the user didn't ask to filter by failure), but it's still an output-schema break for any caller that was unconditionally reading that key on the run-level result. Worth a one-liner in the PR body / changelog so downstream users notice.

If you'd prefer to keep the key for backwards compatibility, you could always emit it and just compute it independently:

```go
failedJobsCount := 0
for _, job := range jobs.Jobs {
    if job.GetConclusion() == "failure" {
        failedJobsCount++
    }
}
result["failed_jobs"] = failedJobsCount
```

…but I don't feel strongly — the "only when asked" shape is arguably cleaner.

### 3. Previously-tolerated combination now errors

Before: passing both `job_id` and `run_id` with `failed_only=true` silently went down the failed-jobs-in-run path (because the old branch checked `failedOnly && runID > 0` first). After: that combination now returns a hard error (`"provide either job_id or run_id, not both"`).

This is the right behavior — that input was always ambiguous — but it is a strict-mode change for any agent that was passing extra fields and relying on the old precedence rule. Probably worth one bullet in the PR description acknowledging it. (Not a blocker.)

### 4. Tiny nit on `handleSingleJobLogs` resource handling

```go
if failedOnly {
    workflowJob, resp, err := client.Actions.GetWorkflowJobByID(ctx, owner, repo, jobID)
    if err != nil {
        return ghErrors.NewGitHubAPIErrorResponse(ctx, "failed to get workflow job", resp, err), nil, nil
    }
    defer func() { _ = resp.Body.Close() }()
    ...
}
```

The `defer` is inside the `if failedOnly` block, so it only schedules the close when we entered that branch — good. But the second `getJobLogData` call below also returns a `resp`; if we made it through the failed-only check, we still close that body via `defer` later in the function. Just double-check the existing close on the lower `resp` in `handleSingleJobLogs` still fires for the non-`failedOnly` path. The diff didn't touch it, so this is probably fine, but it'd be reassuring to confirm via `go vet` / a quick read.

### 5. Test for `failed_only=true` with `run_id` returning an empty selection

There's `"no failed jobs found"` coverage for the run-level branch, and the new test exercises the `job_id`+`failed_only=true`+success-conclusion error path. Consider one more test: `run_id` + `failed_only=true` where all jobs failed (currently only the partial-failure case is exercised in the existing test). Small, but exercises the message construction that was reshuffled.

---

Overall the change is straightforward and the API ergonomics improvement is well worth it. Mostly just want to make sure the toolsnap file is regenerated so the schema change is captured in the diff.

he-yufeng:
Thanks for the detailed pass. I pushed `fbcf77a` with the follow-up changes:

- regenerated and committed `pkg/github/__toolsnaps__/get_job_logs.snap`
- restored compatibility by always returning `failed_jobs` for run-level responses, even when `failed_only=false`
- kept the both-ids validation and added the behavior note to the PR body
- added coverage for `run_id` + `failed_only=true` when all jobs failed
- rechecked the single-job response handling path while updating the tests

Validation run locally:
- `go test ./pkg/github -run 'Test_ActionsGetJobLogs' -count=1`
- `go test ./... -count=1`
- `git diff --check origin/main..HEAD`

he-yufeng:
Pushed `e2d38d1` to refresh the generated README tool docs after the `get_job_logs` schema description changes. That should address the failing `docs-check` job.

Rechecked locally:

```bash
go run ./cmd/github-mcp-server generate-docs
go test ./pkg/github -run Test_ActionsGetJobLogs -count=1
git diff --check
```

he-yufeng:
Rebased this branch onto the latest `github:main` and reran the relevant checks after the generated docs refresh.

Validation:

```text
go test ./pkg/github -run Test_ActionsGetJobLogs -count=1
ok  github.com/github/github-mcp-server/pkg/github  0.220s

go test ./pkg/github -count=1
ok  github.com/github/github-mcp-server/pkg/github  0.906s

go test ./... -count=1
ok  ./... across the repository

git diff --check
```

h5dh2gy26z-sys (approved):

he-yufeng:
Rebased this approved branch onto latest `github:main` again and force-pushed `10df755`.

Validation after rebase:

```text
go test ./pkg/github -run Test_ActionsGetJobLogs -count=1
# ok

go test ./pkg/github -count=1
# ok

go test ./... -count=1
# ok

# passed
```

advancedresearcharray:
Rebased this branch onto latest `main` (`457f599`) and validated locally. The rebased commits are:

- `1c7f4b0` fix: make job logs failed_only a modifier
- `b866f6c` fix: update job logs schema snapshot
- `7022030` docs: update generated tool docs

Validation (Go 1.25):

```text
go test ./pkg/github -run Test_ActionsGetJobLogs -count=1  # ok
go test ./pkg/github -count=1                              # ok
go test ./... -count=1                                      # ok
go run ./cmd/github-mcp-server generate-docs                # no drift
git diff --check origin/main..HEAD                          # passed
```

The branch is currently **behind `main`** on the PR head (`10df755`). A force-push of `fix/actions-job-logs-run-modifier` with the rebased history will clear the merge blocker.

Rebased branch (for sync): https://github.com/advancedresearcharray/github-mcp-server/tree/fix/actions-job-logs-run-modifier

```bash
git fetch https://github.com/advancedresearcharray/github-mcp-server.git fix/actions-job-logs-run-modifier
git checkout fix/actions-job-logs-run-modifier
git push origin fix/actions-job-logs-run-modifier --force-with-lease
```

advancedresearcharray (commented):
Follow-up after rebase validation:

All items from the earlier review appear addressed in the current diff:
- toolsnap regenerated (`pkg/github/__toolsnaps__/get_job_logs.snap`)
- `failed_jobs` always returned for run-level responses (compatibility preserved)
- both-ids validation + PR body note documented
- `run_id` + `failed_only=true` all-failed test added
- generated README/docs refreshed

Rebased onto latest `main` (`457f599`) locally; all tests and docs checks pass. Branch still needs a force-push to clear the **behind main** merge blocker — see latest PR comment for the rebased branch sync URL.

he-yufeng:
Rebased this branch onto current `main` again.

Validated locally after the rebase:
- `go test ./...`
- `git diff --check origin/main..HEAD`

The rebased commits are now:
- `5ecd6f2` fix: make job logs failed_only a modifier
- `de9f34b` fix: update job logs schema snapshot
- `7ba5bd7` docs: update generated tool docs

he-yufeng:
Rebased this branch onto current `main` and resolved the actions job-log conflict by keeping the newer IFC label wrapper while preserving this PR's `failed_only` semantics for both `run_id` and `job_id`.

Validation after the rebase:

```text
go test ./...
```

Current head: `f8d1ebe`.

he-yufeng:
Closing this to keep things tidy as it hasn't attracted review. Turning `failed_only` into a proper log modifier is still a small ergonomic win if you'd like it; happy to reopen and refresh against main. Thanks!

PR #2503, opened 2026-05-19
fix(ui): advertise get_me as an app via _meta.ui.visibility

SamMorrowDrums:
## Bug

The `get_me` tool wires an MCP App UI resource (`ui://github-mcp-server/get-me`) via `_meta.ui.resourceUri`, but the `_meta.ui` map was missing the `visibility` array. MCP clients that key off `visibility` to decide whether to render the resource as an app (e.g. VS Code's MCP Apps renderer) silently skip `get_me` as a result.

## Fix

Add `visibility: ["model", "app"]` to `get_me`'s `_meta.ui`, matching the existing UI-enabled tools (`issue_write`, `create_pull_request`).

## Changes

- `pkg/github/context_tools.go` — one-line addition to the `_meta.ui` map literal in `GetMe`.
- `pkg/github/__toolsnaps__/get_me.snap` — regenerated via `UPDATE_TOOLSNAPS=true go test ./pkg/github -run 'Test_GetMe$'`.

`script/lint` and `script/test` both pass.

### DIFF
--- a/pkg/github/__toolsnaps__/get_me.snap
+++ b/pkg/github/__toolsnaps__/get_me.snap
@@ -1,7 +1,11 @@
 {
   "_meta": {
     "ui": {
-      "resourceUri": "ui://github-mcp-server/get-me"
+      "resourceUri": "ui://github-mcp-server/get-me",
+      "visibility": [
+        "model",
+        "app"
+      ]
     }
   },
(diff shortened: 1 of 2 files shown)


mattdholloway (approved):
lgtm!

PR #2505, opened 2026-05-19
Adding rationale for fields and labels in issues_granular

boazreicher:
## Summary
Modifies the issues_granular tool to support adding rationales for changes to an issue's labels or fields.

## Context

We decided that in order for our experiment to be more valuable we need to include these two other granular tools. We will not expand this capability beyond these tools before we decide to pursue this feature to GA and have a plan on implementation for all tools.

## Why

Closes https://github.com/github/plan-track-agentic-toolkit/issues/156

## What changed
- Using the same approach as in https://github.com/github/github-mcp-server/pull/2458

- **`pkg/github/issues_granular.go`** — Rewrote `GranularUpdateIssueLabels` as a standalone tool (no longer uses `issueUpdateTool` helper). When `rationale` is provided, sends the labels as an object `{"value": "...", "rationale": "..."}` via a raw PATCH request. When omitted, sends `{"type": "..."}` preserving existing behavior.  Also updated `GranularSetIssueFieldss`
- **`pkg/github/granular_tools_test.go`** — Expanded test to table-driven covering both type-only and type-with-rationale cases.
- Updated schema snapshots.


pkg/github/__toolsnaps__/update_issue_labels.snap
```diff
@@ -4,7 +4,7 @@
     "openWorldHint": true,
     "title": "Update Issue Labels"
   },
-  "description": "Update the labels of an existing issue. This replaces the current labels with the provided list.",
+  "description": "Update the labels of an existing issue. This replaces the current labels with the provided list. Each label may be a plain string or an object with a 'name' and optional 'rationale' explaining why the label was chosen.",
```
alondahari: I don't think we should add the tool description, it should be enough to have the rationale parameter present

pkg/github/__toolsnaps__/update_issue_labels.snap
```diff
@@ -13,9 +13,31 @@
         "type": "number"
       },
       "labels": {
-        "description": "Labels to apply to this issue",
+        "description": "Labels to apply to this issue. Each item is either a label name string, or an object with 'name' and an optional 'rationale' (max 280 characters) explaining what specifically about the issue led you to choose this label.",
```
alondahari: same here, I don't think this change is needed

alondahari (dismissed):
Looks good, just a comment on superfluous descriptions

alondahari (approved):

SamMorrowDrums (approved):

PR #2526, opened 2026-05-22
feat(issue-fields): expose fullDatabaseId (BigInt) in list_issue_fields

iulia-b:
- https://github.com/github/github-mcp-server/pull/2492 depends on this
- depends on https://github.com/github/github/pull/433228

## Summary

- Add DatabaseID (int64) to IssueField struct, populated from fullDatabaseId BigInt scalar on all 4 concrete GQL union types
- Repeat fullDatabaseId per union fragment (shurcooL/githubv4 cannot use interface-level fragments at union top-level)
- Add parseFullDatabaseID helper to parse BigInt string to int64
- Update tests to assert DatabaseID is populated from fullDatabaseId

## Why

Fixes #

We are using this tool in two places:
1. Set issue fields granular tool, which is using the GraphQL endpoint to update issue fields
2. Issues write, which is using the REST endpoint to update issue fields on an issue
Because of this, we need to have both node ID and database ID exposed in this tool so both other tools can use it.

## MCP impact

- [ ] No tool or API changes
- [ ] New tool added

- 

## Security / limits

## Tool renaming

## Lint & tests

## Docs

- [ ] Not needed

### DIFF
--- a/pkg/github/issue_fields.go
+++ b/pkg/github/issue_fields.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"strconv"
 
 	ghcontext "github.com/github/github-mcp-server/pkg/context"
 	ghErrors "github.com/github/github-mcp-server/pkg/errors"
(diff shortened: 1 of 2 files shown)


pachecocordovamoiseseduardo-byte (approved):

iulia-b:
Closing as it was addressed  through https://github.com/github/github-mcp-server/pull/2551/

PR #2528, opened 2026-05-22

### LINKED ISSUE #2519: MCP App UI resources for write tools are registered in --read-only mode
## Summary

## Where

```go
}
```

## Why it matters

## Suggested fix

Two options:

## Related

he-yufeng:
Fixes #2519

## Summary
- pass read-only mode into MCP App UI resource registration
- keep get_me_ui available in read-only mode
- skip issue_write_ui and pr_write_ui when write tools are filtered out

## To verify
- go test ./pkg/github -run TestRegisterUIResources -count=1
- go test ./cmd/github-mcp-server ./pkg/github
- go test ./internal/ghmcp ./pkg/github
- git diff --check


pachecocordovamoiseseduardo-byte (approved):

SamMorrowDrums:
I was indeed thinking about this recently too. Thanks for doing it. Will take a look.

SamMorrowDrums:
Closing in favor of #2574, which is your newer PR covering the same change. Thanks!

PR #2548, opened 2026-05-26
Adding suggest bool flag to suggest type changes to an issue

boazreicher:
## Summary
Modifies the issues_granular tool to support a boolean `suggest` flag to the update issue type granular tool.

## Context

Allows agents to only suggest a change to an issue's type, and a user can then choose whether to reject or apply it.

## Why

Closes https://github.com/github/plan-track-agentic-toolkit/issues/128

### DIFF
--- a/pkg/github/issues_granular.go
+++ b/pkg/github/issues_granular.go
@@ -512,6 +512,11 @@ func GranularUpdateIssueType(t translations.TranslationHelperFunc) inventory.Ser
 							"State the concrete signal (e.g. 'Reports a crash when saving' → bug, 'Asks for dark mode support' → feature).",
 						MaxLength: jsonschema.Ptr(280),
 					},
+					"suggest": {
+						Type: "boolean",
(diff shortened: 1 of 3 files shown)


pkg/github/issues_granular.go
```diff
@@ -512,6 +512,11 @@ func GranularUpdateIssueType(t translations.TranslationHelperFunc) inventory.Ser
 							"State the concrete signal (e.g. 'Reports a crash when saving' → bug, 'Asks for dark mode support' → feature).",
 						MaxLength: jsonschema.Ptr(280),
 					},
+					"suggest": {
```
SamMorrowDrums:
```suggestion
					"is_suggestion": {
```

Perhaps this should be more clear in language as a name, it was unclear to me what it meant. I had to re-read a couple of times. Probably for LLMs they also could do with more clarity.

SamMorrowDrums (commented):
Minor comment on field name is all.

SamMorrowDrums (approved):

PR #2553, opened 2026-05-26
feat: gate issue_write and get_issue behind remote_mcp_issue_fields flag

kelsey-myers:
## Summary

Gates `issue_write` and `get_issue` field enrichment behind the `remote_mcp_issue_fields` feature flag. Stacks on top of #2551.

Extends the inventory `FeatureFlagEnable`/`FeatureFlagDisable` fields from `string` to `[]string` (AND semantics for enable, OR for disable) to avoid 4-way tool duplication when gating against multiple flags simultaneously.

## Why

`issue_write` exposes `issue_fields` (Issues 2.0) and `get_issue` runs an extra GraphQL round-trip for field values — both should only happen when the flag is on, consistent with how `list_issues` and `search_issues` were gated in #2520.

## What changed

- `FeatureFlagEnable`/`FeatureFlagDisable` on `ServerTool`, `ServerResourceTemplate`, `ServerPrompt` changed from `string` to `[]string`; `featureFlagAllowed` updated with range loops (enable = all must pass, disable = any blocks)
- `LegacyIssueWrite`: new `FeatureFlagDisable = []string{FeatureFlagIssuesGranular, FeatureFlagIssueFields}` variant — no `issue_fields` schema param, passes `nil` field values
- `IssueWrite`: gains `FeatureFlagEnable = []string{FeatureFlagIssueFields}` — only active when flag is on
- `get_issue` field_values enrichment wrapped in `IsFeatureEnabled(ctx, FeatureFlagIssueFields)` runtime check
- New toolsnaps: `issue_write.snap` (legacy) + `issue_write_ff_remote_mcp_issue_fields.snap` (flag-enabled), following #2520 convention
- `featureFlagAllowed` disable loop modernised to `slices.ContainsFunc`

## MCP impact

- [x] Tool schema or behavior changed — `issue_write` schema splits into two registrations; `issue_fields` param only visible when flag is on

## Prompts tested (tool changes only)

- "Create an issue in github/planning-tracking-demo with Priority P2 and Visibility Medium" → `issue_write` with `issue_fields` ✅
- "Get issue 1992 in github/planning-tracking-demo" → `get_issue` returns `field_values: [{Priority: P2}, {Visibility: Medium}]` ✅

## Security / limits

- [x] No security or limits impact

## Tool renaming

- [x] I am not renaming tools as part of this PR

## Lint & tests

- [x] Linted locally with `./script/lint`
- [x] Tested locally with `./script/test`

## Docs

- [x] Updated (README / docs / examples)

PR #2558, opened 2026-05-27
feat: gate issue_write and get_issue behind remote_mcp_issue_fields flag

SamMorrowDrums:
## Summary

Ports the gating work from #2553 onto `main`. #2553 was merged into the `kelsey/issue-fields-clean` stack base, but that stack base never landed on main — so `issue_write` and `get_issue` are currently ungated on main even though `list_issues`, `search_issues`, and `list_issue_fields` are.

Picks the simpler half of #2553's inventory change: keep `FeatureFlagEnable` as a single `string`, only promote `FeatureFlagDisable` to `[]string` (OR-of-any). The AND-on-enable shape in the original PR had no real use site and encoded dependencies rather than rollout, so I dropped it.

## What changed

- `pkg/inventory`: `FeatureFlagDisable` on `ServerTool`/`ServerResourceTemplate`/`ServerPrompt` becomes `[]string` with OR semantics (any listed flag enabled → tool hidden). `FeatureFlagEnable` stays a single string.
- `pkg/github/issues.go`:
  - Split into `IssueWrite` (`FeatureFlagEnable = FeatureFlagIssueFields`, exposes `issue_fields`) and `LegacyIssueWrite` (`FeatureFlagDisable = {FeatureFlagIssuesGranular, FeatureFlagIssueFields}`, omits `issue_fields`).
  - Both register as `issue_write`; mutually exclusive annotations pick exactly one at runtime.
  - Refactored into a shared `buildIssueWrite(t, includeIssueFields)` helper instead of duplicating the ~250-line tool definition.
  - `GetIssue` field-values enrichment now wrapped in a runtime `IsFeatureEnabled(ctx, FeatureFlagIssueFields)` check; the verbose REST `IssueFieldValues` is always cleared.
- Existing single-flag `FeatureFlagDisable` call sites converted to slice literals.
- New toolsnap: `issue_write_ff_remote_mcp_issue_fields.snap`. The canonical `issue_write.snap` is now owned by `LegacyIssueWrite` (no `issue_fields` property).
- README, `docs/feature-flags.md`, `docs/insiders-features.md` regenerated.

## MCP impact

- Flag OFF: `issue_write` schema has no `issue_fields` property; `get_issue` response has no `field_values` and no raw `issue_field_values`. Behaviour matches pre-Issues-2.0.
- Flag ON: `issue_write` accepts `issue_fields` with `value` / `field_option_name` / `delete: true` (fetch+merge semantics from #2551 preserved); `get_issue` returns the enriched `field_values`.

## Verification

Smoke-tested live against `github/plan-track-agentic-toolkit#33`:

```
WITHOUT --features:                         WITH --features remote_mcp_issue_fields:
  issue_write: issue_fields=False             issue_write: issue_fields=True
  list_issue_fields: hidden                   list_issue_fields: exposed
```

Update via `issue_fields: [{field_name: "Priority", field_option_name: "P2"}]` against an issue with `Priority=P1, Effort=Low` produced `Priority=P2, Effort=Low` — non-destructive merge from #2551 confirmed still working under the gated path.

`script/lint`, `script/test` (with race), and live MCP stdio calls all pass.

### DIFF
--- a/pkg/github/issues.go
+++ b/pkg/github/issues.go
@@ -460,16 +460,19 @@ func fetchExistingIssueFieldValues(ctx context.Context, gqlClient *githubv4.Clie
 
 // mergeIssueFieldValues returns a merged slice where incoming values override existing ones
 // for the same field ID, and existing fields not present in incoming are preserved.
+// Ordering is deterministic: incoming entries first in their original order, followed by any
+// existing entries (in their original order) whose field IDs weren't seen in incoming.
 func mergeIssueFieldValues(existing, incoming []*github.IssueRequestFieldValue) []*github.IssueRequestFieldValue {
-	merged := make(map[int64]*github.IssueRequestFieldValue, len(existing)+len(incoming))
-	for _, v := range existing {
-		merged[v.FieldID] = v
-	}
+	seen := make(map[int64]struct{}, len(incoming))
+	result := make([]*github.IssueRequestFieldValue, 0, len(existing)+len(incoming))
 	for _, v := range incoming {
-		merged[v.FieldID] = v
+		seen[v.FieldID] = struct{}{}
+		result = append(result, v)
 	}
-	result := make([]*github.IssueRequestFieldValue, 0, len(merged))
-	for _, v := range merged {
+	for _, v := range existing {
+		if _, ok := seen[v.FieldID]; ok {
+			continue
+		}
 		result = append(result, v)
 	}
 	return result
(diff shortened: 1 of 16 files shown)


pkg/github/issues.go
```diff
@@ -2013,7 +2029,228 @@ Options are:
 				return utils.NewToolResultError("invalid method, must be either 'create' or 'update'"), nil, nil
 			}
 		})
-	st.FeatureFlagDisable = FeatureFlagIssuesGranular
+	st.FeatureFlagEnable = FeatureFlagIssueFields
```
RossTarrant, line 2032: Do we need to add this flag to `set_issue_fields` as well? It currently is still available when `issues_granular` is enabled. Does that need to also have the `remote_mcp_issue_fields` flag enabled, so that issue field writes are consistently behind the same flag?

SamMorrowDrums: Issues granular I'm ok with having it alreayd.

PR #2574, opened 2026-05-29

he-yufeng:
## Summary
- pass the static read-only setting into MCP App UI resource registration
- keep `get_me_ui` available in read-only mode but skip write-form resources for issues and pull requests
- add a regression test that lists registered UI resources in read-only mode

## To verify
- `go test ./pkg/github -run 'TestRegisterUIResources|TestNewMCPServer_RegistersUIResources' -count=1`
- `go test ./internal/ghmcp -count=1`
- `go test ./...`
- `git diff --check`


he-yufeng:
Superseded by #2581, which carries the current version of the read-only UI resource fix.

PR #2581, opened 2026-05-30

### LINKED ISSUE #2519: MCP App UI resources for write tools are registered in --read-only mode

he-yufeng:
﻿## Summary
- pass the server read-only setting into MCP App UI resource registration
- keep the read-only `get_me_ui` resource available
- skip `issue_write_ui` and `pr_write_ui` when write tools are filtered by `--read-only`

Fixes #2519.

## To verify
- `go test ./pkg/github -run "TestRegisterUIResources_ReadOnlySkipsWriteResources|TestRegisterUIResources_ReadableViaClient|TestNewMCPServer_RegistersUIResources" -count=1`
- `git diff --check`

### DIFF
--- a/pkg/github/server.go
+++ b/pkg/github/server.go
@@ -111,7 +111,7 @@ func NewMCPServer(ctx context.Context, cfg *MCPServerConfig, deps ToolDependenci
 	// remote/HTTP server also serves them, fixing the "-32002 Resource not
 	// found" error clients hit after the tool returns a ui:// URI.
 	if UIAssetsAvailable() {
-		RegisterUIResources(ghServer)
+		RegisterUIResources(ghServer, cfg.ReadOnly)
 	}
 
 	return ghServer, nil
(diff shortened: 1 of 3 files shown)


he-yufeng:
﻿Validated the current branch locally after closing the superseded #2574:

```
go test ./pkg/github -run UIResources
git diff --check origin/main...HEAD
```

he-yufeng:
Rebased on current `main` in `b8743ed`.

Validation:
- `go test ./pkg/github -run UIResources`
- `git diff --check origin/main...HEAD`

he-yufeng:
﻿Rebased on current `main` and reran the targeted validation locally.

Validated:

```powershell
gofmt -w $(git diff --name-only origin/main...HEAD | Where-Object { $_ -like '*.go' })
go test ./pkg/github -run UIResources
go test ./pkg/github
git diff --check origin/main...HEAD
```

All passed after the rebase.

he-yufeng:
Closing this older duplicate branch in favor of #2612, which carries the same read-only UI resources fix on the newer replacement branch. Keeping only #2612 avoids splitting review across two identical PRs.

PR #2591, opened 2026-06-02
fix: let get_me handle missing arguments

### LINKED ISSUE #2587: bug: `get_me` tool fails without parameters due to empty JSON object `{}` not being accepted as valid input
## Description

## Steps to Reproduce

   are required):
```json
   {
     "jsonrpc": "2.0",
     "id": 1,
     "method": "tools/call",
     "params": {
       "name": "get_me",
       "arguments": {}
     }
   }
```
```json
   {
     "jsonrpc": "2.0",
     "id": 2,
     "method": "tools/call",
     "params": {
       "name": "get_me",
       "arguments": { "_": "" }
     }
   }
```

## Expected Behavior

## Actual Behavior

## Root Cause (Hypothesis)

`{}`.

Two possible causes:

   early for the no-param case.

## Environment

## Suggested Fix

```go
```

## Additional Context

he-yufeng:
Fixes #2587.

## Summary

`get_me` is a zero-argument tool, but it was registered through the generic argument-decoding helper. That still makes the call sensitive to whether a client sends `{}` or omits `arguments` entirely.

This patch registers `get_me` with the no-argument handler path instead. The handler does not need parsed arguments, so it can serve both forms consistently:

- `arguments: {}`
- no `arguments` field

## Validation

- `go test ./pkg/github -run Test_GetMe -count=1`
- `go test ./... -count=1`
- `gofmt -w pkg/github/context_tools.go pkg/github/context_tools_test.go`
- `git diff --check origin/main..HEAD`

### DIFF
--- a/pkg/github/context_tools.go
+++ b/pkg/github/context_tools.go
@@ -43,7 +43,7 @@ type UserDetails struct {
 
 // GetMe creates a tool to get details of the authenticated user.
 func GetMe(t translations.TranslationHelperFunc) inventory.ServerTool {
-	return NewTool(
+	return NewToolFromHandler(
 		ToolsetMetadataContext,
 		mcp.Tool{
 			Name:        "get_me",
@@ -63,10 +63,10 @@ func GetMe(t translations.TranslationHelperFunc) inventory.ServerTool {
 			},
 		},
 		nil,
-		func(ctx context.Context, deps ToolDependencies, _ *mcp.CallToolRequest, _ map[string]any) (*mcp.CallToolResult, any, error) {
+		func(ctx context.Context, deps ToolDependencies, _ *mcp.CallToolRequest) (*mcp.CallToolResult, error) {
 			client, err := deps.GetClient(ctx)
 			if err != nil {
-				return utils.NewToolResultErrorFromErr("failed to get GitHub client", err), nil, nil
+				return utils.NewToolResultErrorFromErr("failed to get GitHub client", err), nil
 			}
 
 			user, res, err := client.Users.Get(ctx, "")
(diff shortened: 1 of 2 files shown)


he-yufeng:
﻿Updated this branch onto current `origin/main`.

Validated locally:

```text
go test ./pkg/github -run Test_GetMe
go test ./pkg/github
git diff --check
```

he-yufeng:
Rebased this branch on the latest upstream `origin/main` and reran validation locally.

Validated:

```text
go test ./pkg/github -run Test_GetMe
go test ./pkg/github
go test ./...
```

All passed.

he-yufeng:
Re-ran validation on the current branch:\n\n- \go test ./pkg/github -run Test_GetMe -count=1\\n- \go test ./pkg/github -count=1\\n- \go test ./...\\n- \git diff --check origin/main...HEAD\\n\nAll passed locally on Windows.

he-yufeng:
Rebased this branch onto the latest upstream `main` and force-pushed the cleaned head.

Validated locally:
- `go test ./pkg/github -run Test_GetMe -count=1`
- `go test ./...`
- `git diff --check origin/main..HEAD`

CZNetworks (approved):

he-yufeng:
Rebased this branch onto current upstream `main` again; the PR remains a one-commit, two-file change.

Validation on Windows:

```text
go test ./pkg/github -run Test_GetMe -count=1
go test ./pkg/github -count=1
go test ./...
```

All passed.

he-yufeng:
Rebased the approved branch onto current upstream `main` and resolved the conflict by keeping the newer `attachStaticIFCLabel` path while preserving the missing-arguments handling for `get_me`.

Validation on Windows:
- `go test ./pkg/github -run Test_GetMe -count=1`
- `go test ./...`
- `git diff --check origin/main..HEAD`

he-yufeng:
First real bump after several quiet rebases. context_tools.go is unchanged on current main, so a get_me call with a garbage argument still trips the same path this guards. It stays a two-file change and has been rebased cleanly onto main again this week. If the tool is intentionally strict about unexpected arguments, say so and I will close it instead.

PR #2600, opened 2026-06-03
fix: Empty assignees array should clear assignees

### LINKED ISSUE #2596: issue_write tool ignore parameters set to default values (e.g. empty array)
### Describe the bug

When issue_write is called with 'assignees' set to empty array, the tool behaves as if the parameter was not provided at all (ignores it). This results in inability to e.g. unassign an issue from user. 

Same issue is observed when trying to set 'labels' to empty array. 

### Affected version

Latest, connected to remote GitHub MCP server via MCP inspector.

1. Create an issue.
2. Assign issue to user by calling issue_write with 'assignees' parameter set to [ 'anna-kogai' ] 
3. Unassign issue from user by calling issue_write with 'assignees' parameter set to []
4. Observe tool reporting success

Expected behavior: The list of assignees is empty. 
Actual behavior: Assignees is still set to [ 'anna-kogai' ]

### Logs

Paste any available logs. Redact if needed.

RossTarrant:
## Summary
 Fixes `issue_write` updates so explicitly empty `assignees` or `labels` arrays clear existing values instead of being ignored.
 
 ## Why
 `issue_write` previously treated empty arrays the same as omitted parameters, making it impossible to unassign all users or remove all labels from an issue.
 
 Fixes #2596
 
 ## What changed
 - Track whether `assignees` and `labels` were explicitly provided in `issue_write` requests.
 - Pass presence information into `UpdateIssue` via `UpdateIssueOptions`.
 - Send empty `assignees` / `labels` arrays to the GitHub Issues API when explicitly provided.
 - Added regression tests for both feature-flagged and legacy `issue_write` update paths.
 
 ## MCP impact
 - [ ] No tool or API changes
   - `issue_write` update behavior now honors explicitly empty `assignees` and `labels` arrays.
 - [ ] New tool added
 
 - "Unassign everyone from issue #8 in rosstarrant/octopus-mcp"
 - "Remove all labels from issue #8 in rosstarrant/octopus-mcp"
 
 ## Security / limits
   - Uses the existing issue update permissions; no new scopes are required.
 
 ## Tool renaming
 
 ## Lint & tests
 
 ## Docs
 - [x] Not needed

### DIFF
--- a/pkg/github/issues.go
+++ b/pkg/github/issues.go
@@ -1793,11 +1793,6 @@ func issueWriteHasNonFormParams(args map[string]any) bool {
 	return false
 }
 
-func issueWriteHasNonNilParam(args map[string]any, key string) bool {
-	value, ok := args[key]
-	return ok && value != nil
-}
-
 // IssueWrite is the FeatureFlagIssueFields-enabled variant of issue_write
 // (with the issue_fields parameter). LegacyIssueWrite is served when the flag
 // is off. Both register under the tool name "issue_write"; exactly one is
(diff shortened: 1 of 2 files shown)


pkg/github/issues.go
```diff
@@ -1793,6 +1793,11 @@ func issueWriteHasNonFormParams(args map[string]any) bool {
 	return false
 }
 
+func issueWriteHasNonNilParam(args map[string]any, key string) bool {
+	value, ok := args[key]
+	return ok && value != nil
+}
```
SamMorrowDrums: Seems a bit redundant to inline the helper.

It's both too specifically named for a generic operation and duplication to both the variants is also fine because in the end we will only be keeping one pathway.

RossTarrant: Thanks for the feedback, very fair point! Have moved the logic to be inline now with rest of logic

SamMorrowDrums (approved):
Nice! 👌

PR #2605, opened 2026-06-04
Add confidence parameter to issue mutation MCP tools

alondahari:
## Summary

Adds an optional `confidence` parameter to the three issue mutation MCP tools: `update_issue_type`, `update_issue_labels`, and `set_issue_fields`. The confidence level is a string enum (`"low"`, `"medium"`, `"high"`) passed through to the REST/GraphQL API on every mutation call.

## Changes

- **Renamed structs** from `WithRationale` to `WithIntent` (`labelWithIntent`, `issueTypeWithIntent`) to better reflect they carry multiple intent signals
- **Added `confidence` schema property** as a string enum with values `low`, `medium`, `high`:
  - `high`: clear signal or explicit user request
  - `medium`: reasonable inference with some ambiguity
  - `low`: best guess with limited signal
- **Updated tool descriptions** to encourage LLMs to include confidence levels
- **Passes confidence** in the API request body alongside rationale/suggest
- **Updated toolsnaps, tests, and docs**

## Testing

- All unit tests pass (including new confidence-specific tests for all 3 tools)
- Lint clean
- Toolsnaps regenerated

Closes github/plan-track-agentic-toolkit#219


docs/feature-flags.md
```diff
@@ -240,7 +241,7 @@ runtime behavior (such as output formatting) won't appear here.
   - `owner`: Repository owner (username or organization) (string, required)
   - `pullNumber`: The pull request number (number, required)
   - `repo`: Repository name (string, required)
-  - `reviewers`: GitHub usernames to request reviews from (string[], required)
+  - `reviewers`: GitHub usernames or ORG/team-slug team reviewers to request reviews from (string[], required)
```
alondahari, line 244: This seems to come from an unrelated change that probably didn't update the docs

docs/feature-flags.md
```diff
@@ -198,6 +198,7 @@ runtime behavior (such as output formatting) won't appear here.
 
 - **update_issue_type** - Update Issue Type
   - **Required OAuth Scopes**: `repo`
+  - `confidence`: How confident you are in this choice (0–100). 90–100: very high — clear signal or explicit user request. 70–89: high — strong signals, likely correct. 50–69: moderate — reasonable inference but ambiguous. 30–49: low — best guess, user review recommended. 0–29: very low — speculative. (number, optional)
```
boazreicher: What's the value of adding these range categories?
They seem somewhat arbitrary and don't actually exist anywhere in the code.
We could think about the confidence be something like an enum (with values mapped to named categories), but since we're using a numeric range, simply overlaying arbitrary ranges (which aren't enforced/referenced anywhere in the code) doesn't seem like it makes a lot of sense

alondahari: The idea is that we will have the same buckets as a guide for maintainers when setting the thresholds. I'm afraid that without any guidance it would be totally unpredictable. Do you think using an enum instead would be better?

boazreicher: i think so.  i don't really see a reason to care about a difference between a confidence value of `74` vs `76`...
I think a few (named) categories would make more sense.
we could keep the 0-100 range on the backend, and use categories in all of the API/UI endpoints (which we would translate to a 0-100 number).  That way we can add more fine-grained categories later if we want.

h5dh2gy26z-sys (approved):

omgitsads (approved):

PR #2607, opened 2026-06-04
fix: accept empty/missing arguments for zero-param tools (get_me)

### LINKED ISSUE #2587: bug: `get_me` tool fails without parameters due to empty JSON object `{}` not being accepted as valid input

advancedresearcharray:
## Summary
- Treat nil or empty `arguments` on `tools/call` as `{}` before JSON unmarshaling
- Fixes `get_me` and other zero-parameter tools when clients omit `arguments` entirely

Fixes #2587

## Test plan
- [x] `go test ./pkg/inventory/...`
- [x] `TestNewServerToolWithContextHandler_EmptyArguments_TreatedAsEmptyObject` (nil, empty byte slice, explicit `{}`)

### DIFF
--- a/pkg/github/context_tools_test.go
+++ b/pkg/github/context_tools_test.go
@@ -139,6 +139,47 @@ func Test_GetMe(t *testing.T) {
 	}
 }
 
+func Test_GetMe_OmittedArguments(t *testing.T) {
+	t.Parallel()
+
+	serverTool := GetMe(translations.NullTranslationHelper)
+
+	mockUser := &github.User{
+		Login:     github.Ptr("testuser"),
+		Name:      github.Ptr("Test User"),
+		HTMLURL:   github.Ptr("https://github.com/testuser"),
+		CreatedAt: &github.Timestamp{Time: time.Now()},
+	}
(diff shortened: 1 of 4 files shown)


King-la141 (commented):
King

advancedresearcharray:
Closing as duplicate of #2632, which includes the same fix plus integration coverage for `get_me` with omitted arguments.

advancedresearcharray:
Closing in favor of #2632, which includes the same fix plus integration coverage for get_me with omitted arguments.

PR #2612, opened 2026-06-05

### LINKED ISSUE #2519: MCP App UI resources for write tools are registered in --read-only mode

he-yufeng:
## Summary
- hide the issue/pr write UI resources when the server runs in read-only mode
- keep the read-only-safe get_me UI resource registered
- add a regression test for the resource list returned by RegisterUIResources in read-only mode

Fixes #2519.

## To verify
- go test ./pkg/github -run "Test(RegisterUIResources|NewMCPServer_RegistersUIResources)" -count=1
- git diff --check origin/main...HEAD


advancedresearcharray:
Rebased this branch onto current `main` and validated locally.

**Rebased commit:** `b5e2b7b` (based on latest `github/github-mcp-server` `main`)

**Validation (Go 1.25):**
- `go test ./pkg/github -run "Test(RegisterUIResources|NewMCPServer_RegistersUIResources)" -count=1` — pass (`TestRegisterUIResources_ReadOnlySkipsWriteResources`; UI-asset tests skip without `script/build-ui`)
- `go test ./... -count=1` — pass
- `git diff --check upstream/main...HEAD` — clean

**To update this PR branch:**
```bash
git fetch https://github.com/advancedresearcharray/github-mcp-server.git fix/read-only-ui-resources-v2
git checkout fix/read-only-ui-resources-v2
git reset --hard FETCH_HEAD
git push --force-with-lease origin fix/read-only-ui-resources-v2
```

This should trigger CI on the PR. The branch is currently mergeable with `main`; only maintainer review is blocking.

Note: #2581 is an older duplicate of the same fix and is behind `main` — consider closing it in favor of this PR.

SamMorrowDrums (approved):
Thank you for this! Much appreciated. 🙏

PR #2632, opened 2026-06-06
fix: accept omitted tool arguments for zero-param tools like get_me

### LINKED ISSUE #2587: bug: `get_me` tool fails without parameters due to empty JSON object `{}` not being accepted as valid input

advancedresearcharray:
## Summary
- Treat nil or empty `arguments` on `tools/call` as `{}` before JSON unmarshaling in `NewServerToolWithContextHandler`
- Fixes `get_me` and other zero-parameter tools when MCP clients omit the `arguments` field entirely instead of sending `{}`

Closes #2587

## Problem
Some MCP clients omit the `arguments` field (or send an empty byte slice) when calling tools that declare no required parameters. The server previously attempted to unmarshal nil/empty input directly, producing `invalid arguments: unexpected end of JSON input` before the handler ran.

`get_me` is the most visible affected tool because its schema declares no required arguments, yet strict clients that omit `arguments` could not call it successfully.

## Changes
- `pkg/inventory/server_tool.go`: normalize empty `arguments` to `{}` before unmarshaling
- `pkg/inventory/server_tool_test.go`: cover nil, empty byte slice, and explicit `{}` inputs
- `pkg/github/context_tools_test.go`: integration test for `get_me` with omitted arguments
- `pkg/github/helper_test.go`: add `createMCPRequestWithRawArguments` test helper

## Test plan
- [x] `go test ./pkg/inventory/... -run EmptyArguments`
- [x] `go test ./pkg/github/... -run 'GetMe|OmittedArguments'`

### DIFF
--- a/pkg/github/context_tools_test.go
+++ b/pkg/github/context_tools_test.go
@@ -139,6 +139,48 @@ func Test_GetMe(t *testing.T) {
 	}
 }
 
+func Test_GetMe_OmittedArguments(t *testing.T) {
+	t.Parallel()
+
+	serverTool := GetMe(translations.NullTranslationHelper)
+
+	mockUser := &github.User{
+		Login:     github.Ptr("testuser"),
+		Name:      github.Ptr("Test User"),
+		HTMLURL:   github.Ptr("https://github.com/testuser"),
+		CreatedAt: &github.Timestamp{Time: time.Now()},
+	}
+	mockedClient := MockHTTPClientWithHandlers(map[string]http.HandlerFunc{
+		GetUser: mockResponse(t, http.StatusOK, mockUser),
+	})
+	deps := BaseDeps{Client: mustNewGHClient(t, mockedClient), Obsv: stubExporters()}
+	handler := serverTool.Handler(deps)
+
+	tests := []struct {
+		name      string
+		arguments json.RawMessage
+	}{
+		{name: "nil arguments", arguments: nil},
+		{name: "empty byte slice", arguments: json.RawMessage{}},
+		{name: "explicit empty object", arguments: json.RawMessage(`{}`)},
+	}
+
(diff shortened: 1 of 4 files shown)


advancedresearcharray:
Verified locally:

- `go test ./pkg/inventory/... -run EmptyArguments` — pass
- `go test ./pkg/github/... -run 'GetMe|OmittedArguments'` — pass

Branch is up to date with `main`. Closed duplicate #2607 in favor of this PR.

advancedresearcharray:
Re-verified locally with Go 1.25:

- `go test ./pkg/inventory/... -run EmptyArguments` — pass
- `go test ./pkg/github/... -run 'GetMe|OmittedArguments'` — pass

Branch is up to date with `main`. Ready for review.

advancedresearcharray:
**Merge-ready check-in**

This PR looks ready for maintainer review/merge when you have a moment.

- **Change:** fix: accept omitted tool arguments for zero-param tools like get_me
- **CI:** green / passing
- **Review:** REVIEW_REQUIRED (mergeable: MERGEABLE)

Happy to rebase or address any remaining feedback. Thanks for maintaining the project.

PR #2655, opened 2026-06-09
feat(http): support custom listen address

pyama86:
## Summary
- Add a `--listen-address` flag (env: `GITHUB_LISTEN_ADDRESS`) to the `http` command so the MCP HTTP server can bind to an explicit `host:port` instead of always listening on every interface (`:<port>`).
- The existing `--port` flag is preserved. When `--listen-address` is set it takes precedence; otherwise behavior is unchanged.
- Listen address resolution is factored into a small `resolveListenAddress` helper with unit tests covering empty, host:port, `:port`, and IPv6 forms.

## Motivation
Today the HTTP server hard-codes `addr := fmt.Sprintf(":%d", cfg.Port)`, which binds to `0.0.0.0` on every interface. When this server runs inside Kubernetes (or any shared network), the Pod IP is reachable from anywhere that can route to the cluster network: if an attacker (or just another workload) discovers the Pod IP, they can hit the MCP HTTP server directly and bypass any Service / Ingress / NetworkPolicy that was supposed to gate it. This is especially risky because the MCP server proxies authenticated GitHub API calls.

Allowing operators to bind to `127.0.0.1:8082` (sidecar / loopback-only consumers) or to a specific interface address closes that gap and lets the standard "only the localhost / sidecar can talk to it" deployment pattern work without extra network plumbing.

## Test plan
- [x] `go build ./...`
- [x] `go test ./pkg/http/...` (new `TestResolveListenAddress` + existing tests pass)
- [ ] Manual: run `github-mcp-server http --listen-address 127.0.0.1:8082` and confirm the server only accepts connections on loopback
- [ ] Manual: run `github-mcp-server http --port 9090` (no `--listen-address`) and confirm existing `:9090` behavior is unchanged


SamMorrowDrums:
Hi @pyama86, would you mind splitting out the port. I think listen host should be independent. I am happy with the difference between 0.0.0.0 and 127.0.0.1 being trivially configurable, so I support the work, but I would prefer that it didn't override port and instead was independent.

pyama86:
Hi @SamMorrowDrums, thanks for the review\! Good call — I just pushed [605d6b6](https://github.com/github/github-mcp-server/pull/2655/commits/605d6b6) which splits this into an independent host flag instead of overriding the port.

- Removed `--listen-address` (host:port).
- Added `--listen-host` (env: `GITHUB_LISTEN_HOST`) for the bind host only.
- `--port` is unchanged and always honored.
- When `--listen-host` is empty (default), behavior is identical to today: the server binds to all interfaces on `--port`.
- `resolveListenAddress` now uses `net.JoinHostPort` so IPv6 hosts (e.g. `::1`) are bracketed correctly.

Example: `github-mcp-server http --listen-host 127.0.0.1 --port 8082` binds loopback-only on 8082.

SamMorrowDrums (approved):
Thank you!

PR #2656, opened 2026-06-09
Add per-label is_suggestion and rationale to issue_write

iulia-b:
Extend the general issue_write tool so each label can be sent either as a plain name or as an object with optional rationale (<=280 chars) and is_suggestion. On update, labels with intent are sent in object form (mapping is_suggestion to suggest); the create path strips intent. Mirrors the granular update_issue_labels behavior.

## Summary

image

## Why

Fixes https://github.com/github/plan-track-agentic-toolkit/issues/349

## What changed

- `issue_write` tool now supports labels suggestions

## MCP impact

- [ ] No tool or API changes
- [ ] New tool added

- "let's try using the other mcp server available, suggest label duplicate with rationale "this is a suggestion from the local server""
Screenshot 2026-06-09 at 18 16 28
image

## Security / limits

## Tool renaming

## Lint & tests

## Docs

- [ ] Not needed

### DIFF
--- a/pkg/github/__toolsnaps__/issue_write.snap
+++ b/pkg/github/__toolsnaps__/issue_write.snap
@@ -36,7 +36,33 @@
       "labels": {
         "description": "Labels to apply to this issue",
         "items": {
-          "type": "string"
+          "oneOf": [
+            {
+              "description": "Label name",
+              "type": "string"
+            },
+            {
+              "properties": {
+                "is_suggestion": {
+                  "description": "If true, this label is sent to the API as a suggestion rather than an applied label. Whether the label is applied or recorded as a proposal is determined by the API. Only honored when updating an existing issue.",
+                  "type": "boolean"
+                },
+                "name": {
+                  "description": "Label name",
+                  "type": "string"
+                },
+                "rationale": {
+                  "description": "One concise sentence explaining what specifically about the issue led you to choose this label. State the concrete signal (e.g. 'Reports a crash when saving' → bug).",
+                  "maxLength": 280,
+                  "type": "string"
+                }
+              },
+              "required": [
+                "name"
+              ],
+              "type": "object"
+            }
+          ]
         },
         "type": "array"
       },
--- a/pkg/github/__toolsnaps__/issue_write.snap
+++ b/pkg/github/__toolsnaps__/issue_write.snap
@@ -43,8 +43,17 @@
             },
             {
               "properties": {
+                "confidence": {
+                  "description": "How confident you are in this choice. Use 'high' for clear signal or explicit user request, 'medium' for reasonable inference with some ambiguity, 'low' for best guess with limited signal.",
+                  "enum": [
+                    "low",
+                    "medium",
+                    "high"
+                  ],
+                  "type": "string"
+                },
                 "is_suggestion": {
-                  "description": "If true, this label is sent to the API as a suggestion rather than an applied label. Whether the label is applied or recorded as a proposal is determined by the API. Only honored when updating an existing issue.",
+                  "description": "If true, this value is sent to the API as a suggestion rather than an applied value. Whether it is applied or recorded as a proposal is determined by the API. Only honored when updating an existing issue.",
                   "type": "boolean"
                 },
                 "name": {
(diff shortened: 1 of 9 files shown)


pkg/github/issues.go
```diff
@@ -1858,7 +1858,30 @@ Options are:
...
+										"rationale": {
+											Type: "string",
+											Description: "One concise sentence explaining what specifically about the issue led you to choose this label. " +
+												"State the concrete signal (e.g. 'Reports a crash when saving' → bug).",
+											MaxLength: jsonschema.Ptr(280),
+										},
+										"is_suggestion": {
+											Type: "boolean",
+											Description: "If true, this label is sent to the API as a suggestion rather than an applied label. " +
+												"Whether the label is applied or recorded as a proposal is determined by the API. Only honored when updating an existing issue.",
```
alondahari:
```suggestion
											Description: "If true, this label is sent to the API as a suggestion rather than an applied label. "
```
Is this second line needed?

pkg/github/issues.go
```diff
@@ -2350,6 +2400,92 @@ type UpdateIssueOptions struct {
...
+	LabelsWithIntent []any
+}
+
+// parseIssueWriteLabels parses the issue_write "labels" parameter, which accepts
+// either plain label names (strings) or objects carrying intent metadata
+// (name, rationale, is_suggestion). It returns the plain label names (intent
+// stripped), the object-form payload to send when any label carries intent,
+// whether any label carried intent, whether the labels parameter was provided
+// at all, and an error.
+func parseIssueWriteLabels(args map[string]any) (names []string, payload []any, hasIntent bool, provided bool, err error) {
```
alondahari: Should we make this implementation more reusable by other mutations like issue fields or issue types? Thinking it might be easier for us to build on in the future.

alondahari (dismissed):
Looks good, just a comment about extensibility

pkg/github/issues.go
```diff
@@ -1858,7 +1858,30 @@ Options are:
...
+								{
+									Type: "object",
+									Properties: map[string]*jsonschema.Schema{
+										"name": {
+											Type:        "string",
+											Description: "Label name",
+										},
+										"rationale": {
+											Type: "string",
+											Description: "One concise sentence explaining what specifically about the issue led you to choose this label. " +
```
boazreicher: Since the limit is 280, why do we want to say that it's limited to one sentance?

boazreicher:
@iulia-b what about confidence arguments?

pkg/github/issues.go
iulia-b: Good idea, looking into it

pkg/github/__toolsnaps__/issue_write.snap
```diff
@@ -36,7 +36,42 @@
...
         "items": {
-          "type": "string"
+          "oneOf": [
+            {
+              "description": "Label name",
+              "type": "string"
+            },
+            {
+              "properties": {
+                "confidence": {
```
iulia-b, line 46: @boazreicher I added confidence with values low, medium and high, and without any optional param or anything special, just sending it to the server

pkg/github/issues.go
```diff
@@ -1858,7 +1858,19 @@ Options are:
 						Type:        "array",
 						Description: "Labels to apply to this issue",
 						Items: &jsonschema.Schema{
-							Type: "string",
+							OneOf: []*jsonschema.Schema{
+								{Type: "string", Description: "Label name"},
+								{
+									Type: "object",
+									Properties: withIntentProperties(map[string]*jsonschema.Schema{
```
iulia-b, line 1865: @boazreicher @alondahari , I extracted the parser and also the intent related props in their own type, and using it to extend labels (and soon, types and field values) ; also using this in the `issues_granular`

pkg/github/issues_granular.go
```diff
@@ -258,22 +258,187 @@ func GranularUpdateIssueAssignees(t translations.TranslationHelperFunc) inventor
...
+// the wire.
+type valueIntent struct {
 	Rationale  string `json:"rationale,omitempty"`
 	Confidence string `json:"confidence,omitempty"`
 	Suggest    bool   `json:"suggest,omitempty"`
 }
 
+// HasIntent reports whether any intent metadata was supplied.
+func (v valueIntent) HasIntent() bool {
+	return v.Rationale != "" || v.Confidence != "" || v.Suggest
```
iulia-b, line 279: @boazreicher  is this valid or should it only be `v.Suggest == true` for HasIntent?

boazreicher (approved):

PR #2659, opened 2026-06-10
Add multi_select issue field support

owenniblock:
## Summary

Adds multi-select issue field support across the consolidated and granular issue tools. Multi-select is **unconditional** — it matches the read paths, which already surface multi-select values for orgs that have the dotcom-side feature.

### What changed
- **issue_write** — `field_option_names` slot on `issue_fields[]` items + multi_select resolver
- **list_issues** — `field_filters[].values` slot for multi-select option filtering (AND semantics)
- **list_issue_fields** — description advertises `multi_select`
- **set_issue_fields** (granular) — already multi-select-aware; only gated by `FeatureFlagIssuesGranular`
- Read paths (`IssueFieldValueFragment`, `list_issues` enrichment, etc.) surface multi-select values for orgs with it enabled, matching the dotcom UI

### Feature-flag note
This originally gated the write/filter surface behind `remote_mcp_issue_fields_multiselect`. When merging `main` (which added the `fields_param` feature, itself splitting `list_issues` into flag-gated variants), the two flags collided on `list_issues` — the inventory's single `FeatureFlagEnable` can't express the "both flags on" case. Rather than carry two interacting flags on one tool, the multi-select flag was removed and multi-select made unconditional. `list_issues` now follows `main`'s `fields_param` split (`ListIssues` / `LegacyListIssues`) with the `values` slot always advertised.

`~~Stacked on #2755~~` — that PR (the delete-bug fix) is now merged into `main`.


pkg/github/issues.go
```diff
@@ -121,21 +122,25 @@ func getCloseStateReason(stateReason string) IssueClosedStateReason {
 type issueFieldWriteMetadataNode struct {
 	TypeName       githubv4.String `graphql:"__typename"`
 	IssueFieldText struct {
+		ID             githubv4.ID
```
akenneth: Why is this needed? this wasn't here before for other types right?

pkg/github/issues.go
```diff
@@ -281,25 +331,46 @@ func optionalIssueWriteFields(args map[string]any) ([]issueWriteFieldInput, erro
...
+			FieldOptionName:  fieldOptionName,
+			FieldOptionNames: fieldOptionNames,
 		})
 	}
 
 	return issueFields, nil
 }
 
-func resolveIssueRequestFieldValues(ctx context.Context, gqlClient *githubv4.Client, owner, repo string, issueFields []issueWriteFieldInput) ([]*github.IssueRequestFieldValue, []int64, error) {
+// fieldDeletion carries both identifiers needed to clear an issue field value:
```
akenneth: Are you sure this change is needed?

pkg/github/issues.go
```diff
@@ -386,6 +469,35 @@ func resolveIssueRequestFieldValues(ctx context.Context, gqlClient *githubv4.Cli
...
+				}
+				if matched == "" {
+					return nil, nil, fmt.Errorf("issue field option %q was not found for field %q", requested, fieldInput.FieldName)
+				}
+				resolvedNames = append(resolvedNames, matched)
+			}
+			// REST IssueField#build_value_attributes for multi_select expects an array of option names.
+			resolvedValue = resolvedNames
+		default:
+			// Raw value path. Reject it for multi_select so callers get a clear error
```
akenneth: nit: i think we can reduce the comments here, the `if` is clear enough

pkg/github/issues.go
```diff
@@ -437,6 +562,16 @@ func fetchExistingIssueFieldValues(ctx context.Context, gqlClient *githubv4.Clie
 		case "IssueFieldSingleSelectValue":
 			fieldIDStr = node.SingleSelectValue.Field.FullDatabaseIDStr()
 			value = string(node.SingleSelectValue.Value)
+		case "IssueFieldMultiSelectValue":
+			fieldIDStr = node.MultiSelectValue.Field.FullDatabaseIDStr()
+			// REST IssueField#build_value_attributes for multi_select expects an array
```
akenneth: nit: remove

pkg/github/issues.go
```diff
@@ -1667,7 +1805,8 @@ func fetchIssueFieldValuesByNodeID(ctx context.Context, gqlClient *githubv4.Clie
 	}
 
 	var q searchIssuesNodesQuery
-	if err := gqlClient.Query(ctx, &q, map[string]any{"ids": ids}); err != nil {
+	ctxWithFeatures := ghcontext.WithGraphQLFeatures(ctx, "issue_fields", "repo_issue_fields")
```
akenneth, line 2178: > Latent bug fix — fetchIssueFieldValuesByNodeID was missing the ghcontext.WithGraphQLFeatures("issue_fields", "repo_issue_fields") wrap, so without it the new fragment would have silently no-op'd on dotcom even with the right feature flags enabled.

Interesting! What was broken here? is it searching with fields?

pkg/github/issues.go
```diff
@@ -2433,6 +2587,32 @@ func UpdateIssue(ctx context.Context, client *github.Client, gqlClient *githubv4
 		return ghErrors.NewGitHubAPIStatusErrorResponse(ctx, "failed to update issue", resp, body), nil
 	}
 
+	// Clear any fields marked with delete:true via the GraphQL deleteIssueFieldValue
+	// mutation. The REST PATCH above can't do this reliably — Go's omitempty drops an
+	// empty issue_field_values array, leaving the old values intact.
+	if len(fieldsToDelete) > 0 {
```
akenneth: i am confused, wasn't this code already in place?
what were we doing with the previous `fieldIDsToDelete` ?

akenneth (commented):
nice work! left a few questions I am unsure about.

Also tagging @iulia-b and @kelsey-myers for extra 👀 since you are familiar with this part

pkg/github/issues.go
owenniblock: Yeah, it's new. The `delete: true` path now calls the `deleteIssueFieldValue` GraphQL mutation, which takes the field's node `id` as input — and delete can target any field type, so we grab `id` on every fragment. Before this PR there was no GraphQL delete, so we never needed the node id.

pkg/github/issues.go
owenniblock: I think so — we added `field_option_names` as a third value setter, so the old 2-way `value` vs `field_option_name` mutual-exclusion check had to become 3-way. Happy to simplify if you can see a cleaner way though.

pkg/github/issues.go
owenniblock: Not search/filtering — it's the read enrichment. `fetchIssueFieldValuesByNodeID` backs the `field_values` you get on `get_issue` and `search_issues` output. It was missing the `WithGraphQLFeatures("issue_fields", "repo_issue_fields")` wrap, so on dotcom the new IssueFieldValues fragment would just come back empty even with the flags on — field values silently wouldn't show up. Adding the wrap makes it actually resolve.

pkg/github/issues.go
owenniblock: Good question — it looked in place but it didn't actually work. Previously `resolveIssueRequestFieldValues` returned `[]int64 fieldIDsToDelete` and we leaned on the REST PATCH to clear them by sending an empty value — but Go's `omitempty` drops the empty `issue_field_values` array, so the clear silently no-op'd and the old values stuck around. This PR swaps that for an explicit `deleteIssueFieldValue` GraphQL mutation per field, so `fieldDeletion` now carries both the db id (to keep it out of the REST payload) and the node id (for the mutation). So `delete: true` actually clears now.

pkg/github/issues.go
owenniblock: Done 👍🏼

pkg/github/issues.go
owenniblock: Done 👍🏼

pkg/github/issues.go
```diff
@@ -2433,6 +2587,32 @@ func UpdateIssue(ctx context.Context, client *github.Client, gqlClient *githubv4
...
+			}
+			input := DeleteIssueFieldValueInput{
+				IssueID: issueID,
+				FieldID: deletion.NodeID,
+			}
+			if err := gqlClient.Mutate(ctx, &mutation, input, nil); err != nil {
+				return ghErrors.NewGitHubGraphQLErrorResponse(ctx, "failed to delete issue field value", err), nil
+			}
+		}
+	}
```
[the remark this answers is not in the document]
owenniblock: Good catch — fixed, wrapped it with `WithGraphQLFeatures("issue_fields", "repo_issue_fields")` like the sibling calls.

pkg/github/issues.go
```diff
@@ -264,15 +294,35 @@ func optionalIssueWriteFields(args map[string]any) ([]issueWriteFieldInput, erro
...
+					s, ok := item.(string)
+					if !ok {
+						return nil, fmt.Errorf("field_option_names for field %q must be an array of strings", fieldName)
+					}
+					fieldOptionNames = append(fieldOptionNames, s)
+				}
+			default:
+				return nil, fmt.Errorf("field_option_names for field %q must be an array of strings", fieldName)
+			}
+		}
```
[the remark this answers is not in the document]
owenniblock: Done — `field_option_names` and `multi_select_option_ids` both go through `parseStringSlice` now.

pkg/github/issues.go
```diff
@@ -3094,6 +3347,21 @@ func resolveFieldFilters(rawFilters []rawFieldFilter, fields []IssueField) ([]Is
...
+		if strings.EqualFold(o.Name, value) {
+			return o.Name, nil
+		}
+	}
+	optionNames := make([]string, 0, len(field.Options))
+	for _, o := range field.Options {
+		optionNames = append(optionNames, o.Name)
+	}
+	return "", fmt.Errorf("field_filters: %q is not a valid option for %q. Valid options: %s", value, field.Name, strings.Join(optionNames, ", "))
+}
```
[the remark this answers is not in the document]
owenniblock: Done — trimming both sides now, with a regression test.

tommaso-moro (dismissed):
Thank you @owenniblock ! Regarding the feature flag, I have no issue rolling this out unconditionally, but just so you know we may be in a position to remove the current FF (`fields_param`) in 1-2 weeks - not sure if this is relevant for you guys but wanted to share!

Also, are there any rollout dependencies outside this repository that must be completed, or is this ready to ship from your side?

pelshen-drdoctor:
This would be really helpful, currently agents using the MCP just don't believe multi select fields exist. Can this be merged yet?

PR #2672, opened 2026-06-11
Route github.localhost to split-host URLs for local dotcom dev

owenniblock:
## What

Adds a fourth branch to `parseAPIHost` in `pkg/utils/api.go` for hostnames matching `github.localhost` (or any subdomain of it). It mirrors the dotcom split-host layout — REST at `api.github.localhost`, GraphQL at `api.github.localhost/graphql`, uploads/raw on their own subdomains, oauth at the apex.

## Why

A local dotcom checkout serves its APIs from the same split-host shape as production. But `github.localhost` doesn't match the `github.com` or `*.ghe.com` branches in `parseAPIHost`, so today it falls through to `newGHESHost`, which tries `/api/v3` and `/api/graphql` against the apex and gets 404/422 back.

Hit this while testing the multi_select issue field PR (#2659) against local dotcom — without this patch the MCP server can't talk to the local stack at all without modifying `parseAPIHost` by hand.

## Notes

- Tests cover the new branch and the not-a-subdomain edge case (e.g. `notgithub.localhost` still routes through GHES).
- Only handles `github.localhost` — not bare `localhost:port` style dev setups. Happy to broaden if anyone is using something different.
- Draft because I'd like a sanity check from someone closer to local dev tooling before marking ready.

### DIFF
--- a/pkg/utils/api.go
+++ b/pkg/utils/api.go
@@ -239,9 +239,55 @@ func parseAPIHost(s string) (APIHost, error) {
 		return newDotcomHost()
 	}
 
+	if u.Hostname() == "github.localhost" || strings.HasSuffix(u.Hostname(), ".github.localhost") {
+		return newLocalDevHost(u.Scheme)
+	}
+
 	if u.Hostname() == "ghe.com" || strings.HasSuffix(u.Hostname(), ".ghe.com") {
 		return newGHECHost(s)
 	}
 
 	return newGHESHost(s)
+}
+
+// newLocalDevHost handles a local dotcom checkout served from github.localhost.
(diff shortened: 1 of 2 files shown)


msmahirkingkhan-tech:
"github": {
    "command": "docker",
    "args": [
    "run",
    "-i",
    "--rm",
    "-e",
    "GITHUB_PERSONAL_ACCESS_TOKEN",
    "-e",
    "GITHUB_HOST",
    "ghcr.io/github/github-mcp-server"
    ],
    "env": {
        "GITHUB_PERSONAL_ACCESS_TOKEN": "${input:github_token}",
        "GITHUB_HOST": "https://<your GHES or ghe.com domain name>"
    }
}

PR #2694, opened 2026-06-15
fix(repos): default create_repository to private when visibility omitted

SamMorrowDrums:
## Summary

`create_repository` previously created a **public** repository whenever the `private` parameter was omitted. Because the parameter is optional, agent-driven or automated workflows could unintentionally create public repos, exposing source code, configuration, workflow files, and commit history.

This change makes the secure choice the default: **omitting `private` now creates a private repository**. A public repository is only created when `private` is explicitly set to `false`.

## Changes

- `create_repository` now defaults `private` to `true` when the parameter is omitted (via `OptionalBoolParamWithDefault`).
- Updated the `private` field description and schema `default` so the behavior is explicit to callers.
- Updated the tool snapshot and regenerated docs (README).
- Tests:
  - Minimal-parameters case now asserts the repo is created private.
  - Added a case asserting `private: false` still creates a public repo.

## Behavior

| `private` value | Result |
| --- | --- |
| omitted | private (new default) |
| `true` | private |
| `false` | public |

## Verification

- `script/lint` — 0 issues
- `go test -race ./...` — all passing
- `script/generate-docs` — README updated

### DIFF
--- a/README.md
+++ b/README.md
@@ -1204,7 +1204,7 @@ The following sets of tools are available:
   - `description`: Repository description (string, optional)
   - `name`: Repository name (string, required)
   - `organization`: Organization to create the repository in (omit to create in your personal account) (string, optional)
-  - `private`: Whether repo should be private (boolean, optional)
+  - `private`: Whether the repository should be private. Defaults to true (private) when omitted. (boolean, optional)
 
 - **delete_file** - Delete file
   - **Required OAuth Scopes**: `repo`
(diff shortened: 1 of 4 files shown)


SamMorrowDrums:
This addresses the insecure default where `create_repository` produced a **public** repo when `private` was omitted.

Decision summary:
- Omitting `private` now defaults to a **private** repository (secure default / fail-closed).
- A public repo is only created when `private` is explicitly `false`.
- The `private` field description and JSON schema `default` now make the default explicit to callers.

| `private` | Result |
| --- | --- |
| omitted | private (new default) |
| `true` | private |
| `false` | public |

Verified with `script/lint` (0 issues), `go test -race ./...` (passing), and `script/generate-docs`. Tool snapshot and README updated.

JoannaaKL (approved):

PR #2728, opened 2026-06-18
fix: add DestructiveHint and OpenWorldHint to label_write annotations

### LINKED ISSUE #2723: `label_write` delete is missing `DestructiveHint: true` annotation
### Describe the bug

The `label_write` tool (`pkg/github/labels.go`) supports a `delete` method that deletes a label at the repository level via the `deleteLabel` GraphQL mutation. This permanently removes the label and all of its associations from every issue and PR in the repository.

However, the tool's annotations only set `ReadOnlyHint: false`. It does **not** set `DestructiveHint: true`. As a result, MCP clients receive no destructive-operation signal for what is arguably the most destructive label operation available, and may not surface an appropriate confirmation prompt before executing it.

### Expected behavior

`label_write` should be annotated with `DestructiveHint: true` (at minimum when `method` is `delete`), so MCP clients can warn the user before a repository-wide label deletion.

### Precedent in the codebase

The annotation discipline already exists elsewhere in the same package:
- `remove_sub_issue` correctly sets `DestructiveHint: jsonschema.Ptr(true)`
- The non-destructive issue label update (`update_issue_labels`) sets `DestructiveHint: jsonschema.Ptr(false)`

`label_write` delete simply appears to have been missed.

### Impact

An agent intending a minor change can trigger a repository-wide label deletion with no destructive-operation hint. Because deleting a label removes it from all issues/PRs, the associations cannot be automatically restored even if the label is recreated with the same name.

### Version

Verified against the open-source `github-mcp-server` v1.4.0 (latest release) and the hosted GitHub Copilot MCP endpoint at `api.githubcopilot.com/mcp/` (as of 2026-06-18). The missing annotation has been present since `label_write` was introduced (ref #312).

axelray-dev:
Fixes #2723

## Problem

The `label_write` tool supports a `delete` method that permanently removes a label and all its associations from every issue/PR in the repository. However, its annotations only set `ReadOnlyHint: false` -- it does not set `DestructiveHint: true` or `OpenWorldHint: true`. MCP clients receive no destructive-operation signal for what is arguably the most destructive label operation available.

## Fix

Add `DestructiveHint: jsonschema.Ptr(true)` and `OpenWorldHint: jsonschema.Ptr(true)` to the `LabelWrite` tool annotations, matching the pattern used by other destructive tools in the codebase:

- `actions_run_trigger` (actions.go)
- `projects_write` (projects.go)
- `discussion_comment_write` (discussions.go)
- `remove_sub_issue` (issues_granular.go)

## Changes

- `pkg/github/labels.go`: Add annotation fields to LabelWrite tool definition
- `pkg/github/labels_test.go`: Add DestructiveHint assertion in TestWriteLabel
- `pkg/github/__toolsnaps__/label_write.snap`: Update snapshot to match

### DIFF
--- a/pkg/github/__toolsnaps__/label_write.snap
+++ b/pkg/github/__toolsnaps__/label_write.snap
@@ -1,6 +1,8 @@
 {
   "annotations": {
+    "destructiveHint": true,
     "idempotentHint": false,
+    "openWorldHint": true,
     "readOnlyHint": false,
     "title": "Write operations on repository labels"
   },
--- a/pkg/github/labels.go
+++ b/pkg/github/labels.go
@@ -226,8 +226,10 @@ func LabelWrite(t translations.TranslationHelperFunc) inventory.ServerTool {
 			Name:        "label_write",
 			Description: t("TOOL_LABEL_WRITE_DESCRIPTION", "Perform write operations on repository labels. To set labels on issues, use the 'update_issue' tool."),
 			Annotations: &mcp.ToolAnnotations{
-				Title:        t("TOOL_LABEL_WRITE_TITLE", "Write operations on repository labels"),
-				ReadOnlyHint: false,
+				Title:           t("TOOL_LABEL_WRITE_TITLE", "Write operations on repository labels"),
+				ReadOnlyHint:    false,
+				DestructiveHint: jsonschema.Ptr(true),
+				OpenWorldHint:   jsonschema.Ptr(true),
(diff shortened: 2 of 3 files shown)


axelray-dev:
Closing this as superseded.

#2723 was fixed on main by #2763 (`DestructiveHint` on `label_write`). Keeping this open would only re-litigate an already-merged annotation change.

PR #2755, opened 2026-06-23
Fix delete:true on issue fields by calling deleteIssueFieldValue mutation

owenniblock:
## Summary

REST PATCH alone can't clear an issue field value: `go-github`'s `IssueRequest.IssueFieldValues` carries `json:"issue_field_values,omitempty"`, so Go's encoder strips an empty `[]*IssueRequestFieldValue` from the body. The current merge-and-PATCH flow filters the deleted entry out of the kept list and then assigns it back to the request; when the kept list ends up empty (e.g. you delete the only remaining field value), the entire `issue_field_values` key vanishes from the wire format and the server treats it as a no-op. The field's old value sticks around.

Capture each field's GraphQL node ID alongside its database ID at resolve time, and after the REST PATCH succeeds run `deleteIssueFieldValue` per deletion. The mutation is idempotent, per-field, and not subject to the omitempty silent-no-op. It also sidesteps the race window in the old fetch-merge-PATCH approach (where a concurrent edit between our fetch and our PATCH would be silently clobbered by REST's set semantics).

## Empirical confirmation of the bug

Running an equivalent flow on `main` (mocking `fetchExistingIssueFieldValues` to return one field, calling `UpdateIssue` asking to delete that field's value, capturing what reaches the PATCH endpoint) produces a literal PATCH body of `{}`. The server's update handler in `github/github`'s `app/api/issues.rb` guards the entire field-values block on `data.include?(ISSUE_FIELD_VALUES)` — key absent → nothing touched. Confirmed bug.

## Tests

Three new tests in `pkg/github/issues_delete_test.go`:

- **`Test_IssueRequest_EmptyFieldValues_OmittedByJSON`** — pins the omitempty contract so we know if `go-github` ever drops the tag (at which point this fix could be revisited).
- **`Test_UpdateIssue_DeleteFieldValueRunsGraphQLMutation`** — deletes the only existing field value, captures the REST PATCH body, asserts it has no `issue_field_values` key, and asserts (via mutation matcher) that the GraphQL clear fires.
- **`Test_UpdateIssue_DeleteAndSetFieldsInSameCall`** — combined set-one-delete-another in a single call. REST PATCH carries only the set; GraphQL mutation fires for the deletion.

## Stack

This PR is the universal bug-fix half of #2659. The multi-select feature work (FF-gated) will be rebased on top of this and re-opened as a stacked PR once this lands.

### DIFF
--- a/pkg/github/issues.go
+++ b/pkg/github/issues.go
@@ -2179,6 +2179,17 @@ func UpdateIssue(ctx context.Context, client *github.Client, gqlClient *githubv4
 		issueRequest.Type = github.Ptr(issueType)
 	}
 
+	// fallbackDeleteFieldIDs holds field IDs we need to clear via the dedicated
+	// REST DELETE endpoint after the PATCH lands. This is the omitempty-trap
+	// fallback: when the merged list is empty AND we have deletions to make,
+	// `go-github`'s `omitempty` tag on IssueRequest.IssueFieldValues strips the
+	// empty slice from the JSON body, the dotcom REST handler's top-level
+	// `if data.include?(ISSUE_FIELD_VALUES)` guard short-circuits, and nothing
+	// gets cleared. When the merged list is non-empty the PATCH's set semantics
+	// handle the deletes implicitly (current - new), so no DELETE follow-up is
+	// needed in that path.
+	var fallbackDeleteFieldIDs []int64
+
 	if len(issueFieldValues) > 0 || len(fieldIDsToDelete) > 0 {
 		// The REST update endpoint uses "set" semantics — it overwrites all existing
 		// field values with whatever is sent. Fetch the current values first, merge in
@@ -2201,7 +2212,15 @@ func UpdateIssue(ctx context.Context, client *github.Client, gqlClient *githubv4
 			}
 			merged = kept
 		}
-		issueRequest.IssueFieldValues = merged
+		if len(merged) == 0 && len(fieldIDsToDelete) > 0 {
+			// Omitempty trap: skip the IssueFieldValues assignment so we don't
+			// rely on a value that's about to be stripped from the JSON anyway,
+			// and clear each field via the dedicated DELETE endpoint after the
+			// PATCH lands.
+			fallbackDeleteFieldIDs = fieldIDsToDelete
+		} else {
+			issueRequest.IssueFieldValues = merged
+		}
 	}
 
 	updatedIssue, resp, err := client.Issues.Edit(ctx, owner, repo, issueNumber, issueRequest)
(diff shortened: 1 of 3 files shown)


pkg/github/issues.go
```diff
@@ -363,7 +380,7 @@ func resolveIssueRequestFieldValues(ctx context.Context, gqlClient *githubv4.Cli
 		}
 
 		if fieldInput.Delete {
```
iulia-b, line 365: We do have a delete type for IssueWrite request (see [line46](https://github.com/github/github-mcp-server/blob/8c16c3b9f5d6feb37cddfc291ba645b5004b56c7/pkg/github/issues.go#L46)) it's how a field can be marked for deletion , and it is handled in th emerge and path. The idea of this tool is to call the updateIssue endpoint only, and do all the mutations for an issue in one operation.

iulia-b (commented):
Unless I misunderstand the intention, I wouldn't opt for changing the way we delete fields in the updateIssue MCP tool

pkg/github/issues.go
owenniblock: The problem (as far as I can see it) is that when the last issue field is deleted, we end up with: `issue_field_values: []` but because we're using `omitempty` - the API request doesn't have `issue_field_values` and so the delete doesn't happen.

Another approach could be to update `go_github` so we're not using `omitempty` I think. As I say - I'm not yet great at Go so I could be wrong here 😆

SamMorrowDrums:
@iulia-b as you are on it please let me know if you validate/a fix is required :-D

pkg/github/issues.go
owenniblock: More details from some additional testing I did to double-check:

Verified just now against this repo on [#2756](https://github.com/github/github-mcp-server/issues/2756):

| Action | Tool response | Server state |
|---|---|---|
| Set `Priority = P1` | — | `P1` |
| `main` binary: `issue_write` `delete: true` | success | still `P1` — silent no-op |
| PR branch: same call | success | `[]` — cleared |

The dangerous bit is the identical success response — the model would happily report "cleared" when nothing changed.

iulia-b: >  to update go_github so we're not using omitempty I think.

I can't remember why but that was not an option when i went for this approach

>  when the last issue field is deleted, we end up with: issue_field_values: []

Not sure I follow. The current [schema](https://github.com/github/github-mcp-server/pull/2551/files#diff-77ac78a9743e78b3d9f560b7aee4382b5a55b7fc5dd804e25557ddc433a47bf5R32-R33) states, for issue_fields prop:

```
 "issue_fields": {
        "description": "Issue field values to set or clear. Each item requires 'field_name' and exactly one of 'value', 'field_option_name', or 'delete: true'.", ...
```

so the call should look smt like: `update_issue( issue_fields : [ "priority": { field_option_name: "P1"} , "impact": { delete: true }] )`

kelsey-myers (dismissed):
nit: could we trim the verbose comments 👀 other than that I was able to repro, fix makes sense to me

SamMorrowDrums (approved):

gubonlucid-com (approved):

PR #2760, opened 2026-06-24
MCP: name-based resolution for Projects fields

veralizeth:
## Summary

Let agents address Project v2 fields, options, and item values by name (read +write) instead of pre-resolved IDs. Resolution happens in the wrapper; the mutation stays ID-typed.

## Why

Today an agent that wants to update a project field has to assemble several opaque IDs (project, item, field, single-select option) before it can write. The MCP `update_project_item` tool is only partly there, it accepts owner login, project number, and issue number, but the write still demands the field ID and item ID and does no option-name resolution.

`update_project_item` demands field ID + item ID and does no option-name resolution. Reads are ID-bound too: `list_project_items` returns only Title, so reading a field value needs a `list_project_fields`. This forces a list-and-filter dance, burns context, and invites mis-bound IDs.

The important part: this is a surfacing gap, not a platform gap. The GraphQL API already resolves names `projectV2.field(name:)`, `ProjectV2SingleSelectField.options(names:)`, `ProjectV2Item.fieldValueByName(name:)`, plus top-level lookups by login / owner+name / project number. A single nested read can fetch every ID the write needs. The agent surface simply doesn't pass those resolvers through.

Closes [#3430](https://github.com/github/planning-tracking/issues/3430)

## What changed

- **Write**: `update_project_item` accepts a field via `updated_field.name`, resolves the single-select option by name, and resolves the item by `item_owner` + `item_repo`  + `issue_number` when `item_id` is omitted. `{id}`/`{name}` mutually exclusive.
- **Read**: new optional `field_names` on `get_project_item` / `list_project_items`  no `list_project_fields` preflight.
- **Errors**: name failures return typed, self-correctable JSON (`field_not_found`, `field_ambiguous`, `option_not_found`, `item_not_in_project`, …) with candidates.  Backward compatible, existing ID paths unchanged; new inputs additive/optional.

### Follow-up

- **Pagination**: item resolution reads `projectItems(first: 50)`, bound and paginate Project item read payloads so an agent can pull field values for a large board without overflowing its buffer or burning context. [#3332](https://github.com/github/planning-tracking/issues/3332).

## MCP impact

- [ ] No tool or API changes
- [ ] New tool added

## Security / limits

**Schema token delta, measured with `o200k_base`:**

| Tool | Before | After | Δ |
| --- | ---: | ---: | ---: |
| `projects_get` | 307 | 397 | +90 |
| `projects_list` | 359 | 449 | +90 |
| `projects_write` | 845 | 958 | +113 |
| **TOTAL** | **1,511** | **1,804** | **+293** |

### How this was measured

Used `mcpcurl schema` to dump the `tools/list` payload (`name` + `description` +`inputSchema`) on `main` vs. this branch, tokenized both with `tiktoken` (`o200k_base` encoding), and diffed.

## Prompts tested

- Set Status = "In Progress" on issue # 1 in project "veralizeth-test-p" by name.
- Read Status/Priority via `field_names` no field-ID preflight.
- Two "Status" fields → `field_ambiguous` with candidates.
- `field_names: ["Statuss"]` → `field_not_found` with candidates.

## Tool renaming

## Lint & tests

## Docs

- [ ] Not needed


pkg/github/__toolsnaps__/projects_get.snap
```diff
@@ -10,8 +10,15 @@
...
+      "field_names": {
+        "description": "Specific list of field names to include in the response when getting a project item (e.g. [\"Status\", \"Priority\"]). Resolved server-side to field IDs — pass this instead of 'fields' when you only know the human-readable names. Only used for 'get_project_item' method.",
+        "items": {
+          "type": "string"
+        },
+        "type": "array"
+      },
       "fields": {
-        "description": "Specific list of field IDs to include in the response when getting a project item (e.g. [\"102589\", \"985201\", \"169875\"]). If not provided, only the title field is included. Only used for 'get_project_item' method.",
+        "description": "Specific list of field IDs to include in the response when getting a project item (e.g. [\"102589\", \"985201\", \"169875\"]). If neither 'fields' nor 'field_names' is provided, only the title field is included. Only used for 'get_project_item' method.",
```
zwick: `fields` and `field_names` both resolve to field IDs, and the "use one or the other" rule only lives in the description text. This same pair shows up four times in the PR — the `get_project_item` and `list_project_items` schema blocks in projects.go, plus both .snap fixtures — so whatever we decide needs to apply the same way in all of them (README too).

Right now nothing stops a caller from sending both, and if they do, the handler quietly appends the resolved names onto `fields` (`fields = append(fields, resolvedIDs...)`) — which none of the four descriptions mention.

Two ways to go:

1. Put the constraint in the schema. It sits next to `properties`, at the `inputSchema` level (not inside a property), roughly:

   ```json
   "inputSchema": {
     "properties": {
       "fields":      { "type": "array", "items": { "type": "string" }, "description": "..." },
       "field_names": { "type": "array", "items": { "type": "string" }, "description": "..." }
     },
     "oneOf": [
       { "not": { "required": ["fields", "field_names"] } },
       { "required": ["fields"],      "not": { "required": ["field_names"] } },
       { "required": ["field_names"], "not": { "required": ["fields"] } }
     ],
     "required": ["method"],
     "type": "object"
   }
   ```

   Two catches, and they hit every copy: neither field is required, so you need that first "neither" branch just to keep the common case legal; and these props are shared across `method`s on each tool (`get_project_field`/`get_project_item`, and the list methods), so a top-level `oneOf` would wrongly constrain the others. Doing it right means gating per method (`if method == get_project_item ...`) in all four spots — a lot of schema for what it buys.

2. Leave it in prose but enforce it once in the handler — reject when both are set with a structured error, exactly like `updated_field` already does for `{id}`+`{name}`. One check covers both read methods, and we just make the descriptions symmetric (`field_names` currently says nothing about the both-set case).

I'd go with 2 — the write path already draws this line in code, so matching it on the read path keeps things consistent, avoids repeating awkward schema in four places, and the fix lives in one spot instead of four. Open to 1 if we think the schema-level guarantee is worth the per-method branching.

pkg/github/projects_resolver.go
```diff
@@ -0,0 +1,386 @@
...
+		ProjectV2SingleSelectField struct {
+			ID         githubv4.ID
+			DatabaseID githubv4.Int `graphql:"databaseId"`
+			Name       githubv4.String
+			DataType   githubv4.String
+			Options    []struct {
+				ID   githubv4.String
+				Name githubv4.String
+			}
+		} `graphql:"... on ProjectV2SingleSelectField"`
```
zwick, line 75: We may want to follow-up on this PR with some extra field types. 

[Issue Fields](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/adding-and-managing-issue-fields) and [Issue Field Values](https://docs.github.com/en/graphql/reference/projects#union-projectv2issuefieldvalues) attached to a Project Item have their own Node type that resolves slightly differently than standard project fields.
- One of the Memex teams has been working on MultiSelect fields which I believe recently started shipping to customers. It would be cool to get support for those in place as well (if they are ready, we should confirm first)

These are not blockers, I think this PR is already covering a lot of ground but would be great follow-ups to expand our coverage for different enum types

zwick:
**Heads-up on name-matching case sensitivity — I think we should align both surfaces to the platform.**

While building the CLI sibling of this (name-based resolution for `gh project item-edit`/`item-list`, cli/cli#13807), I dug into how the Projects v2 platform actually resolves names, and it changed my mind on matching behavior.

**The platform API is case-insensitive on all three name resolvers:**
- `ProjectV2.field(name:)` — field names are uniqueness-constrained on a normalized `name_slug` (`name.downcase.gsub(" ","-")` in `memex_project_column.rb`), so "Status" and "status" can't coexist (there's a test asserting the second create returns 422). Effectively case-insensitive.
- `ProjectV2SingleSelectField.options(names:)` — explicitly downcases both sides (`names.map(&:downcase)` in `settings_option_ids`); the GraphQL arg is even documented as "case insensitive."
- `ProjectV2Item.fieldValueByName(name:)` — uses the same case-insensitive field loader.

Our resolver here matches names with exact Go `==` (`resolveProjectFieldByName`: `f.Name == fieldName`; `resolveSingleSelectOptionByName`: `o.Name == optionName`), which is **stricter than the platform** — e.g. `name: "status"` fails where the API would resolve it. The CLI currently has the same client-side `==` behavior, so both agent surfaces diverge from the platform in the same direction.

I'm switching the CLI to case-insensitive (`strings.EqualFold`) to match the platform. Proposing we do the same here so MCP and CLI stay consistent *and* both match the API that ultimately does the write.

**One nuance worth preserving:** for **fields**, case-insensitive matching adds no new ambiguity — the platform guarantees field names are case-insensitively unique, so a fold-match can only ever hit one field. For **options**, names aren't guaranteed case-insensitively unique, so an `EqualFold` match could return >1 — keep the `option_ambiguous` path firing in that case.

Not a blocker for this PR, just flagging so we don't ship two surfaces that are each subtly out of step with the platform. Happy to pair on it.

pkg/github/__toolsnaps__/projects_get.snap
veralizeth: 🙇🏽‍♀️ Following # 2 -> Both now reject with `"provide either 'fields' or 'field_names', not both"` in `get_project_item` and `list_project_items` it runs right after both `fields` and `field_names` are parsed.

pkg/github/projects_resolver.go
veralizeth: Created a follow up issue > https://github.com/github/planning-tracking/issues/3523

veralizeth:
> Heads-up on name-matching case sensitivity — I think we should align both surfaces to the platform.

@zwick Great catch 🎣  I addressed these here to avoid fails for matching, so matched EqualFold + kept option_ambiguouss.

zwick (approved):
Thanks for the updates! Projects tools make sense to me!

tommaso-moro (approved):
lgtm!

gubonlucid-com (approved):

gubonlucid-com (approved):

PR #2769, opened 2026-06-25
fix(tools): list supported methods in unknown-method dispatch errors

### LINKED ISSUE #2712: Consolidated `method`-dispatch tools return unhelpful "unknown method" errors (don't list supported methods)
### Describe the bug

The consolidated tools that dispatch on a method argument return a bare unknown method: <value> when the method is unrecognized, without listing the valid methods. `labels.go` already does the helpful thing ("... Supported methods are: create, update, delete"); the other method-dispatch tools don't, so the behavior is inconsistent.

Separately, `pull_request_review_write` decodes arguments with `mapstructure.WeakDecode` (it does not use RequiredParam like the other tools), so omitting the required method produces "unknown method: "  with an empty value — instead of telling the caller that method is required.

Affected tools: `pull_request_read`, `pull_request_review_write`, `issue_read`, sub-issue write, actions (list/get/run), projects (list/get/write), ui_get.

### Affected version

Version: v1.3.0
Commit: 34227037fc48771baea9af7163e28cb6556ef287
Build Date: 2026-06-11T14:16:55Z

1. Call `pull_request_review_write` with the required method omitted, e.g. `{ "owner": "o", "repo": "r", "pullNumber": 42 }`.
2. View the output.
3. Server returns (tool-result error, HTTP 200 `isError`): 

`unknown method: `
(empty after the colon — method was never supplied, but the error doesn't say it's required or what the valid values are).

Or, with an unrecognized value `{ "method": "bogus", ... }`:

`unknown method: bogus`
No list of supported methods. The same bare error occurs on pull_request_read, issue_read, actions_*, projects_*, and ui_get for any unrecognized method.

Expected: the error names the supported methods (matching labels.go, e.g. unknown method: bogus. Supported methods are: create, submit_pending, delete_pending, resolve_thread, unresolve_thread), and an omitted required method reports method is required. Supported methods are: ....

Actual: a bare unknown method: <value> with no guidance; for pull_request_review_write an omitted method yields unknown method:  (empty), which reads like a routing bug rather than a missing required argument.

Logs
Tool-result error text returned by the server:

unknown method:
unknown method: bogus

syf2211:
## Summary

Improve consolidated method-dispatch MCP tools to return actionable errors when an unrecognized `method` is provided, matching the existing `label_write` behavior.

## Motivation

Fixes #2712. Tools such as `pull_request_read`, `pull_request_review_write`, `issue_read`, `sub_issue_write`, `actions_*`, `projects_*`, and `ui_get` previously returned bare `unknown method: <value>` errors. When `method` was omitted from `pull_request_review_write`, callers saw `unknown method: ` with an empty value instead of a required-parameter error.

## Changes

- Add shared `methodEnum()` and `unknownMethodError()` helpers in `pkg/github/params.go`
- Declare per-tool method slices as the single source for schema enums and runtime validation
- Update consolidated tools to use the helpers in their `default` branches
- Validate required `method` in `pull_request_review_write` before `WeakDecode` routing
- Add regression tests for helpers and `pull_request_review_write` error paths
- Update `sub_issue_write` toolsnap enum

## Tests

```bash
go test ./pkg/github/... -count=1
```

Result: **PASS** (0.161s)

## Notes

- Rebased on current `main`; `issue_read` retains `get_parent` support.
- `label_write` already had inline supported-method messaging and was left unchanged to keep scope minimal.
- Related open PR #2713 appears conflicted/stale; this branch is a fresh implementation on current `main`.

### DIFF
--- a/pkg/github/__toolsnaps__/sub_issue_write.snap
+++ b/pkg/github/__toolsnaps__/sub_issue_write.snap
@@ -21,6 +21,11 @@
       },
       "method": {
         "description": "The action to perform on a single sub-issue\nOptions are:\n- 'add' - add a sub-issue to a parent issue in a GitHub repository.\n- 'remove' - remove a sub-issue from a parent issue in a GitHub repository.\n- 'reprioritize' - change the order of sub-issues within a parent issue in a GitHub repository. Use either 'after_id' or 'before_id' to specify the new position.\nWrites issue hierarchy. To move a sub-issue to a new parent, use `add` with `replace_parent=true`; there is no writable parent field.\n",
+        "enum": [
+          "add",
+          "remove",
+          "reprioritize"
+        ],
         "type": "string"
       },
       "owner": {
--- a/pkg/github/actions.go
+++ b/pkg/github/actions.go
@@ -46,6 +46,15 @@ const (
 	actionsMethodDeleteWorkflowRunLogs    = "delete_workflow_run_logs"
 )
 
+// Method sets for the consolidated actions tools. Each slice is the single
+// source for its tool: it feeds both the schema enum (methodEnum) and the
+// unknown-method error (unknownMethodError). Order matches the advertised enum.
+var (
+	actionsWorkflowListMethods = []string{actionsMethodListWorkflows, actionsMethodListWorkflowRuns, actionsMethodListWorkflowJobs, actionsMethodListWorkflowArtifacts}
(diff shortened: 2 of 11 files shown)


syf2211:
Rebased onto latest `main` and resolved merge conflicts in `issues.go`, `params.go`, and `sub_issue_write.snap` (kept upstream hierarchy description text while preserving `methodEnum` + `unknownMethodError`). `go test ./pkg/github/` passes locally.

PR #2775, opened 2026-06-26
Add fields param to search_code and get_file_contents

tommaso-moro:
## What

Adds an optional `fields` array parameter to `search_code` and `get_file_contents` so callers can request only the fields they need, reducing tool response size and context usage. The feature is gated behind the `fields_param` feature flag and instrumented so we can measure adoption and realized savings.

## How

- **`fields` filtering** — `search_code` filters each result item (selectable: `name`, `path`, `sha`, `repository`, `text_matches`); `get_file_contents` filters each directory entry (selectable: `type`, `name`, `path`, `size`, `sha`, `url`, `git_url`, `html_url`, `download_url`). Filtering is ignored for single-file responses. When `fields` is omitted, behavior is unchanged.
- **Feature flag** — each tool registers two mutually exclusive variants following the existing dual-variant flag pattern: a flag-enabled variant that advertises `fields` and filters, and a legacy variant with the original schema that acts as a kill switch when the flag is off. When the flag is off, `fields` is not advertised at all.
- **Telemetry** — best-effort, low-cardinality metrics emitted at each tool's filter point: `mcp.fields.tool_call` (adoption) and `mcp.fields.bytes_full` / `bytes_sent` (effectiveness). Tags are limited to `tool` and `filtered` so cardinality stays bounded.

Filtering is done by a small generic helper (`filterFields` / `filterEachField` in `minimal_types.go`) that marshals each result item to JSON, decodes it to a `map[string]any`, and keeps only the requested keys. I opted for this so field names stay in sync with each type's json tags and one helper works for every tool.

## Testing

- Unit tests for field filtering, flag gating (both variants), and telemetry emission (adoption + realized savings).
- Toolsnaps updated: the flag-enabled variants own the `_ff_fields_param` snapshots; the legacy variants own the canonical snapshots. README and `docs/feature-flags.md` regenerated.
- `script/lint` clean; `script/test` (full race suite) passing.

## Example Tool Call
Screenshot 2026-07-08 at 09 53 22


kerobbi (approved):
🚀

PR #2751, opened 2026-06-22
Add issue dependency read/write MCP tools

zwick:
## Summary
Add two feature-flagged MCP tools for issue dependencies: `issue_dependency_read` (list an issue's blocked-by / blocking relationships) and `issue_dependency_write` (add / remove a blocked-by or blocking relationship).

## Why
`github-mcp-server` had no way to read or edit issue **dependencies**, although the GitHub API supports them and the `gh` CLI shipped the verbs in v2.94.0. This closes that MCP gap.

Closes #2391
Refs #950

## What changed
- `issue_dependency_read` — `get_blocked_by` / `get_blocking` via the GraphQL `Issue.blockedBy` / `Issue.blocking` connections, cursor-paginated.
- `issue_dependency_write` — `add` / `remove` × `blocked_by` / `blocking` via the `addBlockedBy` / `removeBlockedBy` mutations. Accepts issue numbers (resolved to node IDs in one aliased query); `blocking` is the inverse of `blocked_by` with subject/related swapped. Rejects a self-dependency before any API round-trip.
- Both tools gated behind a new `issue_dependencies` feature flag — opt-in (user-enableable via `--features` / `X-MCP-Features`, auto-enabled in insiders mode) and kept **out** of the default toolset, mirroring the `file_blame` precedent.

## MCP impact
- [x] New tool added — two flag-gated dependency tools. Fixed-context cost added to the **default** toolset is **0** (gated); combined schema is ~700–900 tokens when the flag is enabled.

## Prompts tested (tool changes only)
- "What issues block issue #123 in owner/repo?" → `get_blocked_by`
- "What does issue #123 block?" → `get_blocking`
- "Mark issue #123 as blocked by #45" → `add` + `blocked_by`
- "Remove the dependency where #123 blocks #45" → `remove` + `blocking`

## Security / limits
- [x] Auth / permissions considered — requires `repo` scope; reads are read-only, writes use the documented dependency mutations.

## Tool renaming
- [x] I am not renaming tools as part of this PR

## Lint & tests
- [x] Linted locally with `./script/lint`
- [x] Tested locally with `./script/test`

## Docs
- [x] Updated (feature-flags and insiders-features docs)

---
Split out from #2726, which now contains only the always-on `issue_read` → `get_parent` method.

PR #2778, opened 2026-06-26
Fix OAuth scope satisfaction to AND-of-ORs (required means required)

SamMorrowDrums:
## Problem

Tool scope requirements conflate two different relationships and collapse both into a single OR list, which makes PAT tool‑filtering and OAuth scope‑challenges over‑permissive:

1. **Conjunction (AND)** — a tool needs several *different* capabilities. e.g. `ui_get` reads labels (`repo`) **and** teams (`read:org`).
2. **Hierarchy / substitution (OR)** — a single requirement is satisfied by itself *or* a higher (ancestor) scope. e.g. `read:org` is satisfied by `read:org | write:org | admin:org`.

These compose as **AND‑of‑ORs**: satisfy *every* required scope, where each required scope can be met by itself or an ancestor scope.

Previously `AcceptedScopes = ExpandScopes(required)` was checked with OR, so a token holding only `repo` was treated as satisfying `ui_get` (which also needs `read:org`): the PAT filter showed the tool and **no** scope challenge was issued, then it failed at runtime on the org call. The generated docs also rendered multiple required scopes as `Required OAuth Scopes (any of)`, which is wrong for AND requirements.

## Fix

- **`RequiredScopes` is the single source of truth** for all allow/challenge decisions. Required means required.
- **Satisfaction = AND‑of‑ORs**, implemented by expanding the *token* downward through the existing scope hierarchy (`repo → public_repo, security_events`; `admin:org → write:org, read:org`; …). A tool is satisfied iff, for *every* required scope, the expanded token set contains it.
- **`AcceptedScopes` is now display‑only** metadata and no longer drives any filter/challenge/missing decision.
- Scope challenges now request only the **missing** required scopes (merged with the user's existing scopes for the `scope=` parameter).
- Docs: removed the `(any of)` rendering — all listed required scopes are AND.

## Affected tools

Exactly 3 tools declare multiple required scopes, all `{repo, read:org}`, all genuinely AND:

- `ui_get` (`pkg/github/ui_tools.go`)
- `list_issue_types` (`pkg/github/issues.go`)
- `list_issue_fields` (`pkg/github/issue_fields.go`)

## Security tools are behavior‑neutral

All security tools (`code_scanning.go`, `secret_scanning.go`, `dependabot.go`, `security_advisories.go`) declare a **single** `{security_events}` scope. A single required scope has nothing to conjoin, and `repo` already expands to `security_events` via the hierarchy, so the AND change is **behavior‑neutral** for them: a `repo` token still satisfies them, and a `public_repo`‑only token was already hidden before this change. Explicit neutrality tests were added at both the scope and filter layers.

## Scope filtering is best‑effort, not an authorization boundary (fail‑open)

This is the guiding posture, now encoded in comments and `docs/scope-filtering.md`: scope filtering is a **best‑effort UX nicety for classic `ghp_` PATs only**, not an authorization boundary — the **GitHub API is the source of truth**. The server **fails open**: it only hides (or challenges) when confident the token cannot work, and prefers showing the tool when access is plausible. Filtering stays gated to `ghp_` classic PATs and is skipped when scopes can't be fetched. Fine‑grained / OAuth / GitHub App tokens are unaffected.

### Documented limitations (pre‑existing; intentionally **not** solved here)

The scope hierarchy only models *ancestor* substitution, so it does **not** cover every way a tool can legitimately be satisfied:

1. **Sibling scopes outside the hierarchy.** Code scanning alerts on a **public** repo are readable with `public_repo`, a *sibling* of the declared `security_events` (both children of `repo`) — token expansion can't bridge siblings, so a `public_repo`‑only PAT gets security tools hidden even though it could read public alerts.
2. **Org roles are invisible to scopes.** A *security‑manager* role grants access orthogonally to OAuth scopes; we can't see it at filter/challenge time.
3. **Public‑vs‑private repo is unknowable at filter time**, and which scope suffices depends on it.

A faithful model would make `RequiredScopes` a **CNF list of OR‑groups** (groups AND‑ed, members within a group OR‑ed), with code scanning as the motivating example: `security_events` OR `public_repo` OR `repo`. That structure is **deliberately not built** in this PR (YAGNI); the fail‑open posture keeps these cases from being wrongly hidden.

No tool declarations or scopes were changed; tool input schemas are untouched (toolsnaps unchanged).

## Validation

- `script/lint` — 0 issues
- `UPDATE_TOOLSNAPS=true go test ./...` then `go test -race ./...` — green, no `.snap` changes
- `script/generate-docs` — regenerated `README.md`, `docs/feature-flags.md`, `docs/insiders-features.md`; hand‑updated `docs/scope-filtering.md`
- `script/test` — green

Split out of the PR #2751 review discussion.

### DIFF
--- a/docs/scope-filtering.md
+++ b/docs/scope-filtering.md
@@ -4,6 +4,8 @@ The GitHub MCP Server automatically filters available tools based on your classi
 
 > **Note:** This feature applies to **classic PATs** (tokens starting with `ghp_`). Fine-grained PATs, GitHub App installation tokens, and server-to-server tokens don't support scope detection and show all tools.
 
+> **Important:** Scope filtering is a best-effort UX convenience, **not an authorization boundary**. The GitHub API is always the source of truth and enforces real permissions. The server therefore **fails open**: it only hides a tool when confident your token cannot use it, and shows the tool whenever access is plausible. See [Limitations and Fail-Open Posture](#limitations-and-fail-open-posture).
+
 ## How It Works
 
 When the server starts with a classic PAT, it makes a lightweight HTTP HEAD request to the GitHub API to discover your token's scopes from the `X-OAuth-Scopes` header. Tools that require scopes your token doesn't have are automatically hidden.
@@ -76,6 +78,18 @@ If the server cannot fetch your token's scopes (e.g., network issues, rate limit
 WARN: failed to fetch token scopes, continuing without scope filtering
 ```
 
+## Limitations and Fail-Open Posture
+
+Scope filtering is a **best-effort UX nicety**, not an authorization boundary. The GitHub API is the source of truth and enforces real permissions regardless of what the server shows. Because of this, the server is designed to **fail open**: it only hides a tool (or, for OAuth, issues a scope challenge) when it is confident the token cannot use it. When access is plausible, it prefers to show the tool and let the API decide. Filtering is also limited to classic PATs (`ghp_`) and is skipped entirely when scopes can't be fetched.
+
+A tool's declared scopes are **all required** (logical AND), and each one may be satisfied directly or by an ancestor scope from the [hierarchy](#scope-hierarchy). However, some ways a tool can legitimately be used cannot be determined from OAuth scopes alone, so the server intentionally does not try to model them:
+
+- **Sibling scopes outside the hierarchy.** The hierarchy only models *ancestor* substitution. For example, code scanning alerts on a **public** repository are readable with `public_repo`, which is a *sibling* of the declared `security_events` (both are children of `repo`), not an ancestor. Token expansion can't bridge siblings. Capturing this faithfully would require representing requirements as a list of OR-groups (groups AND-ed, members within a group OR-ed), e.g. code scanning = `security_events` OR `public_repo` OR `repo`. That model isn't implemented today; instead the server relies on the fail-open posture so these cases aren't wrongly hidden.
+- **Organization roles.** Roles such as *security manager* grant access orthogonally to OAuth scopes and are invisible to scope detection. A user may legitimately have access the server cannot see from scopes alone.
+- **Public vs. private repositories.** Whether a given scope suffices depends on the target repository's visibility, which isn't known at filter time.
+
+In each of these cases the server errs toward showing the tool; if the token truly lacks access, the API returns the appropriate error.
+
 ## Classic vs Fine-Grained Personal Access Tokens
 
 **Classic PATs** (`ghp_` prefix) support OAuth scopes and return them in the `X-OAuth-Scopes` header. Scope filtering works fully with these tokens.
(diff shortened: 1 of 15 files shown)


docs/scope-filtering.md
```diff
@@ -76,6 +78,19 @@ If the server cannot fetch your token's scopes (e.g., network issues, rate limit
 WARN: failed to fetch token scopes, continuing without scope filtering
 ```
 
+## Limitations and Fail-Open Posture
+
+Scope filtering is a **best-effort UX convenience** for classic PATs (`ghp_`) only. It is **NOT an authorization boundary** — the GitHub API is the source of truth and enforces real permissions regardless of what the server shows. The server therefore **fails open**: when access is plausible but unprovable at filter/challenge time, the tool is shown rather than hidden.
+
+A tool's declared scopes are **all required** (logical AND), and each one may be satisfied directly or by an ancestor scope from the [hierarchy](#scope-hierarchy). Some ways a tool can legitimately be used cannot be determined from OAuth scopes alone, so the scope model intentionally does not fully capture them:
+
+- **Public vs. private repositories.** Which scope suffices can depend on the target repository, which isn't known when tools are filtered. For example, code scanning alerts on **public** repos are readable with `public_repo`, while **private** repos need `security_events` (or `repo`).
```
SamMorrowDrums:
```suggestion
- **Public vs. private repositories.** Which scope suffices can depend on the target repository, which isn't known when tools are filtered. Public repos are readable with the implicit `public_repo` scope all tokens have, while **private** repos need `security_events` or `repo`.
```

docs/scope-filtering.md
```diff
@@ -76,6 +78,19 @@ If the server cannot fetch your token's scopes (e.g., network issues, rate limit
...
+- **Sibling-OR alternatives.** `security_events` and `public_repo` are *siblings* under `repo` (not parent/child), so token hierarchy expansion can't treat one as satisfying the other. A `public_repo`-only token may therefore have the security tools (code scanning, secret scanning, Dependabot, security advisories) hidden even though it could read public-repo data.
+- **Organization roles.** A *security manager* (or similar) org role grants access orthogonally to OAuth scopes and is invisible to scope filtering.
+- **Other token types.** Fine-grained PATs, OAuth, and GitHub App tokens use different permission models; filtering is skipped for them entirely (gated to `ghp_`), which is fail-open by design.
+
+These cases are deferred to runtime API enforcement. If precise sibling-OR modeling is ever needed, the extension point is making the required scopes a list of OR-groups (AND across groups, OR within a group) — deliberately not built yet.
```
SamMorrowDrums:
```suggestion
```

docs/scope-filtering.md
```diff
@@ -92,7 +106,7 @@ WARN: failed to fetch token scopes, continuing without scope filtering
 |---------|-------|----------|
 | Missing expected tools | Token lacks required scope | [Edit your PAT's scopes](https://github.com/settings/tokens) in GitHub settings |
 | All tools visible despite limited PAT | Scope detection failed | Check logs for warnings about scope fetching |
-| "Insufficient permissions" errors | Tool visible but scope insufficient | This shouldn't happen with scope filtering; report as bug |
+| "Insufficient permissions" errors | Tool visible but scope insufficient | Expected in some cases (fail-open, public/private ambiguity, org roles, or scope detection skipped). The API enforces the real boundary—grant the needed scope or access |
```
SamMorrowDrums:
```suggestion
| "Insufficient permissions" errors | Tool visible but scope insufficient | This shouldn't happen with scope filtering; report as bug |
```

Issue #2718, opened 2026-06-17
add_comment_to_pending_review runs the GraphQL query on a missing required argument instead of validating it


gustavo-sec:
### Describe the bug

`add_comment_to_pending_review` decodes its arguments with `mapstructure.WeakDecode` and never checks the required ones. An omitted required argument (e.g. `owner`) is left as the zero value and sent to the GraphQL API, which then fails downstream with a confusing message instead of a clear "this argument is required". The other pull request tools (`pull_request_read`, etc.) validate with `RequiredParam`/`RequiredInt`.

### Affected version

```
Version: v1.3.0
Commit: 34227037fc48771baea9af7163e28cb6556ef287
Build Date: 2026-06-11T14:16:55Z
```

### Steps to reproduce the behavior

1. Call `add_comment_to_pending_review` with `owner` omitted, e.g. `{ "repo": "gated-probe", "pullNumber": 1, "path": "f.go", "body": "x", "subjectType": "LINE" }`.
2. View the output.
3. The server runs the query with an empty owner and returns:
   ```
   failed to get latest review for current user: Could not resolve to a Repository with the name '/gated-probe'.
   ```
   Nothing indicates that `owner` was the missing required argument.

### Expected vs actual behavior

**Expected:** `missing required parameter: owner` (consistent with the other PR tools), before any API call.

**Actual:** the handler runs on the incomplete input and returns an unrelated downstream repository-resolution error.

### Logs

```
failed to get latest review for current user: Could not resolve to a Repository with the name '/gated-probe'.
```

> Note: the same `WeakDecode`-without-validation pattern also exists in `copilot.go` and `discussions.go` — happy to follow up separately.

PR #2785, opened 2026-06-27
fix(discussions): validate required params in read handlers

### LINKED ISSUE #2740: get_discussion and get_discussion_comments accept calls with missing required parameters instead of returning a validation error
### Describe the bug

When `get_discussion` or `get_discussion_comments` is called without one of the required parameters (`owner`, `repo`, or `discussionNumber`), the tool does not return a parameter-validation error. Instead it silently substitutes a zero value and issues a GraphQL request, which returns a confusing API-level error rather than a clear message identifying the missing field.

**Affected version:** current (tested against `ghcr.io/github/github-mcp-server` latest as of 2026-06)

### Steps to reproduce

1. Configure the MCP server normally.
2. Call `get_discussion` with only `{"repo": "myrepo", "discussionNumber": 1}` (omitting `owner`).
3. Observe that the tool returns a GraphQL error like `"Could not resolve to a Repository"` rather than a message indicating `owner` is required.

Repeat with `get_discussion_comments` and any missing required field.

### Expected behavior

The tool should return a structured error immediately: `"missing required parameter: owner"` (or equivalent), consistent with how `add_discussion_comment` and the other discussion write tools behave when a required parameter is absent.

### Actual behavior

The tool issues a GraphQL query with the missing field set to its zero value (`""` for strings, `0` for numbers) and returns whichever API error results from that invalid query.

### Additional context

The write handlers in the same file (`add_discussion_comment`, `reply_to_discussion_comment`, etc.) were updated in PR #2718 to use explicit parameter validation that returns named errors on missing input. The two read handlers were not included in that pass and still use the legacy decoding pattern.

syf2211:
## Summary

Validate required parameters in `get_discussion` and `get_discussion_comments` before issuing GraphQL queries, returning explicit `missing required parameter` errors instead of confusing API failures from zero-filled inputs.

## Motivation

When callers omit `owner`, `repo`, or `discussionNumber`, the read handlers previously used `mapstructure.WeakDecode`, which silently substituted zero values and produced GraphQL errors like "Could not resolve to a Repository". The discussion write handlers were updated in #2718 to validate inputs explicitly; the read handlers were missed.

Fixes #2740

## Changes

- Replace `mapstructure.WeakDecode` with `RequiredParam` / `RequiredInt` in `GetDiscussion` and `GetDiscussionComments`
- Remove unused `mapstructure` import from `discussions.go`
- Add regression tests for missing `owner`, `repo`, and `discussionNumber` on both tools
- Update comments in string-number tests to reflect `RequiredInt` behavior

## Tests

```bash
go test ./pkg/github/ -run 'Test_GetDiscussion|Test_GetDiscussionComments' -count=1
```

Result: PASS

## Notes

- String `discussionNumber` values (e.g. `"1"`) continue to work via `RequiredInt` / `toInt`.
- Empty-string `owner` / `repo` now fail fast with the same missing-parameter message, matching write-handler behavior.

### DIFF
--- a/pkg/github/discussions.go
+++ b/pkg/github/discussions.go
@@ -11,7 +11,6 @@ import (
 	"github.com/github/github-mcp-server/pkg/scopes"
 	"github.com/github/github-mcp-server/pkg/translations"
 	"github.com/github/github-mcp-server/pkg/utils"
-	"github.com/go-viper/mapstructure/v2"
 	"github.com/google/go-github/v89/github"
 	"github.com/google/jsonschema-go/jsonschema"
 	"github.com/modelcontextprotocol/go-sdk/mcp"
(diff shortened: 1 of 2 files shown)


CAOShurong:
I independently verified this PR at its exact head (`1ebad59ee37003286a3ce926f76a381ea2bbd2fb`, built with go1.26.7, windows/amd64):

- `go build ./pkg/github/` exit 0; `go vet ./pkg/github/` clean.
- Both `get_discussion` and `get_discussion_comments` now decode `owner` / `repo` / `discussionNumber` through `RequiredParam` / `RequiredInt` instead of `mapstructure.WeakDecode`, so omitted parameters fail fast with a clear message instead of issuing GraphQL queries with zero values. The `attachRepoVisibilityIFCLabelLazy` call sites were updated consistently.
- The two new tests (`Test_GetDiscussionRequiredParams`, `Test_GetDiscussionCommentsRequiredParams`, 6 subtests covering each missing parameter) all pass.
- Full `go test ./pkg/github/ -count=1` is green on this head.
- API reports the branch as `MERGEABLE` (behind `main`, no conflicts).

Status note for maintainers: #2741 fixes the same two handlers and is currently `CONFLICTING` with `main`, so if this one is preferred, it applies cleanly today. No action needed from me — happy to rebase-check anything else if useful.

PR #1211, opened 2025-10-13
Issues consolidation

almaleksia:
Consolidation of issue tools.

1.
`get_issue`
`get_issue_comments`
`list_labels` (with `issue_number`)
`list_sub_issues`

are now consolidated under `issue_read` with `method` parameter:

`get`
`get_comments`
`get_labels`
`get_sub_issues`

2.
`create_issue`
`update_issue`

are now consolidated under `issue_write` with `method` parameter:

`create`
`update`

3. 
`add_sub_issue`
`remove_sub_issue`
`reprioritize_sub_issue`

are consolidated under `sub_issue_write` with `method` parameter:

`add`
`reprioritize`
`remove`

Other changes:

I kept `list_labels` tool but only for listing repo labels.

UPD: Changes after benchmarks and testing

1. Added `get_comments` method  to `pull_request_read` that returns PR non-review comments. Models struggle a lot distinguishing the 2 so I decided we better have `get_comments` both in `pull_request_read` and `issue_read`.
2. Added more hints for the model to distinguish review and regular comments.

PR #2808, opened 2026-07-03
fix(issues): accept native integer types for issue_number parameters

### LINKED ISSUE #2807: `add_issue_comment`, `update_issue`, and `get_issue` reject valid integer `issue_number`
### Summary

Three endpoints that take `issue_number` as a required parameter consistently fail when invoked with an integer, even though the tool schema declares `issue_number` as `number`. Other endpoints on the same server that take integer parameters work correctly, and the failure survives reconnecting the MCP server.

### Affected tools
- `add_issue_comment`
- `update_issue`
- `get_issue`

### Unaffected tools (for comparison, same session, same repo, same auth)
- `create_issue` — succeeds (does not take `issue_number`)
- `list_issues`, `search_issues` — succeed
- `create_branch`, `create_or_update_file`, `push_files`, `create_pull_request` — all succeed, including calls with integer fields like `milestone`

### Reproduction

1. Any repo the caller can write issues on.
2. Call, e.g.:
   ```json
   {
     "tool": "add_issue_comment",
     "arguments": {
       "owner": "<owner>",
       "repo": "<repo>",
       "issue_number": 11,
       "body": "test"
     }
   }
   ```
3. Call fails. Retrying multiple times, including after fully reconnecting the MCP server, reproduces the failure.
4. `create_issue` against the same repo in the same session succeeds, ruling out auth, permissions, and rate limiting.
5. Reproduced across at least two independent sessions on the same day.

### Expected
Comment is posted / issue is updated / issue is fetched, matching a direct REST call to `POST /repos/{owner}/{repo}/issues/{issue_number}/comments` (etc.).

### Actual
Call is rejected before it reaches GitHub. The failure pattern is consistent with `issue_number` being serialised as a string somewhere in the request pipeline and then rejected by the server's own integer validation.

[TODO fill in with error string on next occurrence]

### Hypothesis

The three broken endpoints are exactly the ones where `issue_number` is templated into a URL path segment (`/repos/{owner}/{repo}/issues/{issue_number}[/comments]`), whereas the working endpoints pass their integers in the request body or query string. Likely culprits:

- A path-parameter substitution step stringifies the value before an internal validator runs.
- A schema-validation layer runs *after* path substitution and sees a string where a `number` is required.

Anywhere `issue_number` is interpolated into a URL is worth checking first.

### Affected version
- MCP server: GitHub MCP http server
- Transport: stdio.
- Client: custom provider

[Please run ` docker run -i --rm ghcr.io/github/github-mcp-server ./github-mcp-server --version` and paste the output below](https://github.com/github/github-mcp-server/pkgs/container/github-mcp-server/993100275?tag=sha256-27bf6a6927e2a9b0d3c96e6689f6dd6dcd0cd801e59d1916f2c324703b84e1da.sig)
### Workaround
None from the client side; the caller cannot influence how the server serialises the parameter. Affected users have to fall back to the GitHub REST API directly or edit the issue in the web UI.

syf2211:
## Summary

Extend `RequiredInt` / `toInt` coercion to accept native Go integer types (`int`, `int64`, etc.) and `json.Number`, not only `float64` and numeric strings. Also restore missing deprecated aliases for the issues consolidation rename.

## Motivation

Fixes #2807. Some MCP clients (including mcpcurl when schema fields use `integer`, and other providers that construct argument maps programmatically) pass `issue_number` as a native integer. `RequiredInt` previously rejected these with `expected number, got int` before any GitHub API call was made.

Additionally, `get_issue` and `update_issue` were renamed to `issue_read` and `issue_write` in #1211 but aliases were never added to `DeprecatedToolAliases`, breaking users still referencing the old tool names.

## Changes

- Add `numericToFloat64` helper and route `toInt` / `toInt64` through it
- Add deprecated aliases: `get_issue` → `issue_read`, `update_issue` → `issue_write`
- Add regression tests for `int` / `int64` in `RequiredInt` and `add_issue_comment`

## Tests

```bash
go test ./pkg/github/ -run 'Test_RequiredInt|Test_AddIssueComment' -v
```

All tests pass.

## Notes

- composer-2.5 review: **APPROVE**
- Central fix in `params.go` benefits all tools using `RequiredInt`, not only issue tools

### DIFF
--- a/pkg/github/deprecated_tool_aliases.go
+++ b/pkg/github/deprecated_tool_aliases.go
@@ -10,7 +10,10 @@ package github
 //	"get_issue": "issue_read",
 //	"create_pr": "pull_request_create",
 var DeprecatedToolAliases = map[string]string{
-	// Add entries as tools are renamed
+	// Issues tools consolidated (#1211)
+	"get_issue":    "issue_read",
+	"update_issue": "issue_write",
+
 	// Actions tools consolidated
 	"list_workflows":                 "actions_list",
 	"list_workflow_runs":             "actions_list",
--- a/pkg/github/issues_test.go
+++ b/pkg/github/issues_test.go
@@ -192,6 +192,31 @@ func Test_GetIssue(t *testing.T) {
 			expectedIssue: mockIssue,
 		},
 		{
+			name: "successful issue retrieval with int issue_number",
+			mockedClient: MockHTTPClientWithHandlers(map[string]http.HandlerFunc{
+				GetReposIssuesByOwnerByRepoByIssueNumber: mockResponse(t, http.StatusOK, mockIssue),
+			}),
+			requestArgs: map[string]any{
+				"method":       "get",
+				"owner":        "owner2",
+				"repo":         "repo2",
+				"issue_number": int(42),
+			},
+			expectedIssue: mockIssue,
+		},
+		{
+			name:         "invalid issue_number does not call API",
+			mockedClient: MockHTTPClientWithHandlers(map[string]http.HandlerFunc{}),
+			requestArgs: map[string]any{
+				"method":       "get",
+				"owner":        "owner2",
+				"repo":         "repo2",
+				"issue_number": "not-a-number",
+			},
+			expectResultError: true,
+			expectedErrMsg:    "not a valid number",
+		},
+		{
 			name: "issue not found",
 			mockedClient: MockHTTPClientWithHandlers(map[string]http.HandlerFunc{
 				GetReposIssuesByOwnerByRepoByIssueNumber: mockResponse(t, http.StatusNotFound, `{"message": "Issue not found"}`),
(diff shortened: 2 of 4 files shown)


e345ee:
Nice fix, this addresses the native `int` rejection.

One thing I’d consider tightening before merge: the new numeric path accepts `int64`/`uint64`/`json.Number`, but converts them through `float64`. That can lose precision for large integer values. Would it be worth preserving native integer types through validation and only using float parsing for actual float/string inputs?

It might also be useful to add regression coverage for all affected issue tools: `issue_read`, `issue_write`, and `add_issue_comment`, plus a no-API-call assertion for invalid `issue_number` on write tools.🥀 🥀 🥀 🥀

syf2211:
Thanks @e345ee — addressed in d15c2fb:

- `toInt` / `toInt64` now handle native integer types and `json.Number` without a float64 round-trip (large int64 values like `9007199254740993` are preserved)
- Added regression tests for `issue_read`, `issue_write`, and `add_issue_comment` with `int` `issue_number`
- Added no-API-call validation tests for invalid `issue_number` on read and write tools

`go test ./pkg/github/ -run "Test_RequiredInt|Test_RequiredBigInt|Test_GetIssue|Test_AddIssueComment|Test_UpdateIssue"` passes locally.

syf2211:
Rebased onto latest `main` (includes issue_read hierarchy enrichment) and addressed the float64 round-trip feedback in 505c6ec. `go test ./pkg/github/` passes locally.

syf2211:
Rebased onto latest `main`. `go test ./pkg/github/... -short` passes locally.

PR #2909, opened 2026-07-19
Add opt-in intent-aware Copilot issue assignment tool

boazreicher:
Adds `assign_copilot_to_issue_with_intent` in a new non-default toolset `copilot_issue_intents`. The tool uses GraphQL's object-form `assignees: [AssigneeUpdateInput!]` so intent metadata (`rationale`, `confidence`, `is_suggestion`) is attached to the Copilot entry only, while existing assignees are preserved.

Refs: github/plan-track-agentic-toolkit#683

## Design

- Object-form `assignees` (extending our existing `UpdateIssueInput`) is used instead of `assigneeIds`. The legacy `assign_copilot_to_issue` tool is unchanged; the new tool is opt-in.
- Intent metadata (`rationale`, `confidence`, `suggest`) is set only on the Copilot entry. Preserved assignees are sent with just `actorId`.
- \`agentAssignment\` remains the separate launch configuration and is only sent on direct assignments.
- The GraphQL-Features header `issues_copilot_assignment_api_support` is kept, along with the existing `repo` scope requirement.

## Behavior

- **Direct assignment** (`is_suggestion` unset or `false`): launches Copilot with the same `base_ref`, `custom_instructions`, and PR-polling behavior as `assign_copilot_to_issue`.
- **Suggestion** (`is_suggestion: true`): records a pending Copilot assignment intent, returns a suggestion-shaped result, and does not launch Copilot or poll for a linked PR. `base_ref` and `custom_instructions` are ignored in this case.
- Input validation: `rationale` capped at 280 characters (schema \`maxLength\` + runtime rune-count check); `confidence` restricted to \`LOW\`/\`MEDIUM\`/\`HIGH\` (case-insensitive at runtime, uppercase-only in schema enum).

## Toolset

New non-default `copilot_issue_intents` toolset — available via `X-MCP-Tools`/`--toolsets copilot_issue_intents`, `all`, or explicit tool selection. It is intentionally kept out of the default configuration so its inputs do not add schema bloat to the default tool surface, matching the guidance in the source issue.

## Acceptance criteria

- [x] The tool is absent with \`default\` toolsets and present via its non-default toolset, \`all\`, or explicit individual-tool selection.
- [x] Existing assignees are preserved.
- [x] \`is_suggestion: true\` creates a pending Copilot assignment intent, returns a suggestion-shaped result, and does not poll for or claim PR creation.
- [x] A direct assignment launches Copilot and retains existing base-ref, custom-instruction, and PR-polling behavior.
- [x] Invalid rationale length and confidence values fail schema/runtime validation.
- [x] Unit tests cover direct and suggested paths, toolsnaps are updated, generated docs are updated, and the relevant end-to-end coverage is extended.

## Testing

- \`script/lint\` — clean.
- \`script/test\` — all packages pass.
- New unit tests in \`pkg/github/copilot_test.go\` cover: direct assignment with existing assignees + rationale/confidence; direct assignment with \`base_ref\`/\`custom_instructions\`; suggestion path (asserts no \`agentAssignment\`, no polling, no \`pull_request\` in result); invalid rationale length; invalid confidence; and Copilot-not-available. The test also asserts the tool lives in the non-default \`copilot_issue_intents\` toolset.
- New toolsnap: \`pkg/github/__toolsnaps__/assign_copilot_to_issue_with_intent.snap\`.
- Regenerated \`README.md\` and \`docs/remote-server.md\` via \`script/generate-docs\`.
- New e2e test \`TestAssignCopilotToIssueWithIntent\` exercises the \`is_suggestion=true\` path end-to-end (skipped when Copilot is not available for the current account, mirroring the existing test).


pkg/github/__toolsnaps__/assign_copilot_to_issue_with_intent.snap
```diff
@@ -0,0 +1,69 @@
+{
+  "annotations": {
+    "idempotentHint": true,
+    "readOnlyHint": false,
+    "title": "Assign Copilot to issue with intent"
+  },
+  "description": "Assign Copilot to a specific issue with intent metadata attached to the assignment. When is_suggestion is true, a pending Copilot assignment intent is recorded rather than launching the agent; otherwise Copilot is directly assigned and begins work.\n\nThis tool can help with the following outcomes:\n- a Pull Request created with source code changes to resolve the issue (direct assignment)\n- a pending Copilot assignment suggestion is recorded on the issue for later approval (is_suggestion=true)\n\n\nMore information can be found at:\n- https://docs.github.com/en/copilot/using-github-copilot/using-copilot-coding-agent-to-work-on-tasks/about-assigning-tasks-to-copilot\n",
```
alondahari: should we trim this description a bit? it's too verbose and repeats information, imo which adds to the schema bloat

alondahari: Maybe we can have an identical description to the assign_copilot_to_issue tool, and just add to it "prefer this tool over assign_copilot_to_issue when available"?

pkg/github/copilot.go
```diff
@@ -435,6 +435,346 @@ func AssignCopilotToIssue(t translations.TranslationHelperFunc) inventory.Server
...
+						Type:        "string",
+						Description: "How confident you are in this choice. 'HIGH' for clear signal or explicit user request, 'MEDIUM' for reasonable inference with some ambiguity, 'LOW' for best guess with limited signal.",
+						Enum:        []any{"LOW", "MEDIUM", "HIGH"},
+					},
+					"is_suggestion": {
+						Type:        "boolean",
+						Description: "If true, records a pending Copilot assignment intent rather than launching the agent. Approval later supplies the launch context; base_ref and custom_instructions are ignored in this case.",
+					},
+				},
+				Required: []string{"owner", "repo", "issue_number"},
```
alondahari: should we require the intent fields?

alondahari (dismissed):

pkg/github/__toolsnaps__/assign_copilot_to_issue_with_intent.snap
boazreicher: Done — trimmed the description to mirror `assign_copilot_to_issue` with an appended "Prefer this tool over `assign_copilot_to_issue` when available" note. Removed the verbose `is_suggestion` narrative from the schema in bf2defd6.

boazreicher: Went with exactly this — description now matches `assign_copilot_to_issue` plus the preference note. See bf2defd6.

pkg/github/copilot.go
boazreicher: Done — `rationale`, `confidence`, and `is_suggestion` are now required in both the schema and at runtime. Added missing-field tests. `is_suggestion` is always sent explicitly on the Copilot `AssigneeUpdateInput` entry now. See bf2defd6.

alondahari (dismissed):

pkg/github/copilot.go
```diff
@@ -435,6 +435,346 @@ func AssignCopilotToIssue(t translations.TranslationHelperFunc) inventory.Server
...
+			return nil, nil
+		}
+		variables["endCursor"] = githubv4.String(query.Repository.SuggestedActors.PageInfo.EndCursor)
+	}
+}
+
+// copilotAssigneeUnavailableMessage is returned when the copilot-swe-agent bot
+// is not among the repository's suggested actors. The e2e tests depend on this
+// exact message to skip the test.
+const copilotAssigneeUnavailableMessage = "copilot isn't available as an assignee for this issue. Please inform the user to visit https://docs.github.com/en/copilot/using-github-copilot/using-copilot-coding-agent-to-work-on-tasks/about-assigning-tasks-to-copilot for more information."
```
RossTarrant: nit: this link redirects to https://docs.github.com/en/copilot/concepts/agents/cloud-agent/about-cloud-agent currently to just make you aware

pkg/github/copilot.go
```diff
@@ -435,6 +435,346 @@ func AssignCopilotToIssue(t translations.TranslationHelperFunc) inventory.Server
...
+				"name":   githubv4.String(params.Repo),
+				"number": githubv4.Int(params.IssueNumber),
+			}
+			if err := client.Query(ctx, &getIssueQuery, variables); err != nil {
+				return ghErrors.NewGitHubGraphQLErrorResponse(ctx, "failed to get issue ID", err), nil, nil
+			}
+
+			// Build object-form assignees: preserved assignees carry only actorId;
+			// the copilot entry carries any intent metadata provided.
+			existing := getIssueQuery.Repository.Issue.Assignees.Nodes
```
RossTarrant, line 653: If copilot is already an assignee on the issue I think it will be sent twice, once without metadata and once with metadata. Could we exclude copilot from the existing assignees?

pkg/github/copilot.go
```diff
@@ -435,6 +435,346 @@ func AssignCopilotToIssue(t translations.TranslationHelperFunc) inventory.Server
...
+				Owner              string `mapstructure:"owner"`
+				Repo               string `mapstructure:"repo"`
+				IssueNumber        int32  `mapstructure:"issue_number"`
+				BaseRef            string `mapstructure:"base_ref"`
+				CustomInstructions string `mapstructure:"custom_instructions"`
+				Rationale          string `mapstructure:"rationale"`
+				Confidence         string `mapstructure:"confidence"`
+				IsSuggestion       bool   `mapstructure:"is_suggestion"`
+			}
+			if err := mapstructure.WeakDecode(args, &params); err != nil {
```
RossTarrant, line 586: Could we explicitly check that `is_suggestion` is present before decoding? If omitted, it'll default to false and from what I understand would unexpectedly launch Copilot

pkg/github/copilot.go
boazreicher: Good catch — updated the stale link to `https://docs.github.com/en/copilot/concepts/agents/cloud-agent/about-cloud-agent` everywhere it appears (const message, both tool descriptions, and the e2e/unit-test literals that assert on the message). See 385d1a8f.

pkg/github/copilot.go
boazreicher: Fixed — the loop that copies existing assignees now skips the copilot-swe-agent actor, so we send it only once (with intent metadata). Added a unit test covering this. See 385d1a8f.

pkg/github/copilot.go
boazreicher: Fixed — presence-check `is_suggestion` on the raw args map before decoding and reject with "is_suggestion is required". Prevents the mapstructure zero-value from silently launching Copilot. Added a unit test. See 385d1a8f.

RossTarrant (approved):
Looks good, thanks for the updates!

PR #2938, opened 2026-07-23
fix: tolerate duplicate initialize requests on the same session

christensenjairus:
## Summary
go-sdk v1.7.0-pre.1 (bumped in #2787) started rejecting a second "initialize" call on an already-initialized session with `duplicate "initialize" received`, fixing modelcontextprotocol/go-sdk#961 (a duplicate initialize with changed params could silently overwrite the session's stored InitializeParams). go-sdk v1.6.1 and earlier (github-mcp-server v1.5.0) accepted repeat calls and returned the same result every time.

Some MCP clients resend "initialize" on the same transport/session when a handshake is retried (e.g. a slow first response triggers a client-side retry without tearing down the connection), which relied on that old idempotent behavior and now hard-fails on v1.6.0+. I found this change in behavior while using Stacklok [ToolHive](https://github.com/stacklok/toolhive) w/ the vMCP feature.

## Why

Fixes #(no issues for this yet)

## What changed

- Adds a receiving middleware that caches the first successful InitializeResult and replays it for later "initialize" calls on the same server, instead of forwarding to the SDK and hitting its error. The response is otherwise constant for the process's lifetime, so a single cached result (not a per-session cache) is enough and avoids any per-session memory growth

## MCP impact

- [ ] No tool or API changes
- [ ] New tool added

- 

## Security / limits

Upstream's issue (go-sdk#961) was that a duplicate initialize with changed parameters could silently overwrite ServerSession.InitializeParams() mid-session — letting a later request quietly change the negotiated protocol version/capabilities that other code relies on. My middleware short-circuits before calling next() whenever the session is already initialized, so the SDK's own ss.initialize() (the code that does the overwrite) never runs for a duplicate call. InitializeParams() stays pinned to whatever the first call sent, for the life of the session — same protection as upstream's fix. The difference is just what the client sees in response: instead of an error, it gets the original negotiated result. A second call with different params is silently ignored, never honored.

Auth interaction: I checked whether OAuth logic hooks into initialize and could be skipped by my short-circuit. On current main it doesn't — OAuth was refactored (PR #2870) into a ToolHandlerMiddleware that runs at tool-call time, not a receiving-middleware on initialize. (On v1.6.0 there was an OAuth receiving-middleware gated on initialize — worth re-checking if this is ever backported to that older base.)

Data exposure: the cached InitializeResult only holds ProtocolVersion, Capabilities, Instructions, ServerInfo — static server config, nothing per-user or secret.

Blast radius: dedupeInitializeMiddleware() is only wired into NewStdioMCPServer. The HTTP/remote path (pkg/http/handler.go) builds its server via a separate factory and never gets this middleware — so there's no scenario where one HTTP client's negotiated capabilities could leak into another's session through this cache. For stdio, there's exactly one session per process, so the single shared (not per-session) cache never crosses a trust boundary.

## Tool renaming

## Lint & tests

## Docs

- [x] Not needed

### DIFF
--- a/internal/ghmcp/server.go
+++ b/internal/ghmcp/server.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"os/signal"
 	"strings"
+	"sync"
 	"syscall"
 	"time"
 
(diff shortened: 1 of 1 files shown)


christensenjairus:
@SamMorrowDrums @IrynaKulakova Could I get a review on this? Asking you two because you did #2787

albashaalbasha4454-sudo (approved):

SamMorrowDrums:
@christensenjairus what is your target? Self-hosted http?

I ask because this will not fix our remote server in production because we have to do session middleware outside the SDK with persistence in Redis, and I'd have to look at what the impact there is.

Let me know please.

christensenjairus:
> @christensenjairus what is your target? Self-hosted http?
> 
> I ask because this will not fix our remote server in production because we have to do session middleware outside the SDK with persistence in Redis, and I'd have to look at what the impact there is.
> 
> Let me know please.

Yes, the target is the self-hosted Streamable HTTP deployment, not the production remote server.  

I’m not expecting this to fix the Redis-backed session middleware / production remote path; I understand that is outside the SDK and may need separate investigation. This PR is intended to address the duplicate initialize behavior on the self-hosted HTTP path.

SamMorrowDrums:
@christensenjairus hey, I messaged a friend at Stacklock. I don't really want to fix this, because nobody should be sending initialize requests with a session id in the first place, and can call initialize as many times as they like if they don't do this.

Considering the OSS MCP enables version pinning I don't think supporting bugs in clients perpetually is necessary. 

I would recommend contacting the vendor and raising with them and then holding back upgrade (or reverting) until they fix.

amirejaz:
ToolHive maintainer here — this is fixed on our side, so there's no need to carry a compat shim upstream. @christensenjairus, you can close this whenever you like.

@SamMorrowDrums was right that we shouldn't have been sending `initialize` with a session id; one of our proxies was stamping a backend session id onto a client's re-`initialize`, and that's fixed. There was a second bug behind the same error though: our stdio proxy multiplexes several HTTP client sessions onto a single subprocess and forwarded every `initialize` to it. No session id is involved on that path, so it failed regardless of headers. We now do the handshake once ourselves and serve later ones from a cached `InitializeResult`.

Both are in ToolHive v0.42.1. Thanks for the nudge to look at it.

PR #2961, opened 2026-07-27
Add basic project view management

zwick:
## Summary

Adds basic ProjectV2 view lifecycle support to the consolidated Projects tools: list/get views plus create/update/delete with name, layout, and filter. Create supports visible fields by database ID or case-insensitive field name.

## Why

Fixes github/planning-tracking#3776.

## What changed

- Added `list_project_views` and `get_project_view` with normalized output, pagination, and project-content IFC labels.
- Added `create_project_view`, `update_project_view`, and `delete_project_view`; create uses the public REST endpoint and can resolve `visible_field_names` to the required database IDs.
- Verify update/delete view ownership against the requested owner and project before mutation.
- Added focused lifecycle tests, tool snapshots, Projects instructions, and generated README docs.

## MCP impact

- [ ] No tool or API changes
- [x] Tool schema or behavior changed - extends the existing consolidated Projects tools with five methods.
- [ ] New tool added

- "List the views in my user project."
- "Create a table view filtered to `status:Todo` with Status and Assignees visible."
- "Get the new view, rename it, switch it to board layout, and filter to `assignee:@me`."
- "Clear the view filter, then delete the view and list the views again."

Validated through the built stdio MCP server against the public APIs without preview headers, including `visible_field_names` resolution, ownership-mismatch rejection, and final cleanup/read-back.

## Security / limits

- [x] Auth / permissions considered - existing `read:project` and `project` scopes are unchanged, and mutations verify the view belongs to the requested project.
- [x] Data exposure, filtering, or token/size limits considered - read results use project-content IFC labels and list pagination remains capped at 50.

## Tool renaming

## Lint & tests

## Docs

- [ ] Not needed

### DIFF
--- a/pkg/github/projects.go
+++ b/pkg/github/projects.go
@@ -588,7 +588,7 @@ Use this tool to get details about individual projects, project fields, project 
 					if gqlErr != nil {
 						return utils.NewToolResultError(gqlErr.Error()), nil, nil
 					}
-					resolvedIDs, resolveErr := resolveFieldNamesToIDs(ctx, gqlClient, owner, ownerType, projectNumber, fieldNames)
+					resolvedIDs, resolveErr := resolveFieldNamesToIDs(ctx, gqlClient, owner, ownerType, projectNumber, fieldNames, "fields")
 					if resolveErr != nil {
 						var structured *ghErrors.StructuredResolutionError
 						if errors.As(resolveErr, &structured) {
@@ -1228,7 +1228,7 @@ func listProjectItems(ctx context.Context, client *github.Client, gqlClient *git
 		return utils.NewToolResultError("provide either 'fields' or 'field_names', not both"), nil, nil
 	}
 	if len(fieldNames) > 0 {
-		resolvedIDs, resolveErr := resolveFieldNamesToIDs(ctx, gqlClient, owner, ownerType, projectNumber, fieldNames)
+		resolvedIDs, resolveErr := resolveFieldNamesToIDs(ctx, gqlClient, owner, ownerType, projectNumber, fieldNames, "fields")
 		if resolveErr != nil {
 			var structured *ghErrors.StructuredResolutionError
 			if errors.As(resolveErr, &structured) {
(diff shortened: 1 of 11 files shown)


pkg/github/projects.go
```diff
@@ -1678,6 +1859,352 @@ func getProjectStatusUpdate(ctx context.Context, gqlClient *githubv4.Client, sta
...
+	}
+	visibleFields, err := OptionalBigIntArrayParam(args, "visible_fields")
+	if err != nil {
+		return utils.NewToolResultError(err.Error()), nil, nil
+	}
+	if len(visibleFields) > 0 && len(visibleFieldNames) > 0 {
+		return utils.NewToolResultError("provide either 'visible_fields' or 'visible_field_names', not both"), nil, nil
+	}
+	if len(visibleFieldNames) > 0 {
+		resolvedIDs, resolveErr := resolveFieldNamesToIDs(ctx, gqlClient, owner, ownerType, projectNumber, visibleFieldNames)
```
tommaso-moro: This helper’s errors recommend `fields` (in projects_resolver.go at line 569) , but this method only supports `visible_fields`. Might be worth making the message caller-appropriate?

pkg/github/projects.go
```diff
@@ -1678,6 +1859,352 @@ func getProjectStatusUpdate(ctx context.Context, gqlClient *githubv4.Client, sta
...
+		}
+		requestBody.Filter = &value
+	}
+
+	var endpoint string
+	switch ownerType {
+	case "org":
+		endpoint = fmt.Sprintf("orgs/%s/projectsV2/%d/views", owner, projectNumber)
+	case "user":
+		endpoint = fmt.Sprintf("users/%s/projectsV2/%d/views", owner, projectNumber)
```
tommaso-moro, line 2065: `owner` is a login, but I _think_ this endpoint requires the numeric `user_id` (`octocat` → `583231`). If so, we should maybe resolve the user ID before building the URL and update the test? [source](https://docs.github.com/en/rest/projects/views?apiVersion=2026-03-10)

tommaso-moro (commented):
Left two comments, otherwise lgtm!

pkg/github/projects.go
zwick: Great catch! This is actually an error in the docs. Verified against production:

```
POST /users/zwick/projectsV2/7/views
{"name":"Route evidence - login","layout":"table"}

HTTP 201
{"name":"Route evidence - login","project_url":"https://api.github.com/users/zwick/projectsV2/7"}
```

```
POST /users/1871037/projectsV2/7/views
{"name":"Route evidence - numeric ID","layout":"table"}

HTTP 404
{"message":"Not Found","status":"404"}
```

I'll see about getting these docs updated

pkg/github/projects.go
zwick: Whoops! This one got missed when some naming changed during iteration. Update incoming for this

tommaso-moro (approved):

Issue #2981, opened 2026-07-30
`create_or_update_file` content description doesn't specify plaintext, leading models to double-encode


ericsciple:
### Describe the bug

The `content` parameter of `create_or_update_file` expects **plaintext**, but its description
does not say so, and the REST endpoint the tool wraps expects **base64**. Models frequently
resolve that ambiguity by base64-encoding the content themselves, which the server then encodes
again. The result is a successful API call that commits base64 text as the file's contents, with
no error anywhere.

The schema says only:

```json
"content": { "description": "Content of the file", "type": "string" }
```

The server treats it as plaintext and encodes exactly once
([`pkg/github/repositories.go#L484-L485`](https://github.com/github/github-mcp-server/blob/main/pkg/github/repositories.go#L484-L485)):

```go
// json.Marshal encodes byte arrays with base64, which is required for the API.
contentBytes := []byte(content)
```

That is correct behavior. The problem is that a model reading the tool description has no way to
know it, while it has very likely seen
[`PUT /repos/{owner}/{repo}/contents/{path}`](https://docs.github.com/en/rest/repos/contents#create-or-update-file-contents),
whose `content` field is documented as "The new file content, using Base64 encoding." Given a
parameter named `content` on a tool named `create_or_update_file`, pre-encoding is a reasonable
inference.

The failure is silent in both directions: the tool returns success, and the resulting file looks
fine to the API. It surfaces only when a human opens the file and finds one long base64 line.
Decoding it yields exactly the content the model intended to write.

### Affected versions

Not version-specific. The description is byte-identical in `v1.6.0`, `v1.8.0`, and `main`, so
upgrading does not change it. The remote server runs the same code and is equally affected.

### Steps to reproduce

1. Point an agent at a repository with `create_or_update_file` available.
2. Ask it to modify an existing text file and commit the result.
3. Inspect the committed file.

Whether the model pre-encodes is model- and prompt-dependent, so this does not reproduce 100% of
the time. That is arguably what makes it worth fixing: it is an intermittent, silent corruption
that depends on how a given model resolves an ambiguous parameter description.

### Expected behavior

The description states the expected encoding, so a model does not have to guess.

### Suggested fix

A description change is probably sufficient:

```json
"content": {
  "description": "Content of the file as plain text. Do NOT base64-encode it; the server handles encoding.",
  "type": "string"
}
```

Optionally, defensively reject input that is unambiguously already-encoded (valid base64 whose
[...]

PR #2983, opened 2026-07-30
Clarify that create_or_update_file content is plain text, not base64

ericsciple:
Fixes the ambiguity reported in #2981.

`content` is passed to the API as plain text and this server base64-encodes it, but the description said only "Content of the file". The REST endpoint this wraps documents *its* `content` field as base64, so a model reading the tool description has a strong reason to encode the value itself. When it does, the server encodes again and the file is committed containing base64 text, with success reported at every layer.

The new description says how the value should end up on disk rather than only what not to do, so a file whose contents are legitimately base64 stays unambiguous. It also names the encoding step, so the conflict with the REST API docs is resolved rather than merely overridden.

```
Content of the file, exactly as it should appear once written. Do not base64-encode it;
this server does that before calling the REST API.
```

Scoped to `create_or_update_file` only. `push_files` and the gist tools pass content through without encoding, so this bug does not apply to them (see the follow-up comment on #2981).

Regenerated toolsnaps and README per CONTRIBUTING. `go test ./...` and `script/lint` both pass.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1278,7 +1278,7 @@ The following sets of tools are available:
 - **create_or_update_file** - Create or update file
   - **Required OAuth Scopes**: `repo`
   - `branch`: Branch to create/update the file in (string, required)
-  - `content`: Content of the file (string, required)
+  - `content`: Content of the file, exactly as it should appear once written. Do not base64-encode it; this server does that before calling the REST API. (string, required)
   - `message`: Commit message (string, required)
   - `owner`: Repository owner (username or organization) (string, required)
   - `path`: Path where to create/update the file (string, required)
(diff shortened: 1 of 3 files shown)


RossTarrant (approved):

mestamosto-collab:
> ## Pull request overview
> Clarifies that `create_or_update_file` accepts content exactly as it should appear on disk, preventing accidental double base64 encoding.
> 
> **Changes:**
> 
> * Clarified the `content` parameter description.
> * Regenerated documentation and schema snapshot.
> 
> Show a summary per file
> File	Description
> `README.md`	Updates generated tool documentation.
> `pkg/github/repositories.go`	Clarifies the content input contract.
> `pkg/github/__toolsnaps__/create_or_update_file.snap`	Updates the tool schema snapshot.
> ## Review details
> * **Files reviewed:** 3/3 changed files
> * **Comments generated:** 0
> * **Review effort level:** Balanced

/

PR #2999, opened 2026-08-03
Support removing issue types

zwick:
## Summary

Allow consolidated `issue_write` and granular `update_issue_type` to remove an issue type with explicit JSON `null`.

## Why

Related: github/plan-track-agentic-org#193

## What changed

- Preserve omitted/set/clear semantics with `anyOf` schemas (`string` with `minLength: 1` or `null`), avoiding nullable type arrays for client compatibility noted in #325.
- Preserve the same tri-state intent in the `issue-write` MCP App: omitted values may use existing-issue prefill, strings select a type, and explicit null or “Clear selection” submits null.
- Send explicit clears as one REST PATCH containing `"type": null` and render schema unions as `string | null` in generated docs.
- Reject suggestion metadata on granular removals because backend and live validation show removal is applied immediately without creating a pending suggestion.

## MCP impact

- [ ] No tool or API changes
- [x] Tool schema or behavior changed — explicit null removes an issue type; empty strings are invalid.
- [ ] New tool added

- "Set this issue type to Bug"
- "Remove the type from this issue"

## Security / limits

- [x] No security or limits impact — this uses the existing issue update permission and endpoint.

## Tool renaming

## Lint & tests

Focused Go tests cover omitted, string, null, empty-string rejection, exact PATCH JSON, granular suggestions, and metadata rejection without a request. The MCP App passes typecheck and production build. Existing live validation confirmed set and clear through the built server; separate backend/live validation confirmed suggested null removal clears immediately with no pending suggestion.

## Docs

- [ ] Not needed
- [x] Updated (README / docs / examples) — generated references render nullable unions.

### DIFF
--- a/pkg/github/issues.go
+++ b/pkg/github/issues.go
@@ -2150,8 +2150,11 @@ Options are:
 						Description: "Milestone number",
 					},
 					"type": {
-						Type:        "string",
-						Description: "Type of this issue. Only use if issue types are enabled for this repository. Use list_issue_types tool to get valid type values for this repository or its owner organization. If the repository doesn't support issue types, omit this parameter.",
+						AnyOf: []*jsonschema.Schema{
+							{Type: "string", MinLength: jsonschema.Ptr(1)},
+							{Type: "null"},
+						},
+						Description: "Type of this issue. For updates, pass null to remove the current type. Only use if issue types are enabled for this repository. Use list_issue_types to get valid type values for this repository or its owner organization. If the repository doesn't support issue types, omit this parameter.",
 					},
 					"state": {
 						Type:        "string",
@@ -2275,9 +2278,13 @@ Options are:
 			}
 
 			// Get optional type
-			issueType, err := OptionalParam[string](args, "type")
+			issueTypeParam, issueTypeProvided, err := OptionalNullableStringParam(args, "type")
 			if err != nil {
 				return utils.NewToolResultError(err.Error()), nil, nil
+			}
+			issueType := ""
+			if issueTypeParam != nil {
+				issueType = *issueTypeParam
 			}
 
 			// Handle state, state_reason and duplicateOf parameters
@@ -2336,6 +2343,7 @@ Options are:
 				result, err := UpdateIssue(ctx, client, gqlClient, owner, repo, issueNumber, title, body, assignees, labels, milestoneNum, issueType, issueFieldValues, fieldIDsToDelete, state, stateReason, duplicateOf, UpdateIssueOptions{
 					AssigneesProvided: assigneesProvided,
 					LabelsProvided:    labelsProvided,
+					IssueTypeProvided: issueTypeProvided,
 				})
 				return result, nil, err
 			default:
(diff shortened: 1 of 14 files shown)


pkg/github/issues.go
```diff
@@ -2224,7 +2224,15 @@ Options are:
...
 			// execute now (see shouldDeferToForm).
-			if shouldDeferToForm(ctx, deps, req, args, issueWriteFormParams) {
+			deferToForm := shouldDeferToForm(ctx, deps, req, args, issueWriteFormParams)
+			if method == "update" {
+				if issueType, ok := args["type"].(string); ok && issueType == "" {
+					// The form replaces an empty type with the current type, so execute
+					// directly to preserve the clear and any co-submitted values.
+					deferToForm = false
+				}
+			}
```
zwick: This was a little awkward, but a previous attempt to handle this null|string value was requiring a larger change to the Form to support a null value. Added this to bypass that deferment instead of changing how the form validates `nullable` data to keep this change focused

alirezainanlosalehi7-lgtm: -

pkg/github/issues_granular.go
```diff
@@ -770,14 +773,16 @@ func GranularUpdateIssueType(t translations.TranslationHelperFunc) inventory.Ser
...
-
 			client, err := deps.GetClient(ctx)
 			if err != nil {
 				return utils.NewToolResultErrorFromErr("failed to get GitHub client", err), nil, nil
 			}
 
 			var body any
-			if rationale != "" || isSuggestion || confidence != "" {
+			switch {
+			case issueType == "":
```
tommaso-moro: Correct me if I am wrong here but it seems that with this input `{"issue_type":"","is_suggestion":true}` we'd lose `is_suggestion: true` ?

tommaso-moro (commented):
Suggested clears are supported by the API using `{"type" {"value":null,"suggest":true}}`. However, looking at `issues_granolar.go` it seems to be that `{"issue_type":"","is_suggestion":true}` currently sends `{"type":null}` and clears the type immediately? Is this intentional or should we preserve the suggestion flag here?

alirezainanlosalehi7-lgtm:
1✨✨✨✨✨✨✨✨✨✨✨✨

pkg/github/issues_granular.go
zwick: Nice catch! Clearing now explicitly uses `null` and we reject suggestion metadata on removals. Empty strings are invalid, while suggestions for non-null types still work as before.

pkg/github/issues.go
zwick: I've added the missing pieces to the docs generator and the app form to support anyOf for nullable fields.

tommaso-moro:
Looking at `issues.go` I think something like `{"type":null,"state":"closed","state_reason":"duplicate"}` currently clears the type first (due to `"type":null,`) , and then returns an error because `duplicate_of` is missing, but I think it should validate the logic and fail before making any API request?

zwick:
@tommaso-moro Very nice catch. Adding a validation to prevent this from making the API call accompanied with a test

tommaso-moro (approved):
lgtm 🚀

PR #3020, opened 2026-08-05
Add non-default find_duplicate tool gated by duplicate_detection flag

michaeljacholke:
## Summary
Adds a read-only `find_duplicate` MCP tool that returns ranked duplicate candidates for an existing issue, gated behind a new opt-in `duplicate_detection` feature flag.

## Why
Enables agents to answer "what likely duplicates this issue?" using GitHub's semantic-similarity endpoint, without adding schema cost to agents that haven't opted into duplicate detection. Semantic ranking stays server-side; the tool is a thin, read-only client.

Closes https://github.com/github/plan-track-agentic-toolkit/issues/690

## What changed
- Add `find_duplicate` tool (Issues toolset, read-only) calling `GET /repos/{owner}/{repo}/issues/{issue_number}/semantically_similar`
- Add `duplicate_detection` feature flag to `AllowedFeatureFlags` (deliberately not in `InsidersFeatureFlags` so it's never implicitly enabled)
- Map `confidence_threshold`→`threshold`, `perPage`→`per_page`, `page`; omit optional params when unset so the API's high-precision defaults apply. No client-side bounds on `confidence_threshold` — the API owns the scale (observed scores exceed 1)
- Return a trimmed candidate — issue `{ number, title, state, url }` (via the existing `MinimalIssueRef`) plus ranking metadata (`score` nullable, `confidence`, `likely_duplicate`) — rather than the full issue/repository payload
- Surface a clear error for legacy bare-issue responses instead of a success-shaped empty list
- Add unit tests, tool snapshot, and regenerate `docs/feature-flags.md`

## MCP impact
- [x] New tool added
  - Adds `find_duplicate` to the Issues toolset. It is non-default and only registered when the `duplicate_detection` feature flag is enabled via `--features` or `X-MCP-Features`.

Tested against a local build (`--features duplicate_detection`, `repo`-scoped PAT), via both Agent-mode prompts and direct JSON-RPC calls:

- Gating: `find_duplicate` is absent by default and present in `tools/list` only with the flag.
- "Using find_duplicate, find likely duplicates of issue 769 in github/issues-agentic-sandbox" → returned 585 ("Improve the onboarding flow for new users"), score ~1.93, confidence "high", likely_duplicate true.
- "Show the top 5 possible duplicates of issue 769 in that repo" (exercises `perPage`/`confidence_threshold`) → API returned the single genuine candidate; no error.
- Empty case — issue with no similar issues → `[]` (successful, not an error).
- Error case — a repo/issue where the endpoint isn't accessible → API `404` surfaced as an MCP tool error (`isError: true`), not a success-shaped empty result.

## Security / limits
  - Uses the caller's existing GitHub authentication; repository visibility and issue authorization are enforced by the API. The tool is read-only and mutates nothing.
  - Results are scoped to the source issue's repository (API-enforced), trimmed to a compact candidate shape to limit response size, and support pagination via `perPage`/`page`.

## Tool renaming

## Lint & tests

## Docs
  - Regenerated `docs/feature-flags.md` (documents the `duplicate_detection` opt-in and the read-only, single-repository `find_duplicate` tool). README needs no change — gated tools are intentionally excluded from its tool listing.

### DIFF
--- a/docs/feature-flags.md
+++ b/docs/feature-flags.md
@@ -338,4 +338,15 @@ runtime behavior (such as output formatting) won't appear here.
     - 'blocked_by' - the subject issue is blocked by the related issue.
     - 'blocking' - the subject issue blocks the related issue. (string, required)
 
+### `duplicate_detection`
+
+- **find_duplicate** - Find duplicate issues
+  - **Required OAuth Scopes**: `repo`
+  - `confidence_threshold`: Minimum similarity threshold for a candidate to be returned. When omitted, the API's high-precision default is used. (number, optional)
+  - `issue_number`: The number of the existing issue to find duplicates for (number, required)
+  - `owner`: The owner of the repository (string, required)
+  - `page`: Page number for pagination (min 1) (number, optional)
+  - `perPage`: Results per page for pagination (min 1, max 100) (number, optional)
+  - `repo`: The name of the repository (string, required)
+
 <!-- END AUTOMATED FEATURE FLAG TOOLS -->
--- a/docs/feature-flags.md
+++ b/docs/feature-flags.md
@@ -342,7 +342,7 @@ runtime behavior (such as output formatting) won't appear here.
 
 - **find_duplicate** - Find duplicate issues
   - **Required OAuth Scopes**: `repo`
-  - `confidence_threshold`: Minimum similarity threshold for a candidate to be returned. When omitted, the API's high-precision default is used. (number, optional)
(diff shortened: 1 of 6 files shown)


tommaso-moro (dismissed):
Non-blocking but I wonder if `find_duplicate` should attach an IFC label to the successful result returned here?

Example usage: https://github.com/github/github-mcp-server/blob/main/pkg/github/issues.go#L688-L696

michaeljacholke:
> Non-blocking but I wonder if `find_duplicate` should attach an IFC label to the successful result returned here?

Great suggestion, thanks @tommaso-moro. Added an IFC label to the successful result - [d5a2e4d](https://github.com/github/github-mcp-server/pull/3020/commits/d5a2e4ddcfe460017ab6d370556d3eaa80076dea)

tommaso-moro (approved):

PR #3035, opened 2026-08-07
fix(github): sanitize bodies on all issue and pull request read paths

SashaMIT:
## Problem

The server sanitizes issue/PR titles and bodies on read, but two body-bearing read paths missed the control entirely:

1. **`convertToMinimalIssueComment`** (`minimal_types.go`) — comment bodies were returned verbatim, so every `get_comments` read delivered raw attacker-controlled text.
2. **`GetSubIssues`** (`issues.go`) — sub-issue titles/bodies were marshalled verbatim.

A hostile comment or sub-issue could carry hidden prompt-injection payloads — invisible Unicode tag blocks or BiDi overrides, invisible to a human reviewer but legible to the model — straight into the context window.

Fixing that surfaced a second problem in the paths that *were* covered. `sanitize.Sanitize` ends with `FilterHTMLTags`, which runs bluemonday over the content:

```
"a < b && c > d"        -> "a &lt; b &amp;&amp; c &gt; d"
"```go\nif a<b {}\n```"  -> "```go\nif a"        // rest silently dropped
"use <Foo/> component"  -> "use  component"
```

Bodies are the most code-dense fields the server returns — generics, JSX, shell redirects, diffs. Truncating a fenced code block at the first `<` corrupts the content delivered to the model, which is worse than the injection vector it defends against.

## Fix

Add `sanitize.FilterBody` — invisible-character filtering plus code-fence metadata stripping, without the HTML pass — and use it for every body on a read path:

- issue comment bodies (`convertToMinimalIssueComment`)
- PR review bodies (`convertToMinimalPullRequestReview`)
- PR review comment bodies (`convertToMinimalReviewComment`)
- sub-issue bodies (`GetSubIssues`)
- issue bodies (`GetIssue`, `fragmentToMinimalIssue`)
- pull request bodies (`GetPullRequest`, PR list)

Titles keep the full `sanitize.Sanitize` treatment: they are single-line and not code-bearing, so HTML filtering is safe there. Lockdown-mode filtering is untouched and orthogonal.

The injection payloads that motivated this — Unicode tag characters (U+E0000 block), BiDi overrides, zero-width joiners, and hidden code-fence info strings — are all still stripped; only the HTML pass is dropped for bodies.

## Testing

- `TestSanitizeBody` in `pkg/sanitize`: tag characters, BiDi overrides and hidden fence metadata are stripped; angle brackets, fenced code containing `<`, and HTML-like markup survive.
- `pkg/github/minimal_types_test.go`: per-converter coverage for comment, review, review-comment and GraphQL issue-fragment bodies.
- `Test_GetSubIssues_Sanitization`: end-to-end through the `issue_read` handler.

`script/lint` and `script/test` pass. No tool schema or response-shape change, so toolsnaps and README are unaffected.


IrynaKulakova (approved):
Approving. The read-path gap is real and the fix is now consistent across every body-bearing path.

Note that this PR was iterated on during review: bodies use `sanitize.FilterBody` (invisible characters + code-fence metadata) rather than the full `Sanitize`, because the bluemonday pass silently truncates a fenced code block at the first `<`. Titles still get the full treatment. Test coverage was added at the sanitize unit level, per-converter, and end-to-end through the `issue_read` handler.

Disclosure: I pushed the follow-up commits (57d4875, 9915997, 585519e) to this branch, so a second maintainer eye on those would be good before merge.

SashaMIT:
Thanks Iryna. Glad the FilterBody vs full Sanitize cut is the one that keeps fenced code intact, and that the tests now cover unit, converter, and the issue_read path.

Appreciate you landing the follow-ups on the branch. Happy to wait on a second maintainer eye before merge.

PR #3050, opened 2026-08-11
Fix static --tools validation fallback

### LINKED ISSUE #2933: HTTP static --tools validation falls back to the full static inventory
### Describe the bug

For the current main revision, the HTTP static-inventory path handles an
invalid EnabledTools entry differently from the documented startup behavior.

The static builder returns the existing unknown-tool error, but
buildStaticInventory substitutes the full local tool/resource/prompt inventory.
The default HTTP inventory factory then uses that widened static universe.

This is inconsistent with the README and configuration troubleshooting guide,
which say that invalid --tools names cause startup failure.

### Affected version

Source-pinned local review at:

a217a7f43a76a4cd6909ba6814d88a886064488c

This was the repository's main revision during the local preflight. No
deployed server, account, token, or external service was tested.

The local source-pinned fixture used an HTTP ServerConfig with:

~~~go
EnabledTools: []string{"get_file_contents", "synthetic_unknown_tool"}
~~~

1. Construct the current HTTP handler and default inventory path with
   in-memory request/response objects only.
2. Compare the mixed configuration above with a valid
   get_file_contents-only control.
3. Observe that the valid control omits create_repository, while the mixed
   configuration carries create_repository into the local tools/list response
   through the current default factory path.

The fixture used inert dependencies and did not create a listener, network
connection, external client, account, or tool invocation.

Expected: an invalid static EnabledTools entry preserves the typed unknown-tool
error and prevents the HTTP static inventory from being constructed.

Actual: the static helper replaces that error with the full static universe,
which makes the malformed configuration broader than its valid one-tool
control in the local fixture.

### Logs

No production logs were collected. The source-bound local evidence and a
focused candidate patch are ready to share in a follow-on pull request if this
behavior is confirmed as unintended.

Mahmoud772122777:
Fixes #2933

Invalid static --tools configuration was previously falling back
to the full tool inventory instead of propagating the validation error.

This change propagates the inventory build error and adds regression
coverage for invalid static --tools configuration.

### DIFF
--- a/pkg/http/handler.go
+++ b/pkg/http/handler.go
@@ -267,7 +267,7 @@ func DefaultGitHubMCPServerFactory(r *http.Request, deps github.ToolDependencies
 func DefaultInventoryFactory(cfg *ServerConfig, t translations.TranslationHelperFunc, featureChecker inventory.FeatureFlagChecker, scopeFetcher scopes.FetcherInterface) InventoryFactoryFunc {
 	// Build the static tool/resource/prompt universe from CLI flags.
 	// This is done once at startup and captured in the closure.
-	staticTools, staticResources, staticPrompts := buildStaticInventory(cfg, t)
+	staticTools, staticResources, staticPrompts, staticErr := buildStaticInventory(cfg, t)
 	hasStaticFilters := hasStaticConfig(cfg)
 
 	// Pre-compute valid tool names for filtering per-request tool headers.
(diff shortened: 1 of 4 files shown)


De-ASI-INTERFACE (approved):

nishmasbull-bit (changes requested):
Plese

SamMorrowDrums (approved):

annaalieva900-pixel (commented):
@SweetTy69

PR #3077, opened 2026-08-15
fix(issues): allow delete:false in issue_write issue_fields

### LINKED ISSUE #2904: blocked bug from using MCP server in Antigravity IDE
Image

***

### Describe the bug

When using the `github-mcp-server` with an LLM client or gateway that integrates with the Gemini API (such as Antigravity), tool initialization/registration fails with a `400 INVALID_ARGUMENT` error. 

This is caused by an invalid JSON Schema definition for the `delete` parameter within the `issue_fields` array property of the `issue_write` tool. The schema defines `delete` with `Enum: []any{true}`. Because the Gemini API's protobuf representation only supports string enums (`repeated string enum`), non-string values like the boolean `true` are serialized into an empty string (`""`), which fails the API gateway validation.

### Affected version

```
GitHub MCP Server
Version: v1.5.0
Commit: 8cd03c018525ae0bafc9b3cdb84ec2133e01bac2
Build Date: 2026-06-27T07:57:40Z
```

1. Start `github-mcp-server` in stdio mode.
2. Launch a client that bridges the MCP server's JSON schema tool declarations to the Gemini API (such as Antigravity).
3. Attempt to register the `issue_write` tool schema with the Gemini API.
4. The Gemini API rejects the request, returning a `400 INVALID_ARGUMENT` response indicating `enum[0]` is empty.

* **Expected:** The `issue_write` tool schema registers successfully.
* **Actual:** Registration fails with `INVALID_ARGUMENT: enum[0]: cannot be empty` because of the boolean enum constraint on the `delete` parameter. Removing `Enum: []any{true}` (making it a clean boolean type parameter, which matches the schema used in the granular `set_issue_fields` tool) resolves the error.

### Logs

```json
{
  "error": {
    "code": 400,
    "message": "* GenerateContentRequest.tools[51].function_declarations[0].parameters.properties[issue_fields].items.properties[delete].enum[0]: cannot be empty\n",
    "status": "INVALID_ARGUMENT"
  }
}
```

### Fix
kindly, check the following pull request 
fix: remove invalid boolean enum from issue_write delete parameter this appears in antigravity IDE cause error
[https://github.com/github/github-mcp-server/pull/2887](https://github.com/github/github-mcp-server/pull/2887)

tgockel:
## Summary

Removes `Enum: []any{true}` from the `delete` property of `issue_write`'s `issue_fields` item schema, so `delete: false` is a legal no-op instead of a schema-validation error.

## Why

`delete` was declared `{"type": "boolean", "enum": [true]}`. The property is optional, but a client that fills every property of a schema — common, since OpenAI-style strict function calling requires every property to appear in `required` — had no way to say "not deleting this field": there is no `false` in the enum and no `null` in the type. The sibling `value` property offers no alternative either, being typed `["string","number","boolean"]` with no `null`, and `issues.go` explicitly rejects `value: null`.

This is not only cosmetic. The MCP Go SDK validates arguments against the resolved input schema before the handler runs (`go-sdk@v1.7.0/mcp/server.go`, `applySchema`), so `delete: false` was rejected at validation and never reached `optionalIssueWriteFields`. Clients that fill every property instead sent `delete: true` alongside a `value` and hit the handler's mutual-exclusion check, so `issue_write` could never set an issue field for them.

## What changed
  
- `pkg/github/issues.go`: dropped `Enum: []any{true}` from the `delete` property of the `issue_fields` item schema; extended its description to state that omitting the property, or setting it to `false`, leaves the field's current value unchanged.
- `pkg/github/__toolsnaps__/issue_write.snap`: regenerated via `UPDATE_TOOLSNAPS=true`.
- `pkg/github/issues_test.go`: added `Test_optionalIssueWriteFields`, covering `delete: false` alongside a value, `delete: true` alone, `delete` omitted, and both existing error paths. This function previously had no test coverage.

No handler logic changed. `optionalIssueWriteFields` already reads `OptionalParam[bool](itemMap, "delete")` and branches on `if deleteField`, so `false` falls through to the normal value path, and the `delete: true` mutual-exclusion check still applies.

## MCP impact

- [ ] No tool or API changes
- [ ] New tool added

`issue_write.issue_fields[].delete` widens from `enum: [true]` to any boolean, and its description gains the no-op case. Strictly a relaxation: every payload valid before is still valid and behaves identically.

- Worked against a live GitHub repository with org-level issue fields configured.
- Coverage is the new unit tests over `optionalIssueWriteFields` plus the regenerated schema snapshot.
- The reproducing payload, from an agent framework against `api.githubcopilot.com/mcp/`, was `{"method": "update", "issue_number": 5, "issue_fields": [{"field_name": "Start date", "value": "2026-08-14", "delete": true, "field_option_name": ""}]}` — rejected with `issue field "Start date" cannot specify 'delete' together with 'value' or 'field_option_name'` on seven consecutive calls across two runs. That client also sent `milestone: 0`, `duplicate_of: 0`, and `field_option_name: ""`; only `delete` had no benign value. The equivalent prompt is "Set the Start date field on issue #5 to 2026-08-14".
- Caveat: this does not guarantee such a client starts succeeding. Whether a given model emits `false` rather than `true` is unverified, and one copying its own prior call will keep sending `true`. The change removes a schema construct that *forces* failure on an otherwise recoverable mistake.

## Security / limits

Input validation is relaxed for one boolean that routes to an existing code path. No auth, permission, data-exposure, or size-limit surface is touched, and the mutual-exclusion check that prevents `delete: true` from being combined with `value` is unchanged.

## Tool renaming
  
## Lint & tests
  
- [x] Linted locally with `./script/lint` — `0 issues.`
- [x] Tested locally with `./script/test` — full suite passes, including the snapshot check without `UPDATE_TOOLSNAPS` set.
  
## Docs
  
- [x] Not needed
  
Ran `./script/generate-docs`; it produced no diff. The generated tables do not include nested item-property descriptions, and a repo-wide grep found the old description only in `issue_write.snap`.

Fixes #2904.

### DIFF
--- a/pkg/github/issues.go
+++ b/pkg/github/issues.go
@@ -2447,9 +2447,10 @@ Options are:
 								},
 								"delete": {
 									Type: "boolean",
-									Enum: []any{true},
 									Description: "Set to true to clear this field's current value on the " +
-										"issue. Cannot be combined with 'value' or 'field_option_name'.",
+										"issue. Cannot be combined with 'value' or 'field_option_name'. " +
+										"Omit this property, or set it to false, to leave the field's " +
+										"current value unchanged.",
 								},
 							},
 							Required: []string{"field_name"},
--- a/pkg/github/issues.go
+++ b/pkg/github/issues.go
@@ -265,7 +265,10 @@ func optionalIssueWriteFields(args map[string]any) ([]issueWriteFieldInput, erro
 			return nil, err
 		}
 
-		deleteField, _ := OptionalParam[bool](itemMap, "delete")
+		deleteField, err := OptionalParam[bool](itemMap, "delete")
+		if err != nil {
+			return nil, err
+		}
 		value, hasValue := itemMap["value"]
 		if hasValue && value == nil {
 			return nil, fmt.Errorf("value cannot be null for field %q", fieldName)
@@ -2437,20 +2440,19 @@ Options are:
 									Description: "Value to set. Use for text, number, and date fields " +
 										"(date as YYYY-MM-DD). For single-select fields, prefer " +
 										"'field_option_name' so the option is validated before the API " +
-										"call. Cannot be combined with 'field_option_name' or 'delete'.",
+										"call. Cannot be combined with 'field_option_name' or 'delete: true'.",
 								},
 								"field_option_name": {
 									Type: "string",
 									Description: "Option name for single-select fields. Validated against " +
 										"the field's options before the API call. Cannot be combined with " +
-										"'value' or 'delete'.",
+										"'value' or 'delete: true'.",
 								},
 								"delete": {
 									Type: "boolean",
 									Description: "Set to true to clear this field's current value on the " +
-										"issue. Cannot be combined with 'value' or 'field_option_name'. " +
-										"Omit this property, or set it to false, to leave the field's " +
-										"current value unchanged.",
+										"issue. When false or omitted, this property is ignored. Cannot " +
+										"be true when 'value' or 'field_option_name' is provided.",
 								},
 							},
 							Required: []string{"field_name"},
--- a/pkg/github/__toolsnaps__/issue_write.snap
+++ b/pkg/github/__toolsnaps__/issue_write.snap
@@ -37,10 +37,7 @@
           "additionalProperties": false,
           "properties": {
             "delete": {
-              "description": "Set to true to clear this field's current value on the issue. Cannot be combined with 'value' or 'field_option_name'.",
-              "enum": [
-                true
-              ],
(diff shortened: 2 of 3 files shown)


pkg/github/issues.go
```diff
@@ -2287,9 +2287,10 @@ Options are:
 								},
 								"delete": {
 									Type: "boolean",
-									Enum: []any{true},
 									Description: "Set to true to clear this field's current value on the " +
-										"issue. Cannot be combined with 'value' or 'field_option_name'.",
+										"issue. Cannot be combined with 'value' or 'field_option_name'. " +
+										"Omit this property, or set it to false, to leave the field's " +
+										"current value unchanged.",
```
IrynaKulakova: `delete: false` does not necessarily leave the field unchanged: when the same item includes `value` or `field_option_name`, the handler updates the field (and the new test covers this). Could this say that false means "do not clear the field" or that it is ignored, rather than implying the entire field remains unchanged?

PR #3092, opened 2026-08-18
Challenge workflow file writes for OAuth scope

SamMorrowDrums:
## Summary

Conditionally require the `workflow` OAuth scope when file-write tools target `.github/workflows/**`, while preserving normal `repo`-only writes elsewhere.

## Why

GitHub requires the additional `workflow` scope for workflow-file updates, but statically requiring it would over-scope every file write. The remote server needs a per-call 403 challenge before handlers execute.

Fixes # N/A

## What changed

- Added a shared traversal-safe relative path validator used by `create_or_update_file`, `delete_file`, and every `push_files` entry.
- Added per-call scope resolution from parsed MCP arguments and a pre-handler insufficient-scope challenge for workflow paths.
- Projected top-level `path` through SEP-2243 headers while keeping paths nested in arrays body-parsed.
- Removed unused `codespace` OAuth scope metadata and added regression coverage for traversal, arrays, headers, and challenges.

## MCP impact

- [ ] No tool or API changes
- [x] Tool schema or behavior changed — workflow file writes can now receive a 403 OAuth challenge for the additional `workflow` scope; input schemas are unchanged.
- [ ] New tool added

- "Create `.github/workflows/ci.yml` in owner/repo" — exercises the conditional workflow-scope challenge.
- "Push README.md and `.github/workflows/ci.yml` in one commit" — exercises workflow detection inside the `files` array.
- "Delete `.github/workflows/ci.yml`" — exercises conditional scope resolution for deletion.

## Security / limits

- [x] Auth / permissions considered — only workflow paths add `workflow`; existing token scopes are retained in the challenge.
- [x] Data exposure, filtering, or token/size limits considered — authorization and handlers share path validation so traversal cannot bypass scope detection.

## Tool renaming

## Lint & tests

## Docs

- [x] Not needed — `script/generate-docs` produced no changes.

### DIFF
--- a/pkg/inventory/server_tool.go
+++ b/pkg/inventory/server_tool.go
@@ -18,6 +18,9 @@ import (
 // should define their own typed dependencies struct and type-assert as needed.
 type HandlerFunc func(deps any) mcp.ToolHandler
 
+// ScopeResolver returns additional OAuth scopes required for a specific call.
+type ScopeResolver func(arguments map[string]any) []string
+
 // ToolHandlerMiddleware wraps an MCP tool handler. Middleware is applied from
 // right to left, so the first middleware passed to RegisterFunc executes first.
 type ToolHandlerMiddleware func(next mcp.ToolHandler) mcp.ToolHandler
@@ -101,6 +104,10 @@ type ServerTool struct {
 	// RequiredScopeGroups contains one group of accepted alternatives for each
 	// independently required OAuth scope. Every group must be satisfied.
 	RequiredScopeGroups [][]string
+
+	// ScopeResolver returns scopes that are conditionally required based on the
+	// tool call arguments.
+	ScopeResolver ScopeResolver
 }
 
 // IsReadOnly returns true if this tool is marked as read-only via annotations.
@@ -139,8 +146,8 @@ func (st *ServerTool) RegisterFunc(s *mcp.Server, deps any, middleware ...ToolHa
 		toolCopy.Icons = st.Toolset.Icons()
 	}
 	// Project routing-relevant params to standard MCP-Param-* headers (SEP-2243)
-	// so a remote proxy can read owner/repo from headers instead of re-parsing the
-	// JSON-RPC body. No-op for tools without these params.
+	// so a remote proxy can read top-level parameters from headers instead of
+	// re-parsing the JSON-RPC body. No-op for tools without these params.
 	AnnotateHeaderParams(&toolCopy)
 	s.AddTool(&toolCopy, handler)
 }
@@ -149,7 +156,7 @@ func (st *ServerTool) RegisterFunc(s *mcp.Server, deps any, middleware ...ToolHa
 // header-aware proxy reads, avoiding a second parse of the request body. New
 // routing-relevant params should be added here so projection stays automatic
 // for every tool; the enforcement test in pkg/github guards full coverage.
-var HeaderParams = map[string]string{"owner": "owner", "repo": "repo"}
+var HeaderParams = map[string]string{"owner": "owner", "repo": "repo", "path": "path"}
 
 // AnnotateHeaderParams returns a copy of tool whose routing-relevant input
 // properties (per HeaderParams) carry an "x-mcp-header" annotation, which the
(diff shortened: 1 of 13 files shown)


pkg/inventory/server_tool.go
```diff
@@ -149,7 +156,7 @@ func (st *ServerTool) RegisterFunc(s *mcp.Server, deps any, middleware ...ToolHa
 // header-aware proxy reads, avoiding a second parse of the request body. New
 // routing-relevant params should be added here so projection stays automatic
 // for every tool; the enforcement test in pkg/github guards full coverage.
-var HeaderParams = map[string]string{"owner": "owner", "repo": "repo"}
+var HeaderParams = map[string]string{"owner": "owner", "repo": "repo", "path": "path"}
```
IrynaKulakova: Adding `path` here projects `Mcp-Param-path` for every tool with a top-level `path`, not just the three write tools — `get_file_contents`, `list_commits`, `get_file_blame`, `add_comment_to_pending_review`, `add_pull_request_review_comment`, and `search_code` all pick it up. Two things that give me pause:

- Repo file paths now end up in HTTP headers, and therefore in proxy/access logs. That's noticeably more sensitive than `owner`/`repo`.
- Non-ASCII paths (`docs/café.md`) aren't valid header values, so behaviour depends on how the proxy encodes them.

Could we scope the projection to the tools that actually need it, or confirm the proxy encodes these safely? Also worth updating the comment above — `path` is authorization-relevant rather than routing-relevant, and calling it "routing" will mislead the next person adding a param here.

SamMorrowDrums: Removed `path` from SEP-2243 projection entirely, so only owner/repo remain in `Mcp-Param-*` headers. Conditional workflow scope resolution continues from the already parsed MCP arguments before handler execution (including nested `push_files` paths); added coverage proves a non-ASCII workflow path triggers the challenge without any path header.

> [!NOTE]

pkg/scopes/map.go
```diff
@@ -61,18 +68,52 @@ func GetToolScopeMapFromInventory(inv *inventory.Inventory) ToolScopeMap {
...
+		return t
+	}
+
+	resolved := &ToolScopeInfo{
+		RequiredScopes:      append([]string(nil), t.RequiredScopes...),
+		AcceptedScopes:      append([]string(nil), t.AcceptedScopes...),
+		RequiredScopeGroups: append([][]string(nil), t.RequiredScopeGroups...),
+		ScopeResolver:       t.ScopeResolver,
+	}
+	if len(resolved.RequiredScopeGroups) == 0 {
```
IrynaKulakova, line 100: This fallback silently converts "any of" into "all of".

When `RequiredScopeGroups` is empty, a multi-entry `RequiredScopes` means *any of* — `generate_docs.go` renders exactly that (`Required OAuth Scopes (any of)`), and `HasAcceptedScope` falls through to the `AcceptedScopes` union. Giving each entry its own group flips every alternative into an independently mandatory requirement.

No tool hits this today (all three resolvers sit on tools with a single `repo` scope), but it's a quiet landmine for whoever adds the next resolver. I probed it against a `list_issue_fields`-shaped tool:

```go
base := &ToolScopeInfo{
    RequiredScopes: []string{"repo", "read:org"},
    AcceptedScopes: ExpandScopes(Repo, ReadOrg),
    ScopeResolver:  func(map[string]any) []string { return []string{"workflow"} },
}
base.HasAcceptedScope("repo")  // true — repo alone satisfies any-of

resolved := base.Resolve(map[string]any{})
// groups = [[repo] [admin:org read:org write:org] [workflow]]
resolved.HasAcceptedScope("repo", "workflow")  // false  ← now demands read:org too
resolved.MissingScopes("repo")                 // [read:org workflow]
```

A token holding `repo` + `workflow` would get challenged for `read:org` it never needed, and the challenge would ask for it by name.

Collapsing the base into a single group preserves the any-of semantics:

```go
if len(resolved.RequiredScopeGroups) == 0 && len(resolved.AcceptedScopes) > 0 {
    resolved.RequiredScopeGroups = [][]string{resolved.AcceptedScopes}
}
```

Though note that also breaks the positional `groups[i] -> RequiredScopes[i]` assumption `MissingScopes` relies on, so that pairing probably wants to become explicit rather than positional. Happy with a comment documenting the single-scope precondition instead if you'd rather keep this narrow for now — mainly want the constraint written down somewhere.

PR #3119, opened 2026-08-19
fix(copilot): explain review request denials instead of forwarding a bare 404

### LINKED ISSUE #3027: `request_copilot_review` returns 404 for PR author on cross-fork PR while GitHub UI succeeds
### Describe the bug

The `request_copilot_review` tool returns HTTP 404 when the authenticated user is the author of a public cross-fork pull request but does not have write access to the upstream repository.

The same user can successfully request and re-request a Copilot review using the GitHub web UI on the same pull request.

The MCP server is authenticated as `sean-mcmanus`. Other GitHub MCP read and write operations work correctly. The affected pull request is public: https://github.com/microsoft/vscode/pull/328635. It targets `microsoft/vscode:main` from a branch in the author's `sean-mcmanus/vscode` fork.

The tool implementation calls the standard review-request REST endpoint with `copilot-pull-request-reviewer[bot]`. GitHub's documentation states that pull request authors can request reviews and that Copilot reviews can be requested through this endpoint.

### Affected version

This uses the hosted remote GitHub MCP Server with OAuth, not the Docker image, so the Docker version command is not applicable.

VS Code MCP Gallery metadata reports version `1.6.0`.

Sanitized configuration:

```json
{
  "servers": {
    "io.github.github/github-mcp-server": {
      "type": "http",
      "url": "https://api.githubcopilot.com/mcp/",
      "gallery": "https://api.mcp.github.com",
      "version": "1.6.0"
    }
  },
  "inputs": []
}
```

1. Authenticate the remote GitHub MCP Server using OAuth as a GitHub user.
2. As that user, create a pull request from the user's fork into a public upstream repository where the user does not have write access.
3. Request an initial Copilot review through the GitHub web UI and wait for it to complete.
4. Push another commit to the pull request.
5. Invoke the MCP tool:

   ```text
   request_copilot_review(
     owner="microsoft",
     repo="vscode",
     pullNumber=328635
   )
   ```

6. Observe that the tool returns HTTP 404.
7. Open the same pull request in the GitHub web UI and click **Re-request review** next to Copilot.
8. Observe that the web UI successfully requests another Copilot review.

Authentication can be verified with `get_me`, which returns `sean-mcmanus`.

**Expected**

`request_copilot_review` successfully requests a Copilot review or re-review for a pull request authored by the authenticated user, consistent with the GitHub web UI and the documented REST API behavior.

A successful request should produce a new review-request event and start a Copilot review.

**Actual**

The MCP tool calls the review-request endpoint and receives HTTP 404, even though:

- The pull request exists and is publicly readable.
- The authenticated user is the pull request author.
- Copilot review is enabled for the repository.
- The same authenticated user can request the review through the GitHub web UI.
- Other GitHub MCP operations work.

### Logs

```text
failed to request copilot review:
POST https://api.github.com/repos/microsoft/vscode/pulls/328635/requested_reviewers:
404 Not Found []
```

The failing MCP request is equivalent to requesting:

```json
{
  "reviewers": [
    "copilot-pull-request-reviewer[bot]"
  ]
}
```

### Note
This bug report was created by Copilot (with GPT-5.6 Sol) after failing to use the MCP server and reading the docs at https://github.com/github/github-mcp-server/blob/main/README.md . (Claude Opus 4.8 also hit the issue earlier). It works when I have write access to the repo, e.g. https://github.com/microsoft/vscode-cpptools .

dylanpulver:
## Summary

`request_copilot_review` forwarded GitHub's 404 verbatim, so a caller who simply lacked write access on the target repository got a tool result that read as if the repository or the pull request did not exist. The tool now identifies which cause applies and says so.

## Why

Fixes #3027

The reporter authored a cross-fork pull request into `microsoft/vscode`, requested a Copilot review from the GitHub website successfully, then got `404 Not Found` from the same request through this tool. Nothing in the tool result explained the difference, so the report reasonably concluded the tool was calling the wrong endpoint.

**Root cause.** `POST /repos/{owner}/{repo}/pulls/{pull_number}/requested_reviewers` requires write access to the repository, and authoring the pull request does not grant it ([pull request reviews reference](https://docs.github.com/en/pull-requests/reference/pull-request-reviews#requesting-and-requiring-reviews)). The endpoint documents 403 and 422 as its denial responses ([review requests REST reference](https://docs.github.com/en/rest/pulls/review-requests?apiVersion=2022-11-28#request-reviewers-for-a-pull-request)), but what it actually returns is a 404 carrying no permission signal. Confirmed against the live API with a token holding full `repo` scope on a repository where `permissions.push` is `false`:

```
POST /repos/github/github-mcp-server/pulls/3038/requested_reviewers
HTTP/2.0 404 Not Found
X-Oauth-Scopes: gist, read:org, repo, workflow
```

The reviewer login was deliberately invalid so the call could not succeed, which also shows the permission check runs before reviewer validation. Token scope is not the problem, repository write access is.

**There is no fallback to add.** GraphQL exposes only `requestReviews` and `requestReviewsByLogin`, and both enforce the same gate. Calling `requestReviewsByLogin` as the same non-write user returns `FORBIDDEN: ... does not have the correct permissions to execute RequestReviewsByLogin`, so GraphQL refuses the same actor for the same reason and is only more honest about it. Whatever path the website uses to offer Copilot to a fork author is not exposed as a public API endpoint, so the fix goes to the error surface.

## What changed

- `copilotReviewErrMsg` in `pkg/github/copilot.go` enriches the failure message when the response is 403 or 404. It reads the repository once, on the failure path only, and reports which cause applies: no write access (pointing at the pull request page on the website, which is what the reporter found working), a repository that cannot be read at all, or a repository the caller can write to (pointing at the pull request number and Copilot availability instead).
- Rate limiting also surfaces as 403. `*github.RateLimitError` and `*github.AbuseRateLimitError` short-circuit to the original message, so a rate-limited caller is not told they lack write access and does not pay a second call that would be refused for the same reason.
- The shape follows `dependabotErrMsg` in `pkg/github/dependabot.go`, the existing helper in this repo that appends a permission hint on 403 or 404. Any other status keeps the original message, so unrelated failures are not mislabelled, and the success path is untouched.
- Scoped to `request_copilot_review` on purpose. The other callers of `RequestReviewers` hit the same platform behavior, but the remedy text here is specific to Copilot code review, so a shared helper would not carry over cleanly.

## MCP impact

- [x] No tool or API changes
- [ ] New tool added

The tool schema, name, description and success result are untouched. Only the text of an existing error result changes, so the toolsnap is unchanged.

No tool schema change, so no new prompts. The behavior was exercised through the unit tests below and the underlying API behavior was verified directly against `api.github.com` as shown above.

## Security / limits

The added repository read uses the caller's own client and runs only after a request has already failed, so it grants no access the caller did not have and adds no call to the success path. The message names only the owner and repository the caller supplied, and reports write access as a boolean rather than echoing any response body.

## Tool renaming

## Lint & tests

`./script/lint` reports 0 issues and `./script/test` passes across every package. `./script/generate-docs` produces no diff, confirming the tool surface is unchanged. Six cases were added to `Test_RequestCopilotReview`:

- `pull request author without write access`, the case from the issue: 404 plus a repository whose `permissions.push` is false, asserting the message names the missing write access.
- `forbidden is explained the same way as not found`, so the documented 403 gets the same treatment as the 404 that actually shows up.
- `write access present points at the pull request instead`, asserting the message names pull request 999 rather than blaming permissions.
- `unreadable repository`, where the repository read also fails.
- `server error is not explained as a permission problem`, asserting a 500 keeps the original message and gains no permission text.
- `rate limited forbidden is not explained as a permission problem`, asserting a rate-limited 403 keeps the rate limit message and never reads the repository.

## Docs

- [x] Not needed

The tool description and generated docs are unchanged.

### DIFF
--- a/pkg/github/copilot.go
+++ b/pkg/github/copilot.go
@@ -3,6 +3,7 @@ package github
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"net/http"
@@ -900,7 +901,7 @@ func RequestCopilotReview(t translations.TranslationHelperFunc) inventory.Server
 			)
 			if err != nil {
 				return ghErrors.NewGitHubAPIErrorResponse(ctx,
-					copilotReviewErrMsg(ctx, client, "failed to request copilot review", owner, repo, pullNumber, resp),
+					copilotReviewErrMsg(ctx, client, "failed to request copilot review", owner, repo, pullNumber, resp, err),
 					resp,
 					err,
 				), nil, nil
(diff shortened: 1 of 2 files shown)


SamMorrowDrums (approved):

SamMorrowDrums (approved):
Sanity-checked the failure-path behavior, rate-limit handling, tests, and comment cleanup. CI is green and the branch is current.

abdullah3137867-commits (approved):

PR #3142, opened 2026-08-22
Spell perPage the same way in every paginated tool

karpovantonme:
## Summary

`actions_list` advertised `per_page` while the handler reads pagination through `OptionalPaginationParams`, which looks for `perPage`, so whatever the client sent was dropped and the size fell back to 30. `projects_list` advertises `per_page` too and its handlers do read that name, so it works, but it left one tool spelling pagination differently from the other 30 and that is the gap the actions_list bug slipped through.

## Why

Both schemas now say `perPage`, and a test keeps the rest of the inventory from drifting the same way.

I went over the full inventory rather than grepping for one string: 32 of the 117 tools advertise pagination, 30 of them get it from `WithPagination` / `WithCursorPagination` / `WithUnifiedPagination`, and these two are the only places where the properties are written out by hand. `after_id` and `before_id` on the sub-issue tools are positions rather than pagination, so the test leaves them be.

## What changed

- `pkg/github/actions.go`: `per_page` -> `perPage` in the `actions_list` schema
- `pkg/github/projects.go`: same in the `projects_list` schema, and the three handlers read it through one helper. `per_page` is still read when `perPage` is absent, since the projects tools have advertised that name since September 2025 and clients sending it get the size they ask for today
- `pkg/github/tools_validation_test.go`: `TestAllToolInputSchemasUseCanonicalPaginationNames` walks `AllTools` and rejects case and underscore variants of `page`, `perPage`, `after` and `before`. On main it fails on exactly these two tools
- `pkg/github/projects_test.go`: covers the helper -- `perPage`, legacy `per_page`, both together, neither
- snapshots regenerated with `UPDATE_TOOLSNAPS=true`, README with `script/generate-docs`

The `per_page` fallback in projects is the one judgement call in here. Drop those three lines if you would rather make it a clean break, the test stays green either way

## MCP impact

Two properties renamed. `actions_list` clients sending `per_page` were already being ignored, and `projects_list` clients sending it keep working through the fallback.

- "List the most recent workflow run in github/github-mcp-server" -- the case that was silently returning 30 items. Over stdio against this branch, `perPage: 1` comes back with 1 run and `per_page: 1` with 30, byte for byte the same response as sending no page size at all
- I did not exercise `projects_list` over stdio, no project at hand. `Test_optionalProjectsPerPage` covers the four cases instead

## Security / limits

## Tool renaming

## Lint & tests

## Docs

### DIFF
--- a/README.md
+++ b/README.md
@@ -1092,7 +1092,7 @@ The following sets of tools are available:
   - `method`: The action to perform (string, required)
   - `owner`: The owner (user or organization login). The name is not case sensitive. (string, required)
   - `owner_type`: Owner type (user or org). If not provided, will automatically try both. (string, optional)
-  - `per_page`: Results per page (max 50) (number, optional)
+  - `perPage`: Results per page (max 50) (number, optional)
   - `project_number`: The project's number. Required for 'list_project_fields', 'list_project_items', 'list_project_views', and 'list_project_status_updates' methods. (number, optional)
   - `query`: Filter/query string. For list_projects: filter by title text and state (e.g. "roadmap is:open"). For list_project_items: advanced filtering using GitHub's project filtering syntax. (string, optional)
 
--- a/README.md
+++ b/README.md
@@ -631,7 +631,7 @@ The following sets of tools are available:
   - `method`: The action to perform (string, required)
   - `owner`: Repository owner (string, required)
   - `page`: Page number for pagination (default: 1) (number, optional)
-  - `per_page`: Results per page for pagination (default: 30, max: 100) (number, optional)
+  - `perPage`: Results per page for pagination (default: 30, max: 100) (number, optional)
   - `repo`: Repository name (string, required)
   - `resource_id`: The unique identifier of the resource. This will vary based on the "method" provided, so ensure you provide the correct ID:
     - Do not provide any resource ID for 'list_workflows' method.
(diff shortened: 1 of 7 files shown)


SamMorrowDrums:
Can you make sure all schema advertised per_page entries are handled, I think there might be some others, and we should be totally consistent to reduce chances of bugs like this occuring.

karpovantonme:
Went through the whole inventory rather than grepping for the string: 32 of the 117 tools advertise pagination, 30 of them get it from `WithPagination` / `WithCursorPagination` / `WithUnifiedPagination`, and `actions_list` and `projects_list` are the only two that write the properties out by hand. `projects_list` does read `per_page` back under the same name so it works, but it was the other tool spelling pagination its own way, so it is in this PR now -- schema says `perPage`, with `per_page` still read when `perPage` is absent, since the projects tools have advertised that name since Sept 2025.

For the "reduce chances of bugs like this occurring" part I put the check next to your combinator guard in `tools_validation_test.go`: it walks `AllTools` and rejects case and underscore variants of `page`, `perPage`, `after` and `before`. On main it fails on exactly those two tools. Having the inventory-wide file already there meant the guard was a small addition rather than new scaffolding.

The `per_page` fallback in projects is the one judgement call in here -- drop those three lines if you would rather have a clean break, the test is green either way

SamMorrowDrums (approved):

PR #3146, opened 2026-08-24
Enable feature flags via URL query parameter (?features=) for headerless hosted connections

### LINKED ISSUE #3145: Feature flags are header-only, making flagged tools unreachable on managed/hosted MCP connections
Toolsets can be selected two ways: by URL (`https://api.githubcopilot.com/mcp/x/issues`, with `/readonly` as a suffix) or by the `X-MCP-Toolsets` header. Feature flags have only one — the `X-MCP-Features` header.

That makes every flagged tool unreachable whenever the MCP connection is operated by someone other than the end user: a hosted IDE, an agent platform, any client that composes the URL on your behalf.

Concrete case: Claude Code on the web. Its GitHub MCP tools are provisioned by the harness — `claude mcp list` reports no configurable server at all, while `mcp__github__*` tools are present — so there is no header I can set on that connection. `issue_dependency_read` / `issue_dependency_write` are therefore unreachable, even though `issue_dependencies` is in `AllowedFeatureFlags` and the tools are implemented and shipped.

The available workaround is to declare a *second* GitHub MCP server pointing at the same endpoint, with its own PAT, purely to set one header. That's a duplicate connection and an extra long-lived credential to reach a tool that already exists.

### Proposed solution

Mirror the existing toolset pattern for flags. Either shape would work:

- path segment — `https://api.githubcopilot.com/mcp/x/issues/f/issue_dependencies`
- query parameter — `https://api.githubcopilot.com/mcp/x/issues?features=issue_dependencies`

Validation would be unchanged. `ResolveFeatureFlags` already filters user-supplied flags against `AllowedFeatureFlags` and silently drops unknown ones, so a URL-supplied flag is no more privileged than a header-supplied one — it's the same input arriving through a different channel.

This keeps the gate and its rationale intact. Flagged tools stay out of the default surface, so the fixed tool-schema cost doesn't change; it only widens how a user can opt in.

A less preferred alternative would be graduating `issue_dependencies` into the default `issues` toolset now that issue dependencies are GA in the REST API. I'd rather not argue for that — the flag exists specifically to hold down default context cost, and URL-addressability solves the problem without giving that up.

### Example prompts or workflows (for tools/toolsets only)

Split frontend/backend repo pairs are the main driver here, since `related_owner` / `related_repo` make dependencies cross-repo:

- "What's blocking BE-99?" — where some blockers live in the sibling frontend repo
- "Mark FE-66 as blocked by BE-95" — a cross-repo write
- "List everything the children of issue #195 are blocked by, and flag any blocker that's already closed"
- "Before I start BE-100, show me its unresolved blockers"
- "Order these 27 sub-issues into a build sequence using their dependency graph"

### Additional context

- Flag is `FeatureFlagIssueDependencies = "issue_dependencies"` in `pkg/github/feature_flags.go`, present in both `AllowedFeatureFlags` and `InsidersFeatureFlags`. Tools added in #2751.
- `/insiders` does enable it, but also turns on `csv_output`, which changes the response format of every `list_*` tool. That's too broad a side effect to accept in order to reach one tool.
- Client is Claude Code on the web (Anthropic-hosted). Worth noting that "just use the REST API instead" isn't a workaround in that environment: its GitHub REST proxy silently omits cross-repo entries from response bodies — HTTP 200, no `Link` header, entries simply absent. I verified this by comparing `client.SubIssue.ListByIssue` results via the MCP server (27 sub-issues, 8 of them cross-repo) against the identical REST endpoint through the proxy (19, same-repo only). So for cross-repo dependency reads specifically, the MCP tools aren't a convenience over REST — they're the only path that returns complete data.

CAOShurong:
Closes #3145

## Summary

Allow clients that cannot set custom headers to enable approved feature flags with a `features` URL query parameter. `X-MCP-Features` remains authoritative whenever the header is present; the header and query channels are never combined.

## Why

Hosted IDEs, agent platforms, and harnesses often compose the MCP server URL but cannot attach custom headers. Without a URL channel, those clients cannot opt into flagged tools.

Example:

```text
https://api.githubcopilot.com/mcp/x/issues?features=issue_dependencies
```

## What changed

- `pkg/http/middleware/request_config.go` selects feature flags by channel presence: the header wins when present (including empty, whitespace-only, or unknown values); otherwise `?features=` is used. The two channels never merge.
- `pkg/http/oauth/oauth.go` preserves the request's exact raw query in both the advertised OAuth protected-resource metadata URL and the returned `metadata.resource`, so query-bearing MCP resource identifiers remain consistent for RFC 9728 clients.
- Tests cover eight header/query combinations, query preservation with and without a configured base URL, and an end-to-end OAuth challenge-to-metadata round trip.
- `docs/feature-flags.md` and `docs/server-configuration.md` document the query channel and header precedence.

Unknown flags are still filtered by `ResolveFeatureFlags` against `AllowedFeatureFlags`; the query parameter does not widen the allowed feature set. Flagged tools remain absent from the default surface.

## MCP impact

- [ ] No tool or API changes
- [x] Tool schema or behavior changed — feature activation gains an additional request channel; tool schemas are unchanged.
- [ ] New tool added

## Prompts tested

Not applicable; no tool schema or tool implementation changed.

## Security / limits

- [x] Auth / permissions considered — OAuth resource metadata retains the full query-bearing resource identifier.
- [x] Data exposure, filtering, or token/size limits considered — URL-supplied flags use the same allowlist as header-supplied flags and expose no new flag values.

## Tool renaming

## Lint & tests

Verified on exact head `0c3dc680f33a19fae472e92f75efd67ea2f2bd3a`:

- `go test ./pkg/http/... -count=1` — pass
- `go test ./pkg/github -run TestResolveFeatureFlags` — pass
- `go vet ./pkg/http/...` — pass
- `go build ./...` — pass
- `git diff --check` — pass

The repository-wide `script/lint` and `script/test` wrappers were not run on this final head.

## Docs

- [x] Updated (`docs/feature-flags.md`, `docs/server-configuration.md`)

## AI assistance disclosure

Codex assisted with implementation, verification, and this status summary. I reviewed the final diff and the results above.

### DIFF
--- a/pkg/http/middleware/request_config.go
+++ b/pkg/http/middleware/request_config.go
@@ -18,9 +18,9 @@ const queryParamFeatures = "features"
 
 // WithRequestConfig is a middleware that extracts MCP-related headers and sets them in the request context.
 // This includes readonly mode, toolsets, tools, lockdown mode, insiders mode, and feature flags.
-// Feature flags may also arrive via the `features` URL query parameter; when
-// both are present the query parameter wins, matching how the toolset path
-// segments take precedence over their headers.
+// Feature flags may also arrive via the `features` URL query parameter. When
+// both are present, the X-MCP-Features header takes precedence and the two
+// channels are never combined.
 func WithRequestConfig(next http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		ctx := r.Context()
@@ -57,18 +57,26 @@ func WithRequestConfig(next http.Handler) http.Handler {
 
 		// Feature flags: presence-based selection. The URL query parameter and
 		// the X-MCP-Features header are separate channels — whichever is
-		// present is used as-is, and they are never combined. When both are
-		// present the query parameter wins, so a client composing the server
-		// URL can always express its intent even when it cannot control
-		// headers. Unknown flags are dropped later by ResolveFeatureFlags
-		// against AllowedFeatureFlags, so neither channel is privileged.
+		// present is used as-is, and they are never combined. Header presence
+		// takes precedence even when its value is empty or contains only
+		// unknown flags, so a query parameter cannot override an explicit
+		// header. Unknown flags are dropped later by ResolveFeatureFlags
+		// against AllowedFeatureFlags.
 		queryFeatures, hasQuery := r.URL.Query()[queryParamFeatures]
-		headerFeatures := r.Header.Get(headers.MCPFeaturesHeader)
+		headerFeatures, hasHeader := r.Header[http.CanonicalHeaderKey(headers.MCPFeaturesHeader)]
 		switch {
-		case hasQuery && strings.TrimSpace(queryFeatures[0]) != "":
-			ctx = ghcontext.WithHeaderFeatures(ctx, headers.ParseCommaSeparated(queryFeatures[0]))
-		case headerFeatures != "":
-			ctx = ghcontext.WithHeaderFeatures(ctx, headers.ParseCommaSeparated(headerFeatures))
+		case hasHeader:
+			headerValue := ""
+			if len(headerFeatures) > 0 {
+				headerValue = headerFeatures[0]
+			}
+			ctx = ghcontext.WithHeaderFeatures(ctx, headers.ParseCommaSeparated(headerValue))
+		case hasQuery:
+			queryValue := ""
+			if len(queryFeatures) > 0 {
+				queryValue = queryFeatures[0]
+			}
+			ctx = ghcontext.WithHeaderFeatures(ctx, headers.ParseCommaSeparated(queryValue))
 		}
 
 		next.ServeHTTP(w, r.WithContext(ctx))
--- a/pkg/http/middleware/request_config.go
+++ b/pkg/http/middleware/request_config.go
@@ -9,8 +9,18 @@ import (
 	"github.com/github/github-mcp-server/pkg/http/headers"
 )
 
+// queryParamFeatures is the URL query parameter that carries feature flags,
+// mirroring the X-MCP-Features header. It exists so clients that cannot set
+// custom headers on the MCP connection — hosted IDEs, agent platforms, or
+// harnesses that compose the server URL on the user's behalf (see #3145) —
+// can still opt into flagged tools.
+const queryParamFeatures = "features"
+
 // WithRequestConfig is a middleware that extracts MCP-related headers and sets them in the request context.
 // This includes readonly mode, toolsets, tools, lockdown mode, insiders mode, and feature flags.
+// Feature flags may also arrive via the `features` URL query parameter; when
+// both are present the query parameter wins, matching how the toolset path
+// segments take precedence over their headers.
 func WithRequestConfig(next http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		ctx := r.Context()
@@ -45,9 +55,20 @@ func WithRequestConfig(next http.Handler) http.Handler {
 			ctx = ghcontext.WithInsidersMode(ctx, true)
 		}
 
-		// Feature flags
-		if features := headers.ParseCommaSeparated(r.Header.Get(headers.MCPFeaturesHeader)); len(features) > 0 {
-			ctx = ghcontext.WithHeaderFeatures(ctx, features)
+		// Feature flags: presence-based selection. The URL query parameter and
+		// the X-MCP-Features header are separate channels — whichever is
+		// present is used as-is, and they are never combined. When both are
+		// present the query parameter wins, so a client composing the server
+		// URL can always express its intent even when it cannot control
+		// headers. Unknown flags are dropped later by ResolveFeatureFlags
+		// against AllowedFeatureFlags, so neither channel is privileged.
+		queryFeatures, hasQuery := r.URL.Query()[queryParamFeatures]
+		headerFeatures := r.Header.Get(headers.MCPFeaturesHeader)
+		switch {
+		case hasQuery && strings.TrimSpace(queryFeatures[0]) != "":
+			ctx = ghcontext.WithHeaderFeatures(ctx, headers.ParseCommaSeparated(queryFeatures[0]))
+		case headerFeatures != "":
+			ctx = ghcontext.WithHeaderFeatures(ctx, headers.ParseCommaSeparated(headerFeatures))
 		}
 
 		next.ServeHTTP(w, r.WithContext(ctx))
--- a/pkg/http/middleware/request_config.go
+++ b/pkg/http/middleware/request_config.go
@@ -9,20 +9,15 @@ import (
 	"github.com/github/github-mcp-server/pkg/http/headers"
 )
 
-// queryParamFeatures is the URL query parameter that carries feature flags,
-// mirroring the X-MCP-Features header. It exists so clients that cannot set
-// custom headers on the MCP connection — hosted IDEs, agent platforms, or
-// harnesses that compose the server URL on the user's behalf (see #3145) —
-// can still opt into flagged tools.
 const queryParamFeatures = "features"
 
 // WithRequestConfig is a middleware that extracts MCP-related headers and sets them in the request context.
 // This includes readonly mode, toolsets, tools, lockdown mode, insiders mode, and feature flags.
-// Feature flags may also arrive via the `features` URL query parameter. When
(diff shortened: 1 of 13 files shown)


pkg/http/middleware/request_config.go
```diff
@@ -45,8 +55,16 @@ func WithRequestConfig(next http.Handler) http.Handler {
...
 		}
 
-		// Feature flags
-		if features := headers.ParseCommaSeparated(r.Header.Get(headers.MCPFeaturesHeader)); len(features) > 0 {
+		// Feature flags: the URL query parameter wins over the header, so a
+		// client composing the server URL can always express its intent even
+		// when it cannot control headers. Unknown flags are dropped later by
+		// ResolveFeatureFlags against AllowedFeatureFlags, so this channel is
+		// no more privileged than the header — it is the same allowlist.
+		features := headers.ParseCommaSeparated(r.URL.Query().Get(queryParamFeatures))
```
SamMorrowDrums: Headers need full precedence here. Please check whether `X-MCP-Features` is present before reading the query parameter; if it is present, parse only that header and never calculate or fall back to query features. Presence—not parsed length—is important, so an explicitly empty, whitespace-only, or unknown-valued header must still suppress `?features=`. Please invert the precedence and add those cases to this test matrix.

SamMorrowDrums (changes requested):
The query-string option makes sense for clients that cannot set headers, but I do not think this is release-ready yet because it breaks OAuth protected-resource metadata discovery for query-bearing MCP URLs.

For a server URL such as `https://api.githubcopilot.com/mcp/x/issues?features=issue_dependencies`, go-sdk v1.7.0 treats the full URL as the resource identifier and validates `metadata.resource` with exact string equality. Its path-based and root discovery candidates also preserve the query. Today `BuildResourceMetadataURL` and `AuthHandler.buildResourceURL` omit `r.URL.RawQuery`, so the challenge and returned metadata lose `?features=...`; standards-compliant clients can reject the metadata as belonging to a different resource.

This should be fixable here: preserve the feature query consistently in both the advertised `resource_metadata` URL and the metadata document’s `resource`, then extend `TestOAuthChallengeMetadataRouteContracts` (or an equivalent client round-trip test) to cover a query-bearing MCP URL and both supported discovery routes. Please also make the header-selection rule below presence-based so routing/query and header features are never combined or used as fallbacks for one another.

CAOShurong:
Rebased onto current `main` (`8898db9`) and pushed both requested changes as `81d4230`:

**1. Query preservation in OAuth protected-resource metadata.** `BuildResourceMetadataURL` and `AuthHandler.buildResourceURL` now append `r.URL.RawQuery`, via a shared `AppendQuery` helper, so the advertised `resource_metadata` URL and the metadata document's `resource` both carry the exact same query as the MCP server URL — satisfying go-sdk's exact string-equality check for query-bearing URLs like `/mcp/x/issues?features=issue_dependencies`.

**2. Presence-based feature selection.** The features query parameter and `X-MCP-Features` are now separate channels that are never combined or used as fallbacks for one another (except that an *empty* `?features=` value falls back to the header). Query wins when both are present.

**Tests added:**
- `TestOAuthChallengeMetadataRouteContracts/x/repos?features=issue_dependencies` — full round-trip: challenge's `resource_metadata` URL exact-matches, then GET-ing it returns metadata whose `resource` equals `https://mcp.example.com/mcp/x/repos?features=issue_dependencies`.
- `TestWithRequestConfigFeatureSelection` — all four channel combinations plus empty-value fallback.
- `TestBuildResourceMetadataURL` gains two query-preservation cases.

Verification on this head: `go build ./...` exit 0, `go vet ./pkg/http/...` clean, full `go test ./pkg/http/... -count=1` green (handler/middleware/oauth/transport all pass), gofmt clean. Docs updated in `docs/feature-flags.md` and `docs/server-configuration.md`.

CAOShurong (commented):
# P2 review — github/github-mcp-server#3146 (round 2, exact head `81d4230`)

## Verification performed
- Worktree at PR head `81d42309f642d043332f38426d87bf7d1fba29f1`
  ("fix(http): preserve feature query across OAuth resource metadata",
  parent `8898db9` = current origin/main tip).
- `go test ./pkg/http/...` → all packages ok (pkg/http, headers,
  middleware, oauth, transport) on go1.26.7 windows-amd64.
- Targeted: TestOAuthChallengeMetadataRouteContracts including the new
  `/x/repos?features=issue_dependencies` round-trip subtest — PASS.
  TestBuildResourceMetadataURL both new query-preservation cases — PASS.
  TestWithRequestConfigFeatureSelection all five matrix rows — PASS.
- `go vet ./pkg/http/...` clean; gofmt clean on all touched files
  (repo-wide gofmt noise is pre-existing CRLF checkout artifacts).

## Review of the new push (addressing CHANGES_REQUESTED)

The reviewer's two asks were:

1. **Preserve `?features=...` in OAuth resource-metadata discovery** —
   correctly implemented. `AppendQuery` now threads `r.URL.RawQuery` into
   both `buildResourceURL` and `BuildResourceMetadataURL`, so the challenge
   URL and the served metadata document's `resource` field both carry the
   exact full string that go-sdk v1.7.0 validates by strict equality. The
   new round-trip subtest in `TestOAuthChallengeMetadataRouteContracts`
   asserts precisely this contract end-to-end (challenge header → GET the
   advertised metadata URL → `metadata.resource == baseURL + "/mcp" +
   queryPath`). That is exactly the client-side failure mode described in
   the review, and it is now covered by a regression test rather than only
   unit-level assertions.

2. **Presence-based precedence between header and query** — the new
   implementation selects the channel by presence (`hasQuery &&
   non-empty-first-value` vs `headerFeatures != ""`) and never combines or
   falls back from one to the other when both are present; query wins.
   The empty-query-value-falls-back-to-header case is covered in the test
   matrix.

One remaining nuance worth flagging (non-blocking): under "query wins",
an explicitly *empty* `?features=` currently falls back to the header —
the test matrix documents this as intended. If maintainers want the
strictest presence semantics ("a present-but-empty channel suppresses the
other"), that is a one-line change plus one more matrix row, but the
current behavior matches how the toolset path segments behave and is
defensible as-is.

Docs (`feature-flags.md`, `server-configuration.md`) updated consistently,
including an explicit statement of the precedence rule. DCO/CI state
should be re-checked after this push since it was force-pushed today.

Verdict: the changes_requested items are addressed with tests; ready for
re-review from my side.

baseten:
@SamMorrowDrums @CAOShurong curious if all requested changes have been addressed? This appears to have stalled for now

CAOShurong:
Yes — both requested changes are addressed on `81d4230`:

- OAuth resource-metadata discovery now preserves the feature query in both the advertised URL and returned `metadata.resource`, with an end-to-end regression test.
- Header/query feature selection is presence-based, never combines channels, and gives the query precedence when both are non-empty.

All 13 public checks on that head are green, and my local targeted/full `pkg/http` tests and vet passed. @SamMorrowDrums, when convenient, could you re-review the updated head? The branch is currently behind `main` but still mergeable; I can rebase if you prefer.

AI assistance disclosure: Codex helped verify and summarize the current status; I checked the cited head and results.

CAOShurong:
Rebased the branch onto current `main` (`febc3293`); the new head is `c5e3f6fe`. `git range-diff` reports the feature patch as unchanged from `81d42309`.

Fresh verification on the rebased head:

- `go test ./pkg/http/... -count=1` — pass
- `go vet ./pkg/http/...` — pass
- `go build ./...` — pass

The branch is now current and ready for re-review.

AI assistance disclosure: Codex assisted with the rebase and verification; I reviewed the resulting range-diff and test results.

CAOShurong:
Addressed the requested precedence behavior in `0c3dc680`:

- `X-MCP-Features` now wins whenever the header is present, including empty, whitespace-only, or unknown values; the header and query channels are never merged.
- Added coverage for query-only, header-only, header-over-query, empty/whitespace/unknown headers, empty query with a header, and no channel.
- Updated the feature-flag documentation to match.

Validation: `go test ./pkg/http/...`, `go test ./pkg/github -run TestResolveFeatureFlags`, and `go vet ./pkg/http/middleware` pass. Please re-review.

SamMorrowDrums (approved):
The requested header-precedence and query-preserving OAuth changes are implemented with focused coverage; current CI and Copilot review are clean.

pkg/http/middleware/request_config.go
SamMorrowDrums: Header presence now wins before parsing, so empty, whitespace-only, and unknown `X-MCP-Features` values suppress `?features=` without combining channels, with matrix coverage.

> [!NOTE]

PR #3166, opened 2026-08-27
Refactor feature flags to use functional availability rules

SamMorrowDrums:
## Summary

- replace fixed enable/disable gates with typed functional `FeatureRule` predicates
- evaluate rules lazily after request narrowing with normal Go short-circuiting
- memoize each reached flag once per request across tools, resources, prompts, and handlers
- keep `AllowedFeatureFlags` as the separate user-controllable allowlist
- validate rule declarations and duplicate-name exclusivity
- add real-inventory performance benchmarks

## Breaking change and migration

The following exported fields are removed:

- `inventory.ServerTool.FeatureFlagEnable`
- `inventory.ServerTool.FeatureFlagEnableAll`
- `inventory.ServerTool.FeatureFlagDisable`
- `inventory.ServerResourceTemplate.FeatureFlagEnable`
- `inventory.ServerResourceTemplate.FeatureFlagDisable`
- `inventory.ServerPrompt.FeatureFlagEnable`
- `inventory.ServerPrompt.FeatureFlagDisable`

Replace an enable gate with:

```go
flag := inventory.FeatureFlag("my_feature")
tool.FeatureRule = inventory.NewFeatureRule(
    []inventory.FeatureFlag{flag},
    func(featureAsBool inventory.FeatureResolver) bool {
        return featureAsBool(flag)
    },
)
```

Replace a disable gate with the same declaration and `return !featureAsBool(flag)`. Multi-flag AND, OR, and UNLESS behavior belongs directly in the predicate; list every referenced flag in the declaration.

`inventory.FeatureFlagChecker` remains `func(context.Context, string) (bool, error)`, and `github.ToolDependencies.IsFeatureEnabled` continues to accept string flag names.

## Performance

A shared 139-tool/5-resource/2-prompt benchmark compares current `main` with this branch under all-false, mixed, and all-true distributions:

- build performs zero flag checks
- unflagged tool calls perform zero checks
- gated and UI tool calls perform one check
- tools/list checks each reached unique OSS availability flag once: 41/43 repeated checks become 7
- end-to-end build+narrow+register: -9.0% to -3.9% median CPU in the final run, +0.4% bytes, +0.6–0.8% allocations
- a longer all-false rerun measured -0.2% CPU, +0.34% bytes, and +0.77% allocations

## Validation

- targeted feature/query/registration tests under `-race -count=20`
- representative benchmark under three flag distributions
- `script/lint`
- `script/test`
- `script/generate-docs`


pkg/github/feature_flags.go
```diff
@@ -1,47 +1,51 @@
...
 // FeatureFlagThreadResolutionReason exposes resolution reasons for Copilot review threads.
-const FeatureFlagThreadResolutionReason = "thread_resolution_reason"
+const FeatureFlagThreadResolutionReason inventory.FeatureFlag = "thread_resolution_reason"
 
 // AllowedFeatureFlags is the allowlist of feature flags that can be enabled
 // by users via --features CLI flag or X-MCP-Features HTTP header.
 // Only flags in this list are accepted; unknown flags are silently ignored.
 // This is the single source of truth for which flags are user-controllable.
-var AllowedFeatureFlags = []string{
+var AllowedFeatureFlags = []inventory.FeatureFlag{
```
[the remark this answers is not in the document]
SamMorrowDrums: Fixed by preserving the existing string-based `AllowedFeatureFlags`, `InsidersFeatureFlags`, and `ResolveFeatureFlags` APIs. The new public `inventory.FeatureFlag` type is limited to the checker/rule boundary, and the temporary conversion API was removed.

> [!NOTE]

pkg/http/handler.go
```diff
@@ -214,6 +216,7 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	if methodInfo, ok := ghcontext.MCPMethod(r.Context()); ok && methodInfo != nil {
 		invToUse = inv.ForMCPRequest(methodInfo.Method, methodInfo.ItemName)
 	}
+	r = r.WithContext(inventory.WithResolvedFeatures(r.Context(), h.featureChecker, invToUse.RequiredFeatures()))
```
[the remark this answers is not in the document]
SamMorrowDrums: Fixed by moving pre-resolution behind `Inventory.WithResolvedFeatures`. The HTTP handler now asks the selected inventory to seed state with its own checker, so a custom inventory factory cannot be overridden by the handler option.

> [!NOTE]

pkg/inventory/features.go
```diff
@@ -0,0 +1,189 @@
...
+		if _, ok := r.featureSet[feature]; !ok {
+			undeclared = feature
+			return false
+		}
+		return featureAsBool(feature)
+	})
+	if undeclared != "" {
+		fmt.Fprintf(os.Stderr, "Feature rule used undeclared feature %q\n", undeclared)
+		return false
+	}
```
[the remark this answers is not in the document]
SamMorrowDrums: Fixed by tracking undeclared access with a separate boolean, so the empty feature name now fails closed. Added a regression test for a negated empty-feature lookup.

> [!NOTE]

pkg/inventory/features.go
```diff
@@ -0,0 +1,189 @@
...
+	if feature == "" {
+		return false
+	}
+	if state, _ := ctx.Value(featureStateContextKey{}).(*featureState); state != nil {
+		return state.enabled(ctx, feature)
+	}
+	if checker == nil {
+		return false
+	}
+	return newFeatureState(checker).enabled(ctx, feature)
```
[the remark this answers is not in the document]
SamMorrowDrums: Fixed by installing inventory-owned feature state as receiving middleware in `NewMCPServer`. Each stdio method call now gets persistent request state, so repeated handler checks reuse the cached value. Added a middleware caching test.

> [!NOTE]

pkg/inventory/features.go
```diff
@@ -0,0 +1,191 @@
...
+	}
+
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	if enabled, ok := s.values[feature]; ok {
+		return enabled
+	}
+
+	enabled, err := s.checker(ctx, feature)
```
IrynaKulakova: `s.mu` is held across the `s.checker(ctx, feature)` call, which makes this non-reentrant: a checker that consults another flag (via `ResolveFeature` or `deps.IsFeatureEnabled`) on the same context deadlocks the request. Not an error — a hard hang. I reproduced it with a probe test against this branch:

```go
checker := func(c context.Context, flag FeatureFlag) (bool, error) {
	if flag == "meta" {
		return ResolveFeature(c, nil, "base"), nil
	}
	return true, nil
}
ctx := WithResolvedFeatures(context.Background(), checker, nil)
ResolveFeature(ctx, nil, "meta") // never returns
```

The same lock also serializes every flag lookup for the request, which matters if the remote server's checker does I/O. Could we resolve outside the lock (double-check on insert) or use a per-flag `sync.Once`/singleflight?

pkg/inventory/features.go
```diff
@@ -0,0 +1,191 @@
...
+	slices.Sort(features)
+	for _, feature := range features {
+		state.enabled(ctx, feature)
+	}
+	return ctx
+}
+
+// ResolveFeature returns a feature value from request-owned resolution state.
+// Features not resolved up front are resolved lazily and cached.
+func ResolveFeature(ctx context.Context, checker FeatureFlagChecker, feature FeatureFlag) bool {
```
IrynaKulakova: The `checker` argument is silently ignored whenever context state already exists, which has two consequences worth pinning down:

1. `BaseDeps`/`RequestDeps` pass their own `featureChecker`, but the inventory's checker wins if it installed state first. A consumer wiring different checkers for inventory vs. deps gets whichever ran first, with no signal.
2. Behavior change: `IsFeatureEnabled` with a **nil** deps checker previously always returned `false`. It now returns the inventory-resolved value. The doc comment on `BaseDeps.IsFeatureEnabled` still says "Returns false if the feature checker is nil", so at minimum that's stale.

Either drop the parameter and make context state authoritative, or document the precedence explicitly here.

pkg/inventory/features.go
```diff
@@ -0,0 +1,191 @@
...
+	usedUndeclared := false
+	enabled := r.predicate(func(feature FeatureFlag) bool {
+		if _, ok := r.featureSet[feature]; !ok {
+			undeclared = feature
+			usedUndeclared = true
+			return false
+		}
+		return featureAsBool(feature)
+	})
+	if usedUndeclared {
```
IrynaKulakova, line 118: Undeclared-flag detection only fires if the predicate actually evaluates that branch. Because predicates short-circuit, a rule that reads an undeclared flag in a rarely-taken branch works fine until it doesn't, and then silently drops the tool from the surface with one line on stderr — hard to spot in production.

Declared sets here are 1–2 flags, so `NewFeatureRule` could probe the predicate over all 2^n combinations at construction time and fail loudly on an undeclared read. Failing that, a package-level guard test over the real inventory would catch it at CI time rather than at request time.

pkg/inventory/registry.go
```diff
@@ -188,13 +188,83 @@ func (r *Inventory) ToolsetDescriptions() map[ToolsetID]string {
...
+	rules := make([]FeatureRule, 0, len(r.prompts))
+	for i := range r.prompts {
+		rules = append(rules, r.prompts[i].FeatureRule)
+	}
+	return collectFeatures(rules...)
+}
+
+// RequiredFeatures returns the deduplicated feature flags used to expose the
+// inventory's current tools, resources, and prompts.
+func (r *Inventory) RequiredFeatures() []FeatureFlag {
```
IrynaKulakova, line 207: This walk gets repeated several times per request: `handler.ServeHTTP` calls `WithResolvedFeatures` (→ `RequiredFeatures` + `usesMCPAppsMetadata`), then `ToolsForRegistration` runs both again, then `AvailableTools` runs `requiredToolFeatures` a third time. Each pass allocates a rule slice plus a dedup map over the whole inventory.

Since tools/resources/prompts are fixed after `Build()` (and after `ForMCPRequest` narrowing), could the required-feature set and the `usesMCPAppsMetadata` bool be computed once and stored on `Inventory`?

pkg/http/handler.go
```diff
@@ -214,6 +214,7 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	if methodInfo, ok := ghcontext.MCPMethod(r.Context()); ok && methodInfo != nil {
 		invToUse = inv.ForMCPRequest(methodInfo.Method, methodInfo.ItemName)
 	}
+	r = r.WithContext(invToUse.WithResolvedFeatures(r.Context()))
```
IrynaKulakova: This resolves and **caches** the flag values before the MCP receiving middlewares run (`injectFeatureStateMiddleware`, `InjectDepsMiddleware`). Fine for this repo's header-based checker, but a checker that reads actor/session data populated later in the chain would see a thinner context here, and the resulting value then sticks for the rest of the request.

The `RegisterTools` doc comment already warns about exactly this class of bug ("HTTP feature checkers that read insiders mode or user identity from ctx would otherwise see context.Background()"). Worth confirming against the remote server's checker before this lands — if it needs the enriched context, the resolution point should move after the middlewares.

IrynaKulakova (commented):
A few notes on the new feature-rule machinery. Overall direction looks right — replacing the three annotation fields with one declared predicate is a clear improvement, and the fail-closed semantics of the old `featureFlagAllowed` are preserved. Comments below are about the resolution state, not the rule model itself.

FuzzysTodd (approved):
tijr

pkg/inventory/features.go
SamMorrowDrums: `NewFeatureRule` now exhaustively validates declared-flag assignments at construction, so undeclared lookups in short-circuited branches fail loudly with regression coverage.

> [!NOTE]

pkg/http/handler.go
SamMorrowDrums: Confirmed the remote request context is installed before the OSS handler and added ordering coverage while keeping handler-only checks on the live MCP context.

> [!NOTE]

pkg/inventory/registry.go
SamMorrowDrums: The inventory now caches required feature metadata at build time and recomputes it once after request narrowing, removing the repeated per-request walks.

> [!NOTE]

pkg/inventory/features.go
SamMorrowDrums: Request-context feature state is now explicitly authoritative, while dependency checkers are documented and tested as fallback-only for direct handler invocation.

> [!NOTE]

pkg/inventory/features.go
SamMorrowDrums: Feature resolution now uses per-flag single-flight without holding the map lock across checker calls, with nested, cycle, and concurrent regression coverage.

> [!NOTE]

PR #3167, opened 2026-08-27
fix(http): allow projected MCP headers in preflights

SamMorrowDrums:
## Summary

- extend the browser CORS allowlist from #3147 with MCP 2026-07-28 `Mcp-Method` and `Mcp-Name` request headers plus the current `Mcp-Param-owner` and `Mcp-Param-repo` projections
- inspect every `Access-Control-Request-Headers` value on `OPTIONS` and reflect only valid, non-empty `Mcp-Param-*` field names
- deduplicate projected names case-insensitively, sort them deterministically, and emit canonical header casing without exposing request parameters as response headers
- preserve the existing wildcard-origin/no-credentials policy, fixed allowlist, exposed headers, downstream headers, and non-`OPTIONS` behavior

## Exact projected-header contract

CORS has no prefix wildcard, so preflights validate each requested name with `httpguts.ValidHeaderFieldName`. Matching is case-insensitive against the exact `Mcp-Param-` prefix and requires a non-empty suffix. Fixed owner/repo entries remain first; future names are case-insensitively deduplicated, sorted, and emitted with `http.CanonicalHeaderKey`. Arbitrary headers, bare `Mcp-Param-`, malformed/control-character names, and lookalike prefixes are omitted.

`Mcp-Param-*` headers are request-only and are not added to `Access-Control-Expose-Headers`; `Mcp-Session-Id` and `WWW-Authenticate` remain exposed.

## Cross-repository dependencies

- Follow-up to merged #3147 and the v1.11.0 direct/self-hosted HTTP contract.
- `github/github-mcp-server-remote#955` must consume this commit or a release containing it for direct remote traffic; its stable dependency blocker remains until this change is merged and released.
- `github/copilot-api#37229` separately owns matching hosted-edge preflight behavior.

## Validation

- `script/lint`
- `script/test`


pkg/http/middleware/cors.go
```diff
@@ -2,42 +2,90 @@ package middleware
...
+				continue
+			}
+			if _, ok := seen[key]; ok {
+				continue
+			}
+
+			seen[key] = struct{}{}
+			projected = append(projected, key)
+		}
+	}
```
IrynaKulakova, line 90: There's no cap on how many projected names we reflect here. Within Go's 1MB request header limit, a client can send thousands of distinct `Mcp-Param-x` names on a preflight and get every one of them echoed back, growing both `seen` and the response header. That's a cheap request producing a large response — a small amplification vector.

Could we bound this? Stopping after something like 32–64 projected names would close it off at no practical cost, since real clients will only ever send a handful.

SamMorrowDrums: Projected header reflection is now capped at 64 unique valid names, with an adversarial flood test covering the bound, deduplication, and response amplification.

> [!NOTE]

PR #3175, opened 2026-08-28
feat(http): add opt-in static authentication

### LINKED ISSUE #2946: http mode: support static token auth (env) for single-tenant self-hosted deployments

github-mcp-server http requires a per-request Authorization header and never reads GITHUB_PERSONAL_ACCESS_TOKEN from the environment - token extraction happens only in pkg/http/middleware/token.go (ExtractUserToken), and httpCmd's ServerConfig has no token field (cmd/github-mcp-server/main.go). The env var is consulted only by the stdio command.

That fits the multi-tenant/centralized use case http mode was built for (#471, #997). But it leaves a gap for single-tenant self-hosted deployments: an org running one server instance with one service credential, fronted by a gateway/tunnel that cannot inject custom headers (in our case, Anthropic's MCP tunnels; the same applies to any transport that forwards requests as-is). The server starts cleanly with the env var set, then 401s every request - a confusing failure, since the README's env-var examples don't mention they're stdio-only.

### Proposed solution

In http mode, when no Authorization header is present on a request, fall back to a statically configured token (GITHUB_PERSONAL_ACCESS_TOKEN / GITHUB_ACCESS_TOKEN), opt-in via a flag, e.g.:

github-mcp-server http --static-auth  # or --token-from-env

Per-request tokens would still take precedence, preserving the multi-tenant behavior. Notably, the 2026-01-28 GitHub changelog announcing HTTP mode described exactly this fallback ("falls back GITHUB_PERSONAL_ACCESS_TOKEN if no header is provided"), but the shipped behavior doesn't include it.

dawNotPoi:
## Summary

Add explicit opt-in static authentication for single-tenant HTTP deployments.
When enabled, requests without an `Authorization` header use the process
`GITHUB_PERSONAL_ACCESS_TOKEN`; per-request credentials remain authoritative.

## Why

HTTP mode currently requires every caller to supply an `Authorization` header,
even when a single-tenant deployment already has a service credential in the
process environment. Gateways and tunnels that cannot inject custom headers
therefore receive a 401 for every request.

Fixes #2946

## What changed

- Add the HTTP-only `--static-auth` flag and `GITHUB_STATIC_AUTH` environment
  equivalent, with startup validation for the configured token.
- Fall back only when the `Authorization` header is absent; explicit empty,
  malformed, or unsupported headers remain errors.
- Preserve existing token-context and PAT-scope processing, and document the
  shared-credential security boundary.
- Reject browser-origin requests that would implicitly consume the shared
  credential, while preserving explicit bearer-token CORS and OAuth metadata.

## MCP impact

- [ ] No tool or API changes
- [ ] New tool added

HTTP transport authentication behavior changes only when static auth is
explicitly enabled. MCP tool definitions and schemas are unchanged.

- N/A — this changes HTTP authentication configuration, not an MCP tool.

## Security / limits

Static authentication is disabled by default. When enabled, every caller that
omits the header receives the shared credential's permissions, so the docs
require an authenticating trusted gateway or access boundary and recommend
least privilege and `--read-only`; loopback alone is not presented as
authentication. Browser-origin requests cannot consume the fallback token.
Explicit request credentials always take precedence, and configuration errors
never include the token value.

## Tool renaming

No tools are renamed.

## Lint & tests

`./script/lint` reports zero issues. `./script/test` and an uncached focused
`go test -race -count=1 ./pkg/utils ./pkg/http/... ./cmd/github-mcp-server`
both pass.

## Docs

- [ ] Not needed

Updated `docs/streamable-http.md` with configuration, precedence, startup
validation, browser-origin behavior, and shared-credential safety guidance.

### DIFF
--- a/cmd/github-mcp-server/main.go
+++ b/cmd/github-mcp-server/main.go
@@ -166,6 +166,11 @@ var (
 		Short: "Start HTTP server",
 		Long:  `Start an HTTP server that listens for MCP requests over HTTP.`,
 		RunE: func(_ *cobra.Command, _ []string) error {
+			staticToken, err := resolveHTTPStaticToken(viper.GetBool("static-auth"), viper.GetString("personal_access_token"))
+			if err != nil {
+				return err
+			}
+
 			// Parse toolsets (same approach as stdio — see comment there)
 			var enabledToolsets []string
 			if viper.IsSet("toolsets") {
@@ -199,6 +204,7 @@ var (
 			httpConfig := ghhttp.ServerConfig{
 				Version:              version,
 				Host:                 viper.GetString("host"),
+				StaticToken:          staticToken,
 				Port:                 viper.GetInt("port"),
 				ListenHost:           viper.GetString("listen-host"),
 				BaseURL:              viper.GetString("base-url"),
@@ -268,6 +274,7 @@ func init() {
 	httpCmd.Flags().String("authorization-server", "", "Override the authorization server URL in OAuth resource metadata. Useful when deploying behind an OAuth proxy (e.g. for GHES). Env: GITHUB_AUTHORIZATION_SERVER")
 	httpCmd.Flags().Bool("scope-challenge", false, "Enable OAuth scope challenge responses")
 	httpCmd.Flags().Bool("trust-proxy-headers", false, "Honor X-Forwarded-Host and X-Forwarded-Proto when constructing OAuth resource metadata URLs. Only enable when the server is deployed behind a trusted proxy that sets these headers. Ignored when --base-url is set.")
+	httpCmd.Flags().Bool("static-auth", false, "Allow requests without an Authorization header to use GITHUB_PERSONAL_ACCESS_TOKEN. For single-tenant deployments behind a trusted access boundary only. Env: GITHUB_STATIC_AUTH")
 
 	// Bind flag to viper
 	_ = viper.BindPFlag("toolsets", rootCmd.PersistentFlags().Lookup("toolsets"))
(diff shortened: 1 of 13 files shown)


cmd/github-mcp-server/main.go
```diff
@@ -309,6 +317,19 @@ func initConfig() {
...
 
+func resolveHTTPStaticToken(enabled bool, token string) (string, error) {
+	if !enabled {
+		return "", nil
+	}
+	if token == "" {
+		return "", errors.New("HTTP static auth requires GITHUB_PERSONAL_ACCESS_TOKEN")
+	}
+	if _, err := utils.ParseToken(token); err != nil {
+		return "", fmt.Errorf("invalid GITHUB_PERSONAL_ACCESS_TOKEN for HTTP static auth: %w", err)
```
Nithish1142: Testing

tiagovilasboas:
### Staff AppSec (HTTP static-auth)

**Verdict: Approve with comments** (conversation comment — formal review UI often disabled here.)

Opt-in `--static-auth` / `GITHUB_STATIC_AUTH` for single-tenant HTTP is implemented carefully: default-off, fail-closed token validation (secrets not echoed), fallback **only** when `Authorization` is absent (empty/malformed never downgrade to the shared PAT), `StaticAuthBrowserGuard` before CORS so `Origin` requests cannot consume the ambient credential, and PAT-scope middleware still sees the fallback token. Docs WARNING correctly says loopback ≠ auth.

**Residual (acknowledge on merge):** this authenticates callers *to the MCP server* with nothing — any non-browser client that can reach the socket and omits `Authorization` gets the shared PAT. The Origin guard is browser/ambient protection only.

**Non-blocking asks:** (1) loud startup warning when static auth is on and listen host is not loopback; (2) startup warning when static auth is on without `--read-only` (docs recommend it; writes under a shared PAT are the scary case). Happy to re-check if those land as a follow-up.

Refs #2946.

sconklin79-arch (approved):

PR #3182, opened 2026-08-29
Add expectedHeadSha pinning to merge_pull_request

### LINKED ISSUE #3181: Add atomic head SHA pinning to merge_pull_request
## Problem

`merge_pull_request` currently has no way for a caller to atomically assert
which pull-request HEAD is being merged.

An agent can inspect a pull request, validate CI/review readiness, and then
call `merge_pull_request`, but the PR HEAD may change between those steps.
Without a compare-and-swap condition on the merge call, that creates a TOCTOU
gap.

The GitHub REST merge endpoint already supports the optional `sha` field for
this purpose, and `go-github` already exposes it as
`PullRequestOptions.SHA`. The MCP tool currently does not expose or forward
that capability.

## Proposed change

Add an optional `expectedHeadSha` input to `merge_pull_request` and pass it
directly to `github.PullRequestOptions.SHA`.

This follows the existing `expectedHeadSha` naming used by
`update_pull_request_branch`.

When supplied:

- the exact expected HEAD SHA is sent to GitHub as the REST `sha` field;
- a changed HEAD is rejected by GitHub rather than merging a different commit;
- no retry or permissive fallback is introduced.

When omitted, behavior remains backward compatible because the underlying
`sha` field is optional/omitted.

## Scope

No authentication, credential, logging, permission, or unrelated tool changes.

jcosta1970:
## Summary

Adds optional `expectedHeadSha` pinning to `merge_pull_request`.

The value is forwarded to `github.PullRequestOptions.SHA`, which
`go-github` serializes as the GitHub REST merge endpoint's optional
`sha` field.

This closes the TOCTOU gap between reviewing a pull request HEAD and merging
it: if the HEAD changes, GitHub rejects the merge instead of merging the new
commit.

## Changes

- add optional `expectedHeadSha` to the `merge_pull_request` schema;
- pass it to `PullRequestOptions.SHA`;
- verify the value reaches the REST request as `sha`;
- verify a 409 head mismatch fails closed with no retry;
- update the merge tool schema snapshot;
- regenerate project docs when local Go tooling is available.

## Backward compatibility

The new field is optional. Existing callers that omit
`expectedHeadSha` retain the current behavior.

## Local validation

- Go tests: `SKIPPED_GO_NOT_INSTALLED`
- generated docs: `SKIPPED_GO_NOT_INSTALLED`
- golangci-lint: `SKIPPED_GOLANGCI_LINT_NOT_INSTALLED`
- `git diff --check`: PASS

Fixes #3181

### DIFF
--- a/README.md
+++ b/README.md
@@ -1182,6 +1182,7 @@ The following sets of tools are available:
   - **OAuth Challenge Scopes**: `repo`
   - `commit_message`: Extra detail for merge commit (string, optional)
   - `commit_title`: Title for merge commit (string, optional)
+  - `expectedHeadSha`: The expected SHA of the pull request's HEAD ref (string, optional)
   - `merge_method`: Merge method (string, optional)
   - `owner`: Repository owner (string, required)
   - `pullNumber`: Pull request number (number, required)
--- a/docs/feature-flags.md
+++ b/docs/feature-flags.md
@@ -357,7 +357,7 @@ runtime behavior (such as output formatting) won't appear here.
 ### `thread_resolution_reason`
 
 - **pull_request_review_write** - Write operations (create, submit, delete) on pull request reviews
-  - **Required OAuth Scopes**: `repo`
+  - **OAuth Challenge Scopes**: `repo`
   - `body`: Review comment text (string, optional)
   - `commitID`: SHA of commit to review (string, optional)
   - `event`: Review action to perform. (string, optional)
--- a/pkg/github/__toolsnaps__/merge_pull_request.snap
+++ b/pkg/github/__toolsnaps__/merge_pull_request.snap
@@ -27,6 +27,10 @@
         "description": "Title for merge commit",
         "type": "string"
       },
+      "expectedHeadSha": {
+        "description": "The expected SHA of the pull request's HEAD ref",
+        "type": "string"
+      },
       "merge_method": {
         "description": "Merge method",
         "enum": [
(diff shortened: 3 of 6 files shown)


leuasseurfarrelds247-arch (approved):

CAOShurong (approved):
Reviewed exact head `ef6e104f1373ae12a9529bbc8ed4d8041f643fd1`.

The optional input is passed directly to `go-github`'s `PullRequestOptions.SHA`, which serializes it as the REST merge endpoint's `sha` guard. The focused tests cover both the successful request body and the important failure path: a 409 head mismatch returns an MCP error after exactly one request, so the handler does not silently retry against a different HEAD. Omitting the field keeps the existing behavior.

Verification on this exact head:
- `go test ./pkg/github -run '^Test_MergePullRequestSHAPinning$' -count=1` — pass
- `go vet ./pkg/github` — pass
- `go build ./...` — pass
- `git diff --check` — pass

The full `go test ./pkg/github -count=1` run still reports the Windows toolsnap icon-order diffs; the same failures reproduce on exact base `febc3293a4feb70e62399f39a26b082f78b9b176`, so they are not introduced by this patch.

I did not find a blocking issue in the SHA-pinning change.

AI assistance disclosure: Codex assisted with the local verification and base comparison; I verified the cited heads, diff, and results.

SamMorrowDrums (approved):

PR #3187, opened 2026-08-30
Fix e2e harness compilation against go-github v89 and go-sdk v1.7

ppoffice:
## Problem

The e2e test package no longer compiles under --tags e2e, so the e2e suite cannot run at all. It lagged behind two dependency migrations that landed on main:

- **go-github v89** — NewClient now returns (*Client, error) and WithEnterpriseURLs moved from a *Client method to a ClientOptionsFunc.
- **go-sdk v1.7** — ghmcp.NewMCPServer(MCPServerConfig) was replaced by ghmcp.NewStdioMCPServer(ctx, github.MCPServerConfig).

Because the package only builds with --tags e2e, this breakage was invisible to the default CI build.

## Changes

All in e2e/e2e_test.go:

- getRESTClient: build the client with gogithub.NewClient(gogithub.WithAuthToken(token)) (+ WithEnterpriseURLs as an option for GHEC) and handle the new error return.
- Two inline cleanup/review call sites: same NewClient migration.
- ListReviewers: drop the trailing 
il options arg (removed in v89).
- In-process setupMCPClient branch: use ghmcp.NewStdioMCPServer(ctx, github.MCPServerConfig{...}) with the required Logger and Version fields.

## Verification

- go vet -tags e2e ./e2e/ passes.
- go test -tags e2e ./e2e/ now runs (was [build failed] before).

Not run against the live API here (needs a GITHUB_MCP_SERVER_E2E_TOKEN + Docker).

### DIFF
--- a/e2e/e2e_test.go
+++ b/e2e/e2e_test.go
@@ -19,6 +19,7 @@ import (
 	"github.com/github/github-mcp-server/internal/ghmcp"
 	"github.com/github/github-mcp-server/pkg/github"
 	"github.com/github/github-mcp-server/pkg/translations"
+	"github.com/github/github-mcp-server/pkg/utils"
 	gogithub "github.com/google/go-github/v89/github"
 	"github.com/modelcontextprotocol/go-sdk/mcp"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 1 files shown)


SamMorrowDrums (approved):

adamczykmathew5-netizen (changes requested):
Reassessment is need to handle the problem

PR #3234, opened 2026-09-07
fix: include file body in get_file_contents text content

### LINKED ISSUE #3189: get_file_contents returns [non-text content: resource] instead of file text in Claude voice mode
## Describe the bug

When using the hosted GitHub MCP connector from Claude (claude.ai), `get_file_contents` for a Markdown file sometimes returns no file body — the tool result only shows `[non-text content: resource]` instead of the file text. In my testing this happened consistently while using Claude's voice mode, while the exact same file was returned correctly in text mode within the same session.

## Affected version

Hosted GitHub MCP connector (remote server via claude.ai's GitHub connector). Client: Claude (claude.ai), observed on 2026-08-31.

## Steps to reproduce

1. Connect Claude (claude.ai) to the hosted GitHub MCP server.
2. In a voice-mode conversation, ask Claude to read a Markdown file via `get_file_contents` — e.g. `README.md` of this repository (`github/github-mcp-server`), which is where I reproduced it, so it is not specific to my own repositories.
3. The tool result contains `[non-text content: resource]` and no file text, so the model cannot see the file content.
4. In the same session, switching to text mode and requesting the same file returns the full content normally.

## What I observed while isolating the issue

1. **Mode-dependent**: the failure occurred in voice mode; the same file succeeded in text mode. I confirmed both the reproduction and the recovery within a single session.
2. **Workaround exists**: fetching the same content via `list_commits` + `get_commit` (with `detail: full_patch`) returns the full file text even in voice mode.
3. **Not repo-specific**: reproduced with `github/github-mcp-server`'s own `README.md`.
4. **Not size/content-dependent**: a file that failed in voice mode succeeded in text mode in the same session, so file size/content does not seem to be the deciding factor.

I searched existing issues and found #1168 (binary `mimeType` handling, already closed) and #1176 (`resource not found` error), but neither matches this symptom: here the call succeeds but the content arrives as an embedded resource the client renders as `[non-text content: resource]` rather than as text.

## Expected behavior

`get_file_contents` returns the Markdown file body as text (or in a form the client can present to the model) regardless of the client conversation mode.

## Additional context

My guess is that the response uses an embedded resource content type that some client surfaces (e.g. Claude's voice mode) cannot unwrap, whereas plain text content would work everywhere. If there is a way to force a plain-text response, I'm happy to test it.

Bruce-Yii:
## Summary

Text-file responses from `get_file_contents` now carry the file body in the `TextContent` as well as in the embedded resource, so client surfaces that drop embedded resources still expose the file to the model.

## Why

Fixes #3189. In Claude voice mode the same `get_file_contents` call returned only `[non-text content: resource]` plus the `successfully downloaded text file (SHA: …)` line, while text mode worked — the body lived exclusively inside the `EmbeddedResource`.

## What changed

- `pkg/github/repositories.go`: in the text-file branch of `GetFileContents`, append the file body to the text message (SHA line + blank line + body). Embedded resource, content order/length, and tool schema are unchanged.
- `pkg/github/repositories_text_content_test.go`: regression test asserting `Content[0]` (`TextContent`) contains the body and `Content[1]` (resource) still carries it.
- Untouched: binary, >=1MB resource-link, empty-file, directory, and symlink-metadata paths (symlink metadata message must stay parseable JSON).

## MCP impact

- [x] Tool schema or behavior changed — response text now includes the file body alongside the unchanged embedded resource (additive; no schema/snapshot change).

- "Read the README.md of owner/repo via get_file_contents" — text clients see the body in `TextContent`; resource-unwrapping clients see it in both places.

## Security / limits

- [x] No security or limits impact — the inlined bytes are exactly the bytes already returned in the resource; the existing 1MB cap still routes large files to `ResourceLink`.

## Tool renaming

## Lint & tests

- [x] Linted locally with `golangci-lint run ./pkg/github/` (repo config, Go 1.25.12 toolchain): 0 issues; `go vet` clean; `git diff --check` clean
- [x] Tested locally: new regression test failed before the fix and passes after; all `Test_GetFileContents*` (incl. symlink, IFC, field-filtering) pass. Full `pkg/github` suite shows only the pre-existing Windows-checkout failure `TestAssignCopilotToIssueWithIntent`, which fails identically on clean main.
- [x] Docs — Not needed (no tool/schema change; toolsnaps unaffected)


Bruce-Yii:
Closing this since the maintainer direction in #3189 is to keep the embedded-resource response rather than duplicate the file body into TextContent. Thanks for clarifying the intended compatibility boundary.

